PR #83, opened 2021-03-23
Fix copy-paste bug

sidpremkumar:
Ensure that copy-paste is sent as a single string
Relates to JIRA: CWC-574

### DIFF
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -38,6 +38,8 @@ import { Logger } from './logger.service/logger';
 import { LoggerConfigService } from './logger-config.service/logger-config.service';
 import { SsmTunnelService } from './ssm-tunnel/ssm-tunnel.service';
 import { KeySplittingService } from '../webshell-common-ts/keysplitting.service/keysplitting.service';
+import { Observable } from 'rxjs';
+import { bufferTime, filter } from 'rxjs/operators';
 
 export class CliDriver
 {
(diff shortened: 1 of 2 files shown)


Nabeelperson (approved):

src/cli-driver.ts
```diff
@@ -421,12 +422,18 @@ ssh <user>@bzero-<ssm-target-id-or-name>
...
-                    process.stdin.on('keypress', (_, key) => terminal.writeString(key.sequence));
+                    let source = new Observable<string>(function (observer) {
+                        // To get 'keypress' events you need the following lines
+                        // ref: https://nodejs.org/api/readline.html#readline_readline_emitkeypressevents_stream_interface
+                        const readline = require('readline');
+                        readline.emitKeypressEvents(process.stdin);
+                        process.stdin.setRawMode(true);
+                        process.stdin.on('keypress', (_, key) => observer.next(key.sequence));
+                    });
+                    source.pipe(bufferTime(50), filter(buffer => buffer.length > 0)).subscribe(keypresses => {
```
sidpremkumar, line 433: @Nabeelperson asked why the filter call: 
> otherwise it'll just return "" empty string, that buffer timer will wait 50, and if nothing happens it'll just send an event. But we only want events with some value, so we check to ensure that buffer has something there

lipmas (approved):

PR #84, opened 2021-03-25
File transfer now disambiguates targetNames to targetIds (CWC-520)

Nabeelperson:
In this PR:
 - Add targetName disambiguation to the copy command

### DIFF
--- a/src/ssm-tunnel/ssm-tunnel.service.ts
+++ b/src/ssm-tunnel/ssm-tunnel.service.ts
@@ -160,7 +160,7 @@ export class SsmTunnelService
             }
 
             if(matchedTarget.length > 1) {
-                throw new Error(`Multiple targets found with name ${targetName}`);
+                throw new Error(`Multiple targets found with name ${targetName} please use targetId instead (zli lt)`);
             }
 
             return matchedTarget[0];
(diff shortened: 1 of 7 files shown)


src/ssm-tunnel/ssm-tunnel.service.ts
```diff
@@ -160,7 +160,7 @@ export class SsmTunnelService
             }
 
             if(matchedTarget.length > 1) {
-                throw new Error(`Multiple targets found with name ${targetName}`);
+                throw new Error(`Multiple targets found with name ${targetName} please use targetId instead (zli lt)`);
```
sidpremkumar, line 163: `lt` is just the alias (and I didn't know it existed till recently), Maybe: 
```
Multiple targets found with name ${targetName} please use targetId instead (zli list-targets/lt)
```

sidpremkumar (approved):
Tested upload and download and working on my end as expected

src/ssm-tunnel/ssm-tunnel.service.ts
Nabeelperson: For the brevity of the message I just put the alias, if we put both I feel they might copy-paste the whole string learning to this message being confusing.

PR #86, opened 2021-03-30
SSH style syntax

Nabeelperson:
Version 4.13.0

Change log: 
- New target name and id disambiguation that does not require target id and type to be used unless there is a collision, you can either use GUID or pass a -t /  --targetType  flag
- list-targets command now optionally shows GUID with -i / --showId  flag
- list-targets command now optionally shows SSM taget agent version and status, for other target types shows N/A
- list-targets command now has --json  flag to output pipeable json results, best used with --silent  flag to hide all other text output
- Policy checks are preformed before connect/file transfer/tunnel, you are warned if you do not have a valid policy for that verb, if the verb is valid you are warned if the target user is invalid and shown all valid target users
- Revised help examples

### DIFF
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -1,7 +1,7 @@
-import { IdP, TargetType } from './types';
+import { IdP, ParsedTargetString, TargetSummary, TargetType } from './types';
 import {
-    targetStringExample,
-    TargetSummary,
+    disambiguateTarget,
+    targetStringExample
 } from './utils';
 import { ConfigService } from './config.service/config.service';
 import { MixpanelService } from './mixpanel.service/mixpanel.service';
@@ -14,7 +14,7 @@ import { KeySplittingService } from '../webshell-common-ts/keysplitting.service/
 // Handlers
 import { initMiddleware, oAuthMiddleware, mixedPanelTrackingMiddleware, fetchDataMiddleware } from './handlers/middleware.handler';
 import { sshProxyConfigHandler } from './handlers/ssh-proxy-config.handler';
-import { sshProxyHandler } from './handlers/ssh-proxy.handler';
+import { sshProxyHandler, SshTunnelParameters } from './handlers/ssh-proxy.handler';
 import { loginHandler } from './handlers/login.handler';
 import { connectHandler } from './handlers/connect.handler';
 import { listTargetsHandler } from './handlers/list-targets.handler';
@@ -25,6 +25,7 @@ import { logoutHandler } from './handlers/logout.handler';
 // 3rd Party Modules
 import { Dictionary, includes } from 'lodash';
 import yargs from 'yargs';
+import { cleanExit } from './handlers/clean-exit.handler';
 
 
 export class CliDriver
@@ -95,36 +96,6 @@ export class CliDriver
                 this.envs = fetchDataResponse.envs;
             })
             .command(
-                'ssh-proxy-config',
-                'Generate ssh configuration to be used with the ssh-proxy command',
-                (_) => {},
-                async (_) => {
-                    sshProxyConfigHandler(this.configService, this.logger, this.processName);
-                }
-            )
-            .command(
-                'ssh-proxy <host> <user> <port> <identityFile>',
-                'ssh proxy command (run ssh-proxy-config command to generate configuration)',
-                (yargs) => {
-                    return yargs
-                        .positional('host', {
-                            type: 'string',
-                        })
-                        .positional('user', {
-                            type: 'string',
-                        })
-                        .positional('port', {
-                            type: 'number',
-                        })
-                        .positional('identityFile', {
-                            type: 'string'
-                        });
-                },
-                async (argv) => {
-                    await sshProxyHandler(this.configService, this.logger, argv, this.keySplittingService, this.envMap);
-                }
-            )
-            .command(
                 'login <provider>',
                 'Login through a specific provider',
                 (yargs) => {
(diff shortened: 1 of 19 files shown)


lipmas (approved):

lipmas:
Fix  linter errors by running `npm run lint -- --fix`

src/cli-driver.ts
```diff
@@ -199,33 +174,133 @@ export class CliDriver
...
+                                choices: this.targetTypeChoices,
+                                demandOption: false,
+                                alias: 't'
+                            }
+                        )
+                        .example('copy ssm-user@neat-target:/home/ssm-user/file.txt /Users/coolUser/newFileName.txt', 'Download example, relative to your machine')
+                        .example('copy /Users/coolUser/secretFile ssm-user@neat-target:/home/ssm-user/newFileName', 'Upload example, relative to your machine');
+                },
+                async (argv) => {
+                    const sourceParsedTarget = await disambiguateTarget(argv.targetType, argv.source, this.logger, this.dynamicConfigs, this.ssmTargets, this.sshTargets, this.envs);
```
sidpremkumar, line 232: I think all of this should be inside `copyHandler`

src/cli-driver.ts
```diff
@@ -199,33 +174,133 @@ export class CliDriver
...
                 async (argv) => {
-                    await copyHandler(this.configService, this.logger, argv, this.dynamicConfigs, this.ssmTargets, this.sshTargets, this.envs);
+                    let prefix = 'bzero-';
+                    const configName = this.configService.getConfigName();
+                    if(configName != 'prod') {
+                        prefix = `${configName}-${prefix}`;
+                    }
+
+                    if(! argv.host.startsWith(prefix)) {
+                        // TODO throw versus exit?
```
sidpremkumar, line 282: I think exit makes sense, its not an error really, its an invalid params issue. but no real preference

src/cli-driver.ts
```diff
@@ -199,33 +174,133 @@ export class CliDriver
...
-                        .example('copy ssm ssm-user@95b72b50-d09c-49fa-8825-332abfeb013e:/home/ssm-user/file.txt /Users/coolUser/newFileName.txt', 'SSM Download example')
-                        .example('copy ssm /Users/coolUser/secretFile.txt ssm-user@neat-target:/home/ssm-user/newFileName', 'SSM Upload example')
-                        .example('copy ssh /Users/coolUser/neatFile.txt cool-alias:/home/ssm-user/newFileName.txt', 'SSH Upload example');
+                        .positional('identityFile', {
+                            type: 'string'
+                        });
                 },
                 async (argv) => {
-                    await copyHandler(this.configService, this.logger, argv, this.dynamicConfigs, this.ssmTargets, this.sshTargets, this.envs);
+                    let prefix = 'bzero-';
```
sidpremkumar, line 275: Similarly, this should all go inside `sshProxyHandler`

src/handlers/copy.handler.ts
```diff
@@ -1,74 +1,71 @@
...
+import { FileService, PolicyQueryService } from '../http.service/http.service';
 import { ConfigService } from '../config.service/config.service';
 import { Logger } from '../logger.service/logger';
 import { cleanExit } from './clean-exit.handler';
-
 import fs from 'fs';
-import { disambiguateTargetName } from './connect.handler';
-import { EnvironmentDetails } from '../http.service/http.service.types';
-import { TargetType } from '../types';
+import { ParsedTargetString, TargetType } from '../types';
```
sidpremkumar, line 6: I was trying to separate imports by `local` vs `3rd Party`. If we want to maintain that, this should change

src/handlers/copy.handler.ts
```diff
@@ -1,74 +1,71 @@
...
+    }
 
-    // Given the command level parsing only accepts ssm and ssh
-    // we should not expect to his this code block
     if(parsedTarget.type == TargetType.DYNAMIC)
     {
-        logger.error('Cannot file transfer with a dynamic access config');
-        logger.warn('Please create a new dynamic access target or fetch an existing one (zli lt)');
-        cleanExit(1, logger);
+        this.logger.error('Cannot file transfer with a dynamic access config');
```
sidpremkumar, line 45: `this.logger` will be undefined no?

src/cli-driver.ts
Nabeelperson: Moving forward I want to list the argument parsing out of the handlers and into the cli-driver. This way we can recycle arg parsing across commands that require the same arguments (`targetString` being the main one).

src/cli-driver.ts
Nabeelperson: Exit then.

src/handlers/copy.handler.ts
Nabeelperson: Noted.

src/handlers/copy.handler.ts
Nabeelperson: Noted, local should go first then external libs?

src/handlers/copy.handler.ts
Nabeelperson: Yup, that is a bug for sure, will remove `this`.

Nabeelperson: 4b19eb1

src/handlers/copy.handler.ts
Nabeelperson: 4b19eb1

sidpremkumar: Honestly not sure, I was just separating them and putting local first, but you can follow this: https://github.com/bastionzero/cloud-common/pull/164#discussion_r605931241

Which states:
```
# Build In

# 3rd Party

# Local
```

No preference, as long as they are separated IMO. You can also ignore me and wait till we agree on a standard for `zli` up 2 you

src/cli-driver.ts
Nabeelperson: 4b19eb1

src/handlers/copy.handler.ts
Nabeelperson: I think this is hard to enforce via linter or otherwise, my preference is by vendor in alphabetical order but yeah, not going to change this for now.

PR #88, opened 2021-04-01
Linter fixes for variable declarations

lipmas:
updates submodule to use new linter rules from https://github.com/bastionzero/webshell-common-ts/pull/26

### DIFF
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -65,7 +65,7 @@ export class CliDriver
             .usage('$0 <cmd> [args]')
             .wrap(null)
             .middleware(async (argv) => {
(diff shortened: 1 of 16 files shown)


Nabeelperson (approved):
Bump version number to 4.12.1 please.

sidpremkumar (approved):

lipmas:
@Nabeelperson are version updates necessary for non-functional changes like this one? Adding overhead of new version release for every change seems like a bit much IMO

Nabeelperson:
Ideally this is automated, we can skip it for this update but at some point a pipeline should manage bumping versions when we want to similar to webshell-release-ci.

PR #89, opened 2021-04-08
(CWC-599) Feat: Keysplitting for Interactive Shell

lipmas:
webshell-common: https://github.com/bastionzero/webshell-common-ts/pull/28

### DIFF
--- a/src/config.service/config.service.ts
+++ b/src/config.service/config.service.ts
@@ -70,15 +70,15 @@ export class ConfigService implements ConfigInterface {
         this.tokenService = new TokenService(this, logger);
     }
 
-    public updateKeySplitting(data: KeySplittingConfigSchema) {
+    public updateKeySplitting(data: KeySplittingConfigSchema): void {
         this.config.set('keySplitting', data);
     }
 
-    public loadKeySplitting() {
+    public loadKeySplitting(): KeySplittingConfigSchema {
         return this.config.get('keySplitting');
     }
 
-    public removeKeySplitting() {
+    public removeKeySplitting(): void {
         this.config.delete('keySplitting');
     }
 
--- a/src/config.service/config.service.ts
+++ b/src/config.service/config.service.ts
@@ -78,6 +78,10 @@ export class ConfigService implements ConfigInterface {
         return this.config.get('keySplitting');
     }
 
+    public removeKeySplitting() {
+        this.config.delete('keySplitting');
+    }
+
     public getConfigName() {
         return this.configName;
     }
--- a/src/handlers/copy.handler.ts
+++ b/src/handlers/copy.handler.ts
@@ -31,21 +31,21 @@ export async function copyHandler(
     if(! response.allowed)
     {
         logger.error('You do not have sufficient permission to file transfer with the target');
-        cleanExit(1, logger);
+        await cleanExit(1, logger);
     }
 
     const allowedTargetUsers = response.allowedTargetUsers.map(u => u.userName);
     if(response.allowedTargetUsers && ! _.includes(allowedTargetUsers, parsedTarget.user)) {
         logger.error(`You do not have permission to file transfer as targetUser: ${parsedTarget.user}`);
         logger.info(`Current allowed users for you: ${allowedTargetUsers}`);
-        cleanExit(1, logger);
+        await cleanExit(1, logger);
     }
 
     if(parsedTarget.type == TargetType.DYNAMIC)
     {
         logger.error('Cannot file transfer with a dynamic access config');
         logger.warn('Please create a new dynamic access target or fetch an existing one (zli lt)');
-        cleanExit(1, this.logger);
+        await cleanExit(1, this.logger);
     }
 
     // figure out upload or download
--- a/src/handlers/copy.handler.ts
+++ b/src/handlers/copy.handler.ts
@@ -45,7 +45,7 @@ export async function copyHandler(
     {
         logger.error('Cannot file transfer with a dynamic access config');
         logger.warn('Please create a new dynamic access target or fetch an existing one (zli lt)');
-        await cleanExit(1, this.logger);
+        await cleanExit(1, logger);
     }
 
     // figure out upload or download
--- a/src/oauth.service/oauth.service.ts
+++ b/src/oauth.service/oauth.service.ts
@@ -182,18 +182,18 @@ export class OAuthService implements IDisposable {
                         this.logger.error('Stale log in detected');
                         this.logger.info('You need to log in, please run \'zli login --help\'');
                         this.configService.logout();
-                        cleanExit(1, this.logger);
+                        await cleanExit(1, this.logger);
                     } else {
                         this.logger.error('Unexpected error during oauth refresh');
                         this.logger.info('Please log in again');
                         this.configService.logout();
-                        cleanExit(1, this.logger);
+                        await cleanExit(1, this.logger);
                     }
                 }
             }
         } else {
             this.logger.warn('You need to log in, please run \'zli login --help\'');
-            process.exit(1);
+            await cleanExit(1, this.logger);
         }
 
         return this.configService.getAuth();
--- a/src/oauth.service/oauth.service.ts
+++ b/src/oauth.service/oauth.service.ts
@@ -1,4 +1,4 @@
-import { AuthorizationParameters, Client, custom, generators, Issuer, TokenSet, UserinfoResponse } from 'openid-client';
+import { AuthorizationParameters, Client, custom, errors, generators, Issuer, TokenSet, UserinfoResponse } from 'openid-client';
 import open from 'open';
 import { IDisposable } from '../../webshell-common-ts/utility/disposable';
 import { ConfigService } from '../config.service/config.service';
@@ -7,6 +7,7 @@ import { setTimeout } from 'timers';
 import { Logger } from '../../src/logger.service/logger';
 import { loginHtml } from './templates/login';
 import { logoutHtml } from './templates/logout';
+import { cleanExit } from '../../src/handlers/clean-exit.handler';
 
 export class OAuthService implements IDisposable {
     private server: http.Server; // callback listener
@@ -158,6 +159,44 @@ export class OAuthService implements IDisposable {
         const tokenSet = this.configService.tokenSet();
         const userInfo = await client.userinfo(tokenSet);
         return userInfo;
+    }
+
+    // Returns the current OAuth idtoken. Refreshes it before returning if expired
+    public async getIdToken(): Promise<string> {
+
+        const tokenSet = this.configService.tokenSet();
+
+        // decide if we need to refresh or prompt user for login
+        if(tokenSet)
+        {
+            if(this.configService.tokenSet().expired())
+            {
+                try {
+                    this.logger.debug('Refreshing oauth token');
+
+                    const newTokenSet = await this.refresh();
+                    this.configService.setTokenSet(newTokenSet);
+                    this.logger.debug('Oauth token refreshed');
+                } catch(e) {
+                    if(e instanceof errors.RPError || e instanceof errors.OPError) {
+                        this.logger.error('Stale log in detected');
+                        this.logger.info('You need to log in, please run \'zli login --help\'');
+                        this.configService.logout();
+                        cleanExit(1, this.logger);
+                    } else {
+                        this.logger.error('Unexpected error during oauth refresh');
+                        this.logger.info('Please log in again');
+                        this.configService.logout();
+                        cleanExit(1, this.logger);
+                    }
+                }
+            }
+        } else {
+            this.logger.warn('You need to log in, please run \'zli login --help\'');
+            process.exit(1);
+        }
+
+        return this.configService.getAuth();
     }
 
     dispose(): void {
--- a/src/terminal/terminal.ts
+++ b/src/terminal/terminal.ts
@@ -5,7 +5,7 @@ import { IDisposable } from '../../webshell-common-ts/utility/disposable';
 import { KeySplittingService } from '../../webshell-common-ts/keysplitting.service/keysplitting.service';
 
 import { ConfigService } from '../config.service/config.service';
-import { IShellWebsocketService, ShellState, TerminalSize } from '../../webshell-common-ts/shell-websocket.service/shell-websocket.service.types';
+import { IShellWebsocketService, ShellEvent, ShellEventType, TerminalSize } from '../../webshell-common-ts/shell-websocket.service/shell-websocket.service.types';
 import { ZliAuthConfigService } from '../config.service/zli-auth-config.service';
 import { Logger } from '../logger.service/logger';
 import { SsmTargetService } from '../http.service/http.service';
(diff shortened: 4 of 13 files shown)


tflpd (approved):

tflpd (approved):

src/terminal/terminal.ts
```diff
@@ -63,9 +126,17 @@ export class ShellTerminal implements IDisposable
...
 
-    public writeString(input: string) : void {
-        if(! this.blockInput)
+    public async writeString(input: string) : Promise<void> {
+        if(! this.attached && input === '\u001br') {
+            await this.shellWebsocketService.shellReattach();
+            this.attached = true;
+            this.blockInput = false;
+            return;
+        }
```
Nabeelperson, line 135: Can we get a comment here about what is happening with watching for that sequence please.

src/config.service/config.service.ts
```diff
@@ -78,6 +78,10 @@ export class ConfigService implements ConfigInterface {
         return this.config.get('keySplitting');
     }
 
+    public removeKeySplitting() {
```
Nabeelperson, line 81: This along with the other keysplitting related function should be typed for the return or void.

src/terminal/terminal.ts
```diff
@@ -1,49 +1,112 @@
...
+            } else {
+                this.logger.warn(`Agent version ${ssmTargetInfo.agentVersion} not compatible with keysplitting...falling back to non-keysplitting shell`);
+                return this.createSshShellWebsocketService();
+            }
+        } else {
+            throw new Error(`Unhandled target type ${targetType}`);
+        }
+    }
+
+    private createSshShellWebsocketService() {
```
Nabeelperson, line 51: Please type return.

src/terminal/terminal.ts
```diff
@@ -1,49 +1,112 @@
...
+        return new SshShellWebsocketService(
+            this.logger,
+            new ZliAuthConfigService(this.configService, this.logger),
+            this.connectionId,
+            this.inputSubject,
+            this.resizeSubject
+        );
+    }
+
+    private createSsmShellWebsocketService(ssmTargetInfo: SsmTargetSummary) {
```
Nabeelperson, line 61: Please type return.

src/terminal/terminal.ts
```diff
@@ -1,49 +1,112 @@
...
-    private inputSubject: BehaviorSubject<string> = new BehaviorSubject<string>(null);
-    private resizeSubject: BehaviorSubject<TerminalSize> = new BehaviorSubject<TerminalSize>({rows: 0, columns: 0});
+    private inputSubject: Subject<string> = new Subject<string>();
+    private resizeSubject: Subject<TerminalSize> = new Subject<TerminalSize>();
     private blockInput: boolean = true;
     private terminalRunningStream: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(true);
     public terminalRunning: Observable<boolean> = this.terminalRunningStream.asObservable();
 
-    constructor(private logger: Logger, configService: ConfigService, connectionId: string)
+    constructor(private logger: Logger, private configService: ConfigService, private connectionId: string, private parsedTarget: ParsedTargetString)
```
Nabeelperson, line 28: I prefer explicit assignments of all member variables but not picky about this, although it would be more consistent with the rest of the code base.

src/terminal/terminal.ts
```diff
@@ -1,49 +1,112 @@
...
 export class ShellTerminal implements IDisposable
 {
-    private websocketStream : ShellWebsocketService;
+    private shellWebsocketService : IShellWebsocketService;
+    private attached: boolean = false;
+
     // stdin
-    private inputSubject: BehaviorSubject<string> = new BehaviorSubject<string>(null);
-    private resizeSubject: BehaviorSubject<TerminalSize> = new BehaviorSubject<TerminalSize>({rows: 0, columns: 0});
+    private inputSubject: Subject<string> = new Subject<string>();
```
Nabeelperson, line 22: Why change from BehaviorSubject to Subject? I am curious, I don't know the difference very well here.

src/terminal/terminal.ts
Nabeelperson: I think this sequence is ALT+R but it is not clear if that is the case.

src/terminal/terminal.ts
tflpd: TLDR; When a subscriber subscribes to a BehSubj he will receive the initial/current value. When a subscriber subscribes to a Subj he will receive only any new values emitted from that point on. For a more detailed example look [here](https://stackoverflow.com/a/46858988/15425862)

src/terminal/terminal.ts
lipmas: I like this syntax if we are simply injecting services/variables into the class as it cuts down on boiler plate code needed. We use this syntax consistently in the web app but not as much in the ZLI.

lipmas: @Nabeelperson dont have a very strong preference here but is it ok if we leave this one?

src/terminal/terminal.ts
Nabeelperson: Good to know, thank you for explination.

src/terminal/terminal.ts
Nabeelperson: Yeah, it is fine, we can move towards this syntax for future services and constructor refactors.

src/terminal/terminal.ts
```diff
@@ -1,49 +1,112 @@
...
+        this.shellWebsocketService.shellEventData.subscribe(
+            async (shellEvent: ShellEvent) => {
+                this.logger.trace(`Got new shell event update: ${JSON.stringify(shellEvent)}`);
 
-        this.websocketStream.shellStateData.subscribe(
-            (newState: ShellState) => {
-                this.logger.trace(`Got new shell state update: ${JSON.stringify(newState)}`);
-                if (newState.start) {
+                switch(shellEvent.type) {
+                case ShellEventType.Ready:
```
sidpremkumar, line 91: nit: I think these case statements need to be indented?

src/terminal/terminal.ts
```diff
@@ -1,49 +1,112 @@
...
+                    this.logger.warn('Another active session has been detected and input has been suspended. Press ALT+R at any time to resume this terminal');
+                    this.blockInput = true;
+                    this.attached = false;
+                    break;
+                case ShellEventType.Disconnect:
+                case ShellEventType.Delete:
                     this.dispose();
+                    break;
+                default:
+                    this.logger.warn(`unhandled shell event type ${shellEvent.type}`);
```
sidpremkumar, line 109: Since the user might see this, we probs wanna capitalize the `unhandled`

sidpremkumar:
Tested: 
* `aws sts get-caller-identity`: **Passed**
```
[root@ip-172-31-80-172 ~]# aws sts get-caller-identity
{
    "Account": "238681891460", 
    "UserId": "AROATPEUMUKCGWXNHDMAE:i-070734975e2182cc9", 
    "Arn": "arn:aws:sts::238681891460:assumed-role/webshell-development-vpc-0301-161709-sid-I-IamRole-CXSHS4KP1UAS/i-070734975e2182cc9"
}
[root@ip-172-31-80-172 ~]# 
```

* `Shell connect in zli -> connect in webapp`: **Failed**
   * This left me in a broken state:
```
  ➜  zli git:(feat/shell-keysplitting) ✗ zli-d connect ssm-user@sid-keysplitting-test    
Logged in as: sid@commonwealthcrypto.com, bzero-id:619a9eb0-0ee3-4cd1-a251-b88fd75d53f4, session-id:03d633fc-22f7-4587-8fdc-7df947b3ac56
[ssm-user@ip-172-31-80-172 ~]$ ls
[ssm-user@ip-172-31-80-172 ~]$ ls
[ssm-user@ip-172-31-80-172 ~]$ vim testAnother active session has been detected and input has been suspended. Press ALT+R at any time to resume this terminal
[ssm-user@ip-172-31-80-172 ~]$ vim test
```
 * Then when I disconnected in the webapp, I just was booted off my zli shell:
  
```
[ssm-user@ip-172-31-80-172 ~]$ vim testAnother active session has been detected and input has been suspended. Press ALT+R at any time to resume this terminal
[ssm-user@ip-172-31-80-172 ~]$ vim test
[ssm-user@ip-172-31-80-172 ~]$ exit
logout

BastionZero 4/22/2021 8:20:14 PM >>> This target has gone offline or disconnected
```

* Could be something on my end, so might wanna double check that   
* _Note:_ I did not type exit or logout 
* Tried it a couple of times afterwords and it seems to be consistently broken like this

* `Vim with resizing`: **Failed**
    * It looks like we have a regression of our terminal resizing bug. I need to trigger a resize event for vim to render correctly:
image
    * Could be something on my end, so might wanna double check that

* `Ensuring that my input in the shell trigger a "sending data message" in the logs`: **Passed**

* `Slower overall feel in the zli`: **Failed**
    * Please take this with a grain of salt, this could **totally** be my computer, but on my end it feels slower
    * Please ignore if your testing does not match up with this

src/terminal/terminal.ts
lipmas: This logic is being removed in #94

lipmas:
@sidpremkumar 

> Shell connect in zli -> connect in webapp: Failed

Yeah handling all the edge cases with this flow was getting too messy so we are going to change the behavior so that connecting to an existing zli connection in the web app simply causes the zli to report an error and exit. See #94 for more details

 > Vim with resizing: Failed
 
I will dig deeper into this tomorrow and make sure we are not introducing a regression. I think before we were implicitly relying on the backend to send the terminal resize message for zli connections but with keysplitting now the client needs to send a signed resize keysplitting message in order for it to be processed by the agent.

src/terminal/terminal.ts
lipmas: Fixing in #94

src/terminal/terminal.ts
lipmas: Fixing in #94

lipmas:
@sidpremkumar 667f0f7 Added a terminal resize input that is sent as soon as the zli connects to the target. This was previously relying on the backend to send the message at startup which no longer works for keysplitting enabled targets. Tested out vim/emacs resizing and i think its working properly now when the terminal is first started.

sidpremkumar:
Retesting the tests that failed: 
* `Shell connect in zli -> connect in webapp`: **Failed (sorta)**
    * When starting a connection from zli, and then connecting from the webapp I get the expected behavior: 
        * `[ssm-user@ip-172-31-16-199 ~]$ Web App session has been detected.`
    * When starting a connection from the webapp, and then connecting from the zli. I do not get kicked from the webapp (see images)
image
image
    * Maybe I'm doing something incorrectly?
  
And both shells let me type. But the webapp one does not register any commands (i.e. nothing in the logs like `CommandEventHostedService processing new command`)
* `Vim with resizing`: **Passed**
    * Vim was set to the expected size in my terminal without any resizing

lipmas:
@sidpremkumar 
> When starting a connection from the webapp, and then connecting from the zli. I do not get kicked from the webapp (see images)

When you connect from the zli you are always creating a new connection. The zli doesnt currently allow attaching to an existing connection so its not possible to really test what you are trying to test. After connecting in the zli, if you refresh the web app you should see a new connection appear in the `cli-space` that is an entirely different connection than the one you opened in the web app initially.

sidpremkumar:
> @sidpremkumar
> 
> > When starting a connection from the webapp, and then connecting from the zli. I do not get kicked from the webapp (see images)
> 
> When you connect from the zli you are always creating a new connection. The zli doesnt currently allow attaching to an existing connection so its not possible to really test what you are trying to test. After connecting in the zli, if you refresh the web app you should see a new connection appear in the `cli-space` that is an entirely different connection than the one you opened in the web app initially.

Gotcha right yeah that makes sense, in that case its looking good on my end then.

lipmas:
> Are you saying that this should also apply to all spaces? I.e. shouldn't I see this if I try to initialize a connection not in the cli-space
 
The zli will only create new connections in the `cli-space` and there is no way to override this currently that's why you see any zli new connections show up there. In theory the cli-space is no different than any other space from the web app perspective though. So if you opened up multiple tabs in the web app with the cli-space open only the newest tab would be attached and all other tabs would be unattached.

sidpremkumar (approved):

src/oauth.service/oauth.service.ts
```diff
@@ -160,6 +161,44 @@ export class OAuthService implements IDisposable {
...
+                    if(e instanceof errors.RPError || e instanceof errors.OPError) {
+                        this.logger.error('Stale log in detected');
+                        this.logger.info('You need to log in, please run \'zli login --help\'');
+                        this.configService.logout();
+                        cleanExit(1, this.logger);
+                    } else {
+                        this.logger.error('Unexpected error during oauth refresh');
+                        this.logger.info('Please log in again');
+                        this.configService.logout();
+                        cleanExit(1, this.logger);
```
Nabeelperson, line 190: Should this be `await cleanExit()` now?

src/oauth.service/oauth.service.ts
```diff
@@ -160,6 +161,44 @@ export class OAuthService implements IDisposable {
...
+                    }
+                }
+            }
+        } else {
+            this.logger.warn('You need to log in, please run \'zli login --help\'');
+            process.exit(1);
```
Nabeelperson, line 196: Why is `cleanExit()` not used here?

src/terminal/terminal.ts
```diff
@@ -1,71 +1,149 @@
...
             },
             () => {
-                this.terminalRunningStream.error(undefined);
+                this.terminalRunningStream.error('ShellEventData subscription completed prematurely');
             }
         );
     }
 
-    public resize(resizeEvent: TerminalSize)
+    public resize(terminalSize: TerminalSize)
```
Nabeelperson, line 130: Return type please

src/terminal/terminal.ts
```diff
@@ -1,71 +1,149 @@
...
-            this.resizeSubject.next({rows: resizeEvent.rows, columns: resizeEvent.columns});
+            this.resizeSubject.next({rows: terminalSize.rows, columns: terminalSize.columns});
     }
 
-    public writeString(input: string) : void {
-        if(! this.blockInput)
+    public async writeString(input: string) : Promise<void> {
+        if(! this.blockInput) {
             this.inputSubject.next(input);
+        }
```
Nabeelperson, line 146: Why is this async now?

lipmas: no longer needs to be async

src/oauth.service/oauth.service.ts
lipmas: b6161dd

src/oauth.service/oauth.service.ts
lipmas: b6161dd

src/terminal/terminal.ts
lipmas: 16750aa

src/handlers/copy.handler.ts
```diff
@@ -31,21 +31,21 @@ export async function copyHandler(
...
-        cleanExit(1, logger);
+        await cleanExit(1, logger);
     }
 
     if(parsedTarget.type == TargetType.DYNAMIC)
     {
         logger.error('Cannot file transfer with a dynamic access config');
         logger.warn('Please create a new dynamic access target or fetch an existing one (zli lt)');
-        cleanExit(1, this.logger);
+        await cleanExit(1, this.logger);
```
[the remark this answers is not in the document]
lipmas: ok fixed it ee3676f

PR #90, opened 2021-04-12
Ensure oauth middleware is awaited in case a token refresh is needed

lipmas:
This fixes a bug where the ouath middleware wasnt being properly awaited in the pipeline causing api requests to be made before the ouath tokens had been refreshed and resulting in 401 errors. This regression was introduced in #65 when we refactored the cli driver code. 

When refresh was needed zli would just exit with the following error until user manually logged out/in:

```
/bin/zli-linux lt 
Logged in as: sebby@commonwealthcrypto.com, bzero-id:4f2c2811-e7bc-4fbb-ad1a-3837e32d1d26, session-id:6a0fd763-386e-4d68-97d8-fc21bd1aea93
Authentication Error:
Response code 401 (Unauthorized)
```

### DIFF
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -71,13 +71,13 @@ export class CliDriver
                 this.configService = initResponse.configService;
                 this.keySplittingService = initResponse.keySplittingService;
             })
-            .middleware(() => {
-                checkVersionMiddleware(this.logger);
+            .middleware(async () => {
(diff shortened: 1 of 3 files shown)


lipmas:
Fixed linter issue

Nabeelperson (approved):

sidpremkumar (approved):
whoops, thanks for fixing this

PR #91, opened 2021-04-14
Remove ssh-proxy from list of command where we skip pre-fetching

Nabeelperson:
Currently ssh-proxy is broken without this fix.


src/cli-driver.ts
```diff
@@ -294,7 +294,8 @@ export class CliDriver
 
                     // modify argv to have the targetString and targetType params
                     const targetString = argv.user + '@' + argv.host.substr(prefix.length);
-                    const parsedTarget = await disambiguateTarget('ssm', targetString, this.logger, this.dynamicConfigs, this.ssmTargets, this.sshTargets, this.envs);
+                    this.logger.warn(targetString);
```
lipmas, line 297: should probably remove this warning log

lipmas (approved):

chad-cwc (approved):

src/cli-driver.ts
Nabeelperson: 2bdd16b

PR #93, opened 2021-04-20
Add refresh OAuth id token functionality

tflpd:
This adds the ability to check whether the current OAuth id token has expired every time there is an input from the user. If it has, it refreshes before continuing.

It addresses CWC-651

In order to test this:
- Connect to a server with zli that has an agent which supports keysplitting
- Get the current Unix time
- Open with an editor the config file which can be found by `./zli-macos config`
- Find the `expires_at` property
- Copy the current Unix time. Edit that time and add a couple seconds (usually I add 40s). Save the file.
- Type input in the zli-dev. If you want to verify that the token is getting refreshed look for "refresh" in the logs or monitor the value in the config json file

### DIFF
--- a/src/config.service/zli-auth-config.service.ts
+++ b/src/config.service/zli-auth-config.service.ts
@@ -5,11 +5,15 @@ import { OAuthService } from '../../src/oauth.service/oauth.service';
 
 export class ZliAuthConfigService implements AuthConfigService {
 
+    private oauth: OAuthService;
+
     constructor(
         private configService: ConfigService,
         private logger: Logger
     )
-    {}
+    {
+        this.oauth = new OAuthService(this.configService, this.logger);
+    }
 
     getServiceUrl() {
         return this.configService.serviceUrl() + 'api/v1/';
@@ -20,7 +24,6 @@ export class ZliAuthConfigService implements AuthConfigService {
     }
 
     async getIdToken() {
-        const oauth = new OAuthService(this.configService, this.logger);
-        return oauth.getIdToken();
+        return await this.oauth.getIdToken();
     }
 }
--- a/src/config.service/zli-auth-config.service.ts
+++ b/src/config.service/zli-auth-config.service.ts
@@ -1,10 +1,13 @@
 import { AuthConfigService } from '../../webshell-common-ts/auth-config-service/auth-config.service';
 import { ConfigService } from './config.service';
+import { Logger } from '../logger.service/logger';
+import { OAuthService } from '../../src/oauth.service/oauth.service';
 
 export class ZliAuthConfigService implements AuthConfigService {
 
     constructor(
-        private configService: ConfigService
+        private configService: ConfigService,
+        private logger: Logger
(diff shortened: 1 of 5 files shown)


src/middlewares/oauth-middleware.ts
```diff
@@ -6,7 +6,7 @@ import { cleanExit } from '../../src/handlers/clean-exit.handler';
 
 export async function oauthMiddleware(configService: ConfigService, logger: Logger) : Promise<void> {
 
-    const ouath = new OAuthService(configService, logger);
+    const oauth = new OAuthService(configService, logger);
```
lipmas, line 9: Why not replace all this logic in oauthMiddleware with `oauthService.getIdToken()` so we dont have to duplicate the logic?

src/oauth.service/oauth.service.ts
```diff
@@ -160,6 +161,45 @@ export class OAuthService implements IDisposable {
...
+                }
+            }
+        } else {
+            this.logger.warn('You need to log in, please run \'zli login --help\'');
+            process.exit(1);
+        }
+
+
+        return this.configService.getAuth();
+    }
```
Nabeelperson, line 201: It seems to me this is a copy-paste of the oauth middleware. What you are suggesting it to have the oauthService be responsible for refresh logic and not the middleware. You should probably refactor the middleware to use this new function, even if the return it not needed in the middleware. 

https://github.com/bastionzero/zli/blob/6f8b53d27f974dbc0bb35652770a8b6b05f64b79/src/middlewares/oauth-middleware.ts

src/oauth.service/oauth.service.ts
```diff
@@ -160,6 +161,45 @@ export class OAuthService implements IDisposable {
         return userInfo;
     }
 
+    // Returns the current OAuth idtoken. Refreshes it before returning if expired
+    async getIdToken() {
```
Nabeelperson, line 165: Please add the return type to the function's definition.

src/config.service/zli-auth-config.service.ts
```diff
@@ -17,6 +20,7 @@ export class ZliAuthConfigService implements AuthConfigService {
     }
 
     async getIdToken() {
-        return this.configService.getAuth();
+        const oauth = new OAuthService(this.configService, this.logger);
```
Nabeelperson, line 23: You should not construct the OAuthService ever time we need the IdToken, probably best to pass this into the constructor of the ZliAuthConfigService. Unless you have a reason we need to reconstruct on call.

tflpd: Moved this to be constructed in the middleware's constructor which should save us some calls. I should probably be able to make that an argument as well, I will just have to add this argument to a whole chain of function calls to reach the point we have access to the OAuthService. I should be able to do that, I just have a small concern that this is not ideal..? Let me know, maybe this is not something I should actually be concerned about

lipmas (approved):

PR #98, opened 2021-04-27
Disambiguate targets now ignores terminated and errored targets

Nabeelperson:
## Description of the change

In this PR: 
 - Disambiguate targets now ignores terminated and errored targets
 - New Status flag for lt command (-s online offline etc)
 - Refactor of silent flag to quiet

Testing:
 - Autodiscover two new targets with the exact same name (Manual Name mode)
 - Edit one of the targets in DB to be in Error state
 - Try to connect using `zli connect user@target` and you should be able to connect to the one listed as Online
 - Old zli will give you "Multiple targets detected with same name error"

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-653

## Potential Security Impacts

None, this is a client side change of UX.

### Development

- [x] Do all tests related to the changed code pass in development?
- [x] Have you ensured that at least one other person has tested your code?


src/cli-driver.ts
```diff
@@ -328,7 +338,7 @@ export class CliDriver
             )
             .option('configName', {type: 'string', choices: ['prod', 'stage', 'dev'], default: this.envMap['configName'], hidden: true})
             .option('debug', {type: 'boolean', default: false, describe: 'Flag to show debug logs'})
-            .option('silent', {alias: 's', type: 'boolean', default: false, describe: 'Silence all zli messages'})
+            .option('quiet', {alias: 'q', type: 'boolean', default: false, describe: 'Silence all zli messages, only returns command output'})
```
lipmas, line 341: This `-s` flag is used in ssh-proxy-config handler when generating the ProxyCommand config so this is going to be a breaking change unless you allow both.

https://github.com/bastionzero/zli/blob/1a9fa233630c6ceefa404234102660df6b014672/src/handlers/ssh-proxy-config.handler.ts#L20

src/handlers/list-targets.handler.ts
```diff
@@ -39,12 +41,32 @@ export async function listTargetsHandler(
         allTargets = allTargets.filter(t => t.type === targetType);
     }
 
+    if(!! argv.status) {
+        const statusArray: string[] = argv.status;
+
+        if(statusArray.length < 1) {
+            logger.warn('Status filter flag passed with no arguments, please indicate at least one status');
+            cleanExit(1, logger);
```
lipmas, line 49: missing await

src/handlers/list-targets.handler.ts
```diff
@@ -39,12 +41,32 @@ export async function listTargetsHandler(
...
+        }
+
+        let targetStatusFilter: SsmTargetStatus[] = _.map(statusArray, (s: string) => parseTargetStatus(s)).filter(s => s); // filters out undefined
+        targetStatusFilter = _.uniq(targetStatusFilter);
+
+        if(targetStatusFilter.length < 1) {
+            logger.warn('Status filter flag passed with no valid arguments, please indicate at least one valid status');
+            cleanExit(1, logger);
```
lipmas, line 57: missing await

src/handlers/list-targets.handler.ts
```diff
@@ -1,12 +1,14 @@
...
-    getTableOfTargets
+    getTableOfTargets,
+    parseTargetStatus
 } from '../utils';
 import { Logger } from '../logger.service/logger';
 import { EnvironmentDetails } from '../http.service/http.service.types';
 import { cleanExit } from './clean-exit.handler';
-import { TargetSummary } from '../types';
+import { SsmTargetStatus, TargetSummary, TargetType } from '../types';
+import _ from 'lodash';
```
lipmas, line 11: can we change this to  explicitly import functions from lodash that we use instead of importing the entire namespace.

Nabeelperson: Yes, although this is an issue across the entire repo, we should probably open a ticket and just refactor all of the lodash uses in one shot.

src/handlers/list-targets.handler.ts
Nabeelperson: d9e4ec3

src/handlers/list-targets.handler.ts
Nabeelperson: d9e4ec3

src/cli-driver.ts
Nabeelperson: d9e4ec3

src/utils.ts
```diff
@@ -130,17 +147,20 @@ export async function disambiguateTarget(
 
     let zippedTargets = _.concat(await ssmTargets, await sshTargets, await dynamicConfigs);
 
+    // Filter out Error and Terminated SSM targets
+    zippedTargets = _.filter(zippedTargets, t => t.type !== TargetType.SSM || (t.status !== SsmTargetStatus.Error && t.status !== SsmTargetStatus.Terminated));
```
sidpremkumar, line 151: Maybe I'm understanding this incorrectly, but does this mean if the user types `zli connect ssm-user@test` and `test` is offline, they will just see an error saying unable to find a matching target? 

I guess what I'm wondering is how does the fact the target they are trying to connect to is offline/errored bubble up to the user?

Nabeelperson: The bug I am resolving in this PR is as follows:
A target is attempted to be produced by a pipeline (theoretically) and fails to initialize properly the first time, but the autodiscovery script has run and reported an error'ed target. The pipeline then runs again and now succeeds to deploy the target. Both the error'ed target and online target now have the same name and the disambiguation logic then throws a "please use GUID" error since it is not filtering for error'ed and terminated targets.

src/cli-driver.ts
```diff
@@ -204,7 +214,7 @@ export class CliDriver
                         )
                         .example('lt -t ssm', 'List all SSM targets only')
                         .example('lt -i', 'List all targets and show unique ids')
-                        .example('lt -e prod --json --silent', 'List all targets targets in prod, output as json, pipeable');
+                        .example('lt -e prod --json --quiet', 'List all targets targets in prod, output as json, pipeable');
```
Nabeelperson, line 217: Revert this to --silent

src/utils.ts
sidpremkumar: thanks for clarifying that makes sense.

sidpremkumar (approved):

lipmas (approved):

PR #99, opened 2021-04-27
Fix/close ssh tunnels on logout [CWC-628]

tflpd:
## Description of the change

This fixes the bug when there are multiple instances of the zli running and one of them logs out, the others would not pick up the change and close any open tunnels.

The suggested fix is the following:
When a zli logs out, before exiting, it lists all the open cli connections and sends for each one of them a close connection request to the back end. The back end then closes them from its side and this way all the open tunnels are getting closed as well.

In order to test this:
- Open a terminal. Login to zli. Connect to a server from this terminal.
- Open a second terminal. Execute logout.
- Check both terminals. Both should be disconnected and logged out without any error messages. Make sure you look in the logs of both the zli and the webapp please, we have noticed some discrepancies and we want to make sure those do not show up anymore.
- Extra test : try logging in and connecting again to the server.

## Type of change

- [X] Bug fix (non-breaking change that fixes an issue)

## Related JIRA tickets

Relates to JIRA: CWC-628

### Development


src/handlers/logout.handler.ts
```diff
@@ -1,9 +1,37 @@
...
+    // If there is no session there are no connections to close, we can exit
+    if(cliSpace.length === 0) {
+        configService.logout();
+        logger.info('Logout successful');
+        await cleanExit(0, logger);
+        return;
+    } else {
+        // there should only be 1 active 'cli-space' session
+        cliSession = cliSpace.pop();
+    }
```
Nabeelperson, line 29: This should probably be made into a utility function as the logic is repeated in connect-handler as well.

EthanHeilman:
In the PR description describe not just how to replicate and test the problem but what the proposed solution in the PR is.

src/handlers/logout.handler.ts
```diff
@@ -1,9 +1,37 @@
+import { ConnectionState, SessionDetails } from '../../src/http.service/http.service.types';
+import { ConnectionService, SessionService } from '../../src/http.service/http.service';
 import { ConfigService } from '../config.service/config.service';
 import { Logger } from '../logger.service/logger';
+import { SessionState } from '../types';
 import { cleanExit } from './clean-exit.handler';
+import _ from 'lodash';
```
sidpremkumar, line 7: Similar to what Sebby said in Nabeels PR, we should probably explicitly import what we need from `lodash`

src/handlers/logout.handler.ts
```diff
@@ -1,9 +1,37 @@
...
+        configService.logout();
+        logger.info('Logout successful');
+        await cleanExit(0, logger);
+        return;
+    } else {
+        // there should only be 1 active 'cli-space' session
+        cliSession = cliSpace.pop();
+    }
+    const connectionService = new ConnectionService(configService, logger);
+    logger.debug('Closing any open cli connections');
```
sidpremkumar, line 31: Should this be a `logger.info`? Thinking the user might be confused if they log out of the zli and their connections close, they might not understand what closed them. 

Not sure though.

tflpd:
> In the PR description describe not just how to replicate and test the problem but what the proposed solution in the PR is.

@EthanHeilman added

src/handlers/logout.handler.ts
tflpd: The logic is indeed very similar, however the `cliSpace.length === 0` case is different here from the connect handler. I guess I can add something like a flag and apply the right logic every time - do we want that? I don’t really like this idea

src/handlers/logout.handler.ts
tflpd: @Nabeelperson what's your take in this?

src/handlers/logout.handler.ts
tflpd: Hm, okay, I think I agree with you in this. Let's see what the others say and if they agree as well I will add it

src/handlers/logout.handler.ts
Nabeelperson: Well the logic here can return you `undefined` or the `cliSession` object, then the caller can choose how to handle the returned value.

src/handlers/logout.handler.ts
Nabeelperson: @sidpremkumar @tflpd we have this issue across the entire code base. When we decide to fix this we should do it in one shot across the entire code base and then make it part of the our code standards for thoum to not import lodash in its entirety.

src/handlers/logout.handler.ts
Nabeelperson: Info is good, should probably say `Closing all other instances of zli` or something along those lines.

src/handlers/logout.handler.ts
tflpd: Yes but the if case in the connect handler should not have the `cleanExit` in line 24 which the logout needs, right?

src/handlers/connect.handler.ts
```diff
@@ -115,19 +115,7 @@ export async function connectHandler(
...
-            // if connection not already closed
-            if(conn.state == ConnectionState.Open)
-                await connectionService.CloseConnection(connectionId);
-
-            logger.debug('Connection closed');
-
-            if(error)
-                await cleanExit(1, logger);
-
-            await cleanExit(0, logger);
```
lipmas, line 130: Why are we removing this logic? Now when a user exits a `zli connect` command normally (typing exit, logout, or ctrl+d) the connection would no longer be cleaned up automatically and will have to be closed through the web app or logging out of the zli.

Is this change in behavior what we want @Nabeelperson @tflpd ?

lipmas (changes requested):
It seems like this PR is only addressing closing existing connections opened with `zli connect` when the user logs out...but it wont have any effect on any existing open ssh tunnels. In order to close any existing ssh tunnels that are opened with the zli i think you would need to catch the logout event in the ssh-proxy handler https://github.com/bastionzero/zli/blob/master/src/handlers/ssh-proxy.handler.ts

src/handlers/connect.handler.ts
Nabeelperson: Good catch, that was an oversight when I made the changes to just get the logout flow working. @tflpd may you revert this change and see how to address the ssh tunnel logout cases?

src/handlers/logout.handler.ts
Nabeelperson: Your observation is correct. The function I am proposing returns a cli space, does not do anything about closing thoum, that is the responsiblity of the caller to that function.

tflpd:
Following up [CWC-628](https://commonwealthcrypto.atlassian.net/browse/CWC-628) comments this feature is not the requested solution, thus this PR is being closed

PR #105, opened 2021-05-10
Attaching to existing open zli connections [CWC-667]

tflpd:
## Description of the change

This adds functionality to manage zli open connections. This includes three new zli commands `list-connections` which lists any open zli connections, `attach` which lets the user re-attach to an existing open zli connection and `close` which lets the user close an existing open zli connection.

Please make sure you test this against both key-splitting enabled and key-splitting disabled agents, since the flow is different there might exist bugs on the one that do not exist on the other.

To test this:
- Open a new connection to a target with zli
- Type `exit` to exit that connection. Note that this exit does not close anymore the connection
- Using `lc` (list-connections`) get a list of any open zli connections
- Using `attach` reconnect to any of those. They should work as you never exited
- "Stress test" it by attaching to the same connection by different zli/webapp clients

## Type of change

- [X] New feature (non-breaking change that adds functionality)

## Related JIRA tickets

Relates to JIRA: CWC-667

## Potential Security Impacts

For non-keysplitting agents, attaching to an existing connection will not remove any other attached clients. This is not a real threat but it should still be noted.

### Development

### DIFF
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -101,7 +101,6 @@ export class CliDriver
                 this.ssmTargets = fetchDataResponse.ssmTargets;
                 this.sshTargets = fetchDataResponse.sshTargets;
                 this.envs = fetchDataResponse.envs;
-                this.cliSpaceId = fetchDataResponse.cliSpaceId;
             })
             .command(
                 'login <provider>',
@@ -150,7 +149,7 @@ export class CliDriver
                 async (argv) => {
                     const parsedTarget = await disambiguateTarget(argv.targetType, argv.targetString, this.logger, this.dynamicConfigs, this.ssmTargets, this.sshTargets, this.envs);
 
-                    await connectHandler(this.configService, this.logger, this.mixpanelService, parsedTarget, this.cliSpaceId);
+                    await connectHandler(this.configService, this.logger, this.mixpanelService, parsedTarget);
                 }
             )
             .command(
@@ -168,7 +167,7 @@ export class CliDriver
                         this.logger.error(`Passed connection id ${argv.connectionId} is not a valid Guid`);
                         await cleanExit(1, this.logger);
                     }
-                    await attachHandler(this.configService, this.logger, argv.connectionId, this.cliSpaceId);
+                    await attachHandler(this.configService, this.logger, argv.connectionId);
                 }
             )
             .command(
@@ -186,7 +185,7 @@ export class CliDriver
                         this.logger.error(`Passed connection id ${argv.connectionId} is not a valid Guid`);
                         await cleanExit(1, this.logger);
                     }
-                    await closeConnectionHandler(this.configService, this.logger, argv.connectionId, this.cliSpaceId);
+                    await closeConnectionHandler(this.configService, this.logger, argv.connectionId);
                 }
             )
             .command(
@@ -280,7 +279,7 @@ export class CliDriver
                         .example('lc --json', 'List all open zli connections, output as json, pipeable');
                 },
                 async (argv) => {
-                    await listConnectionsHandler(argv, this.configService, this.logger, this.ssmTargets, this.sshTargets, this.cliSpaceId);
+                    await listConnectionsHandler(argv, this.configService, this.logger, this.ssmTargets, this.sshTargets);
                 }
             )
             .command(
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -341,7 +341,7 @@ export class CliDriver
                         .positional('connectionId', {
                             type: 'string',
                         })
-                        .example('attach d5b264c7-534c-4184-a4e4-3703489cb917', 'attach example, unique connection id')
+                        .example('attach d5b264c7-534c-4184-a4e4-3703489cb917', 'attach example, unique connection id');
                 },
                 async (argv) => {
                     await attachHandler(this.configService, this.logger, argv.connectionId);
@@ -355,7 +355,7 @@ export class CliDriver
                         .positional('connectionId', {
                             type: 'string',
                         })
-                        .example('close d5b264c7-534c-4184-a4e4-3703489cb917', 'close example, unique connection id')
+                        .example('close d5b264c7-534c-4184-a4e4-3703489cb917', 'close example, unique connection id');
                 },
                 async (argv) => {
                     await closeConnectionHandler(this.configService, this.logger, argv.connectionId);
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -47,6 +47,7 @@ export class CliDriver
     private ssmTargets: Promise<TargetSummary[]>;
     private dynamicConfigs: Promise<TargetSummary[]>;
     private envs: Promise<EnvironmentDetails[]>;
+    private cliSpaceId: Promise<string>;
 
     // use the following to shortcut middleware according to command
     private noOauthCommands: string[] = ['config', 'login', 'logout'];
@@ -100,6 +101,7 @@ export class CliDriver
                 this.ssmTargets = fetchDataResponse.ssmTargets;
                 this.sshTargets = fetchDataResponse.sshTargets;
                 this.envs = fetchDataResponse.envs;
+                this.cliSpaceId = fetchDataResponse.cliSpaceId;
             })
             .command(
                 'login <provider>',
@@ -148,7 +150,7 @@ export class CliDriver
                 async (argv) => {
                     const parsedTarget = await disambiguateTarget(argv.targetType, argv.targetString, this.logger, this.dynamicConfigs, this.ssmTargets, this.sshTargets, this.envs);
 
-                    await connectHandler(this.configService, this.logger, this.mixpanelService, parsedTarget);
+                    await connectHandler(this.configService, this.logger, this.mixpanelService, parsedTarget, this.cliSpaceId);
                 }
             )
             .command(
@@ -166,7 +168,7 @@ export class CliDriver
                         this.logger.error(`Passed connection id ${argv.connectionId} is not a valid Guid`);
                         await cleanExit(1, this.logger);
                     }
-                    await attachHandler(this.configService, this.logger, argv.connectionId);
+                    await attachHandler(this.configService, this.logger, argv.connectionId, this.cliSpaceId);
                 }
             )
             .command(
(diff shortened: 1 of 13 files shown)


src/utils.ts
```diff
@@ -128,6 +128,26 @@ export function getTableOfTargets(targets: TargetSummary[], envs: EnvironmentDet
     return table.toString();
 }
 
+export function getTableOfConnections(connections: ConnectionSummary[]) : string
+{
+    const connIdLength = max(connections.map(c => c.id.length).concat(16));
```
Nabeelperson, line 133: This will always be 36 since they are all GUID are 36 characters.

src/utils.ts
```diff
@@ -128,6 +128,26 @@ export function getTableOfTargets(targets: TargetSummary[], envs: EnvironmentDet
...
+    const header: string[] = ['Connection ID'];
+    const columnWidths = [connIdLength + 2];
+
+    // ref: https://github.com/cli-table/cli-table3
```
Nabeelperson, line 137: Delete duplicate ref

src/utils.ts
```diff
@@ -128,6 +128,26 @@ export function getTableOfTargets(targets: TargetSummary[], envs: EnvironmentDet
...
+    connections.forEach(connection => {
+        const row = [connection.id];
+
+        table.push(row);
+    }
+    );
+
+    return table.toString();
+
+}
```
Nabeelperson, line 149: Please adjust this whitespace.

src/terminal/terminal.ts
```diff
@@ -26,26 +25,24 @@ export class ShellTerminal implements IDisposable
     private terminalRunningStream: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(true);
     public terminalRunning: Observable<boolean> = this.terminalRunningStream.asObservable();
 
-    constructor(private logger: Logger, private configService: ConfigService, private connectionId: string, private parsedTarget: ParsedTargetString)
+    constructor(private logger: Logger, private configService: ConfigService, private connectionId: string, private targetType: TargetType, private serverId: string)
```
Nabeelperson, line 28: Why did you remove `parsedTarget` here? It contains all the fields you are now explicitly passing in.

src/http.service/http.service.types.ts
```diff
@@ -77,6 +77,7 @@ export interface ListSsmTargetsRequest
 }
 
 export interface SsmTargetSummary {
+    // The following is serverId
```
Nabeelperson, line 80: I don't think this is a necessary comment.

src/handlers/create-shell.handler.ts
```diff
@@ -0,0 +1,61 @@
+import { ConfigService } from '../../src/config.service/config.service';
+import { Logger } from '../../src/logger.service/logger';
+import { ShellTerminal } from '../../src/terminal/terminal';
+import { TargetType } from '../types';
+import { cleanExit } from './clean-exit.handler';
+import termsize from 'term-size';
+
+export async function createShellHandler(
```
Nabeelperson, line 8: `createAndRunShell()` is probably a more appropriate name for this, as this is not a cli-driver handler method.

src/handlers/close-connection.handler.ts
```diff
@@ -0,0 +1,24 @@
+import { ConfigService } from '../../src/config.service/config.service';
+import { ConnectionService } from '../../src/http.service/http.service';
+import { ConnectionState } from '../../src/http.service/http.service.types';
+import { Logger } from '../../src/logger.service/logger';
+import { cleanExit } from './clean-exit.handler';
+
+export async function closeConnectionHandler(
```
Nabeelperson, line 7: Make sure to check if the connectionId is from the cli-space, otherwise you should not be closing the connection.

src/cli-driver.ts
```diff
@@ -322,6 +325,42 @@ export class CliDriver
...
+            .command(
+                ['list-connections', 'lc'],
+                'List all open connections',
+                () => {},
+                async () => {
+                    await listConnectionsHandler(this.configService, this.logger);
+                }
+            )
+            .command(
+                'attach <connectionId>',
```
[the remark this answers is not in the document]
Nabeelperson: You should have a regex check before passing in the connectionId to the handler, to make sure the argument is well formed. 

Probably this regex check should be made into a function and used in both places
https://github.com/bastionzero/zli/blob/feat/shell-reattachment/src/utils.ts#L77

Similar should be done for close connection command.

src/utils.ts
tflpd: @lipmas Lint was not detecting this as an issue, should we add a rule?

src/terminal/terminal.ts
tflpd: This was needed because when attaching we do not have a `parsedTarget` struct to pass here, we have only the `connectionId` and the `targetType`

src/handlers/list-connections.handler.ts
```diff
@@ -1,15 +1,19 @@
...
 
 export async function listConnectionsHandler(
     configService: ConfigService,
-    logger: Logger
+    logger: Logger,
+    ssmTargets: Promise<TargetSummary[]>,
+    sshTargets: Promise<TargetSummary[]>
 ){
+    // await and concatenate
+    const allTargets = [...await ssmTargets, ...await sshTargets];
```
Nabeelperson, line 16: You should probably only await on this after checking the length of openConnections.

src/utils.ts
```diff
@@ -128,21 +133,19 @@ export function getTableOfTargets(targets: TargetSummary[], envs: EnvironmentDet
...
+    const columnWidths = [targetNameLength + 2, connIdLength + 2, 20];
 
-    // ref: https://github.com/cli-table/cli-table3
     const table = new Table({ head: header, colWidths: columnWidths });
-
+    const dateOptions = {year: '2-digit', month: 'numeric', day: 'numeric', hour:'numeric', minute:'numeric', hour12: true};
     connections.forEach(connection => {
-        const row = [connection.id];
-
+        const row = [allTargets.filter(t => t.id == connection.serverId).pop().name, connection.id, new Date(connection.timeCreated).toLocaleString('en-US', dateOptions as any)];
```
Nabeelperson, line 146: Does mike agree about the order of information here?

Nabeelperson (commented):
Are we including a `--json` flag for list-connections?

src/utils.ts
tflpd: Yes

src/handlers/attach.handler.ts
```diff
@@ -0,0 +1,20 @@
...
+import { createAndRunShell } from '../../src/shell-utils';
+
+export async function attachHandler(
+    configService: ConfigService,
+    logger: Logger,
+    connectionId: string
+){
+    const connectionService = new ConnectionService(configService, logger);
+    const connectionSummary = await connectionService.GetConnection(connectionId);
+    if (connectionSummary.state !== ConnectionState.Open){
```
lipmas, line 15: Also add a check here for connection belonging to cli-space?

src/shell-utils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    // space names are not unique, make sure to find the latest active one
+    const cliSpace = listSessions.sessions.filter(s => s.displayName === 'cli-space' && s.state == SessionState.Active); // TODO: cli-space name can be changed in config
+
+    // maybe make a session
+    let cliSessionId: string;
+    if(cliSpace.length === 0) {
+        cliSessionId =  await sessionService.CreateSession('cli-space');
+    } else {
+        // there should only be 1 active 'cli-space' session
```
lipmas, line 80: can we log a warning if there is more than 1 active cli-space

src/handlers/list-connections.handler.ts
```diff
@@ -0,0 +1,61 @@
...
+    const cliSpace = listSessions.sessions.filter(s => s.displayName === 'cli-space' && s.state == SessionState.Active); // TODO: cli-space name can be changed in config
+
+    let cliSessionId: string;
+    if(cliSpace.length === 0) {
+        logger.info('There is no cli session available');
+        await cleanExit(0, logger);
+    } else {
+        // there should only be 1 active 'cli-space' session
+        cliSessionId = cliSpace.pop().id;
+    }
```
lipmas, line 30: use common `getCliSpaceId` function here

tflpd: I should not. There is a small difference. The `getCliSpaceId` creates a session if there is none available. However in here we do not want to create one, we want to exit if none is available

lipmas: Ah i see ok. Then maybe you could just split `getCliSpaceId` and `createCliSpace` into 2 separate functions so that we could reuse the `getCliSpaceId` code here.

src/handlers/middleware.handler.ts
```diff
@@ -48,11 +51,14 @@ export function fetchDataMiddleware(configService: ConfigService, logger: Logger
 
     const envs = envService.ListEnvironments();
 
+    const cliSpaceId = getCliSpaceId(sessionService, logger);
```
lipmas, line 54: I dont think we should be continuing to use this patten of pushing data retrieval to this middleware. This increases the number of API calls for every single zli command. We are already starting to see some serious performance issues when running stress tests where many concurrent zli commands are run and each one currently results in 4 API calls just from this middleware layer.

Ideally we should load just the data that is needed in the beginning of each command handler. We can have the data retrieval functions be in common util methods and have each handler load what data it needs in parallel by combining all the methods in a single `await Promise.All()` instead of running all of them in series. 

cc @Nabeelperson If you agree we can open up another issue to refactor the existing code in a separate PR but lets at least not add more middleware API calls here.

src/terminal/terminal.ts
```diff
@@ -97,16 +93,19 @@ export class ShellTerminal implements IDisposable
                 case ShellEventType.Start:
                     this.blockInput = false;
                     this.terminalRunningStream.next(true);
+                    // Trigger resize to force the terminal to refresh the output
+                    const tempTerminalSize : TerminalSize = {rows: this.currentTerminalSize.rows + 1, columns: this.currentTerminalSize.columns + 1};
+                    if(! this.blockInput)
```
lipmas, line 98: nit: dont need this if because we set `this.blockInput = false;` on line 94

src/handlers/middleware.handler.ts
Nabeelperson: Yes, I agree we should move away from the greedy fetching. Yes lets open up another issue. I don't mind leaving this code for cliSpaceId where it is for now though since the refactor is coming anyways.

Nabeelperson:
For the `--json` flag, when there are no connections to return, it should return an empty json.

src/handlers/middleware.handler.ts
lipmas: created #114

PR #108, opened 2021-05-11
(CWC-705) Fix Dynamic Access Target Connect

lipmas:
## Description of the change

This fixes a bug where we were not using the ssm target id and instead using the DAT id when sending a request to get ssm target info. The ssm target info is needed in order to determine which agent version the target is running.

Fixes #107

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-705

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,11 @@
 {
   "name": "zli",
-  "version": "4.15.2",
+  "version": "4.17.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "name": "zli",
-      "version": "4.15.2",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -1,12 +1,11 @@
 {
   "name": "zli",
-  "version": "4.15.2",
+  "version": "4.17.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "name": "zli",
```
sidpremkumar, line 8: don't think we want to get ride of this

PR #112, opened 2021-05-14
Handle non existent ssh proxy target [CWC-592]

tflpd:
## Description of the change

Added error handling code for the case that the target of an ssh tunnel does not exist

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-592

### Development

### DIFF
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -27,7 +27,6 @@ import { Dictionary, includes } from 'lodash';
 import yargs from 'yargs';
 import { cleanExit } from './handlers/clean-exit.handler';
 import { autoDiscoveryScriptHandler } from './handlers/autodiscovery-script-handler';
-import { OAuthService } from './oauth.service/oauth.service';
 
 
 export class CliDriver
@@ -81,11 +80,6 @@ export class CliDriver
             .middleware(async (argv) => {
                 if(includes(this.noOauthCommands, argv._[0]))
                     return;
-                const oauth = new OAuthService(this.configService, this.logger);
-                if(argv._[0] == 'ssh-proxy' && !oauth.isAuthenticated()){
-                    this.logger.error('You need to log in, please run \'zli login --help\'');
-                    await cleanExit(1, this.logger);
-                }
                 await oAuthMiddleware(this.configService, this.logger);
             })
             .middleware(async (argv) => {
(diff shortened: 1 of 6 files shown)


Nabeelperson (approved):

Nabeelperson:
Version bump please.

src/cli-driver.ts
```diff
@@ -80,6 +81,11 @@ export class CliDriver
             .middleware(async (argv) => {
                 if(includes(this.noOauthCommands, argv._[0]))
                     return;
+                const oauth = new OAuthService(this.configService, this.logger);
+                if(argv._[0] == 'ssh-proxy' && !oauth.isAuthenticated()){
+                    this.logger.error('You need to log in, please run \'zli login --help\'');
+                    await cleanExit(1, this.logger);
+                }
```
Nabeelperson, line 88: Shouldn't this be after the oauthMiddleware call in case the token gets refreshed before checking for auth validity?

src/handlers/middleware.handler.ts
```diff
@@ -86,7 +86,9 @@ export async function oAuthMiddleware(configService: ConfigService, logger: Logg
 export async function initMiddleware(argv: any) {
     // Configure our logger
     const loggerConfigService = new LoggerConfigService(<string> argv.configName);
-    const logger = new Logger(loggerConfigService, !!argv.debug, !!argv.silent);
+    // When in a tunnel the zli does not have control of the stdio
+    const isStdEnabled = argv._[0] != 'ssh-proxy';
```
Nabeelperson, line 90: I don't like the use of a static string to determine if a standard stream is available to us, I believe something in the process API should help us here.

src/cli-driver.ts
tflpd: It was not needed at that point after binding the log.error to stderr, removed

src/handlers/middleware.handler.ts
```diff
@@ -86,10 +86,7 @@ export async function oAuthMiddleware(configService: ConfigService, logger: Logg
 export async function initMiddleware(argv: any) {
     // Configure our logger
     const loggerConfigService = new LoggerConfigService(<string> argv.configName);
-    // When in a tunnel the zli does not have control of the stdio
-    const isStdEnabled = argv._[0] != 'ssh-proxy';
-    const logger = new Logger(loggerConfigService, !!argv.debug, !!argv.silent, isStdEnabled);
-
+    const logger = new Logger(loggerConfigService, !!argv.debug, !!argv.silent, !!process.stdout.isTTY);
```
Nabeelperson, line 89: Add a comment about why `!!process.stdout.isTTY` works where please.

Nabeelperson (approved):

PR #118, opened 2021-05-26
Fix connect command to work with dynamic access targets

lipmas:
## Description of the change

This fixes a bug where we were not using the ssm target id and instead using the DAT id when sending a request to get ssm target info. The ssm target info is needed in order to determine which agent version the target is running.

This was originally fixed in https://github.com/bastionzero/zli/pull/108 but got reverted with some refactors that happened to the connect handler.

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-XXX

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
   "name": "zli",
-  "version": "4.17.0",
+  "version": "4.20.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "version": "4.17.0",
+      "version": "4.20.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@microsoft/signalr": "^5.0.0",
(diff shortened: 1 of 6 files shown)


sidpremkumar (approved):
Tested and working on my end, only thing might be to bump version

Nabeelperson (approved):

tflpd (commented):
This way we will be making an extra `GetConnection()` request on `zli attach`. I would suggest to keep the previous flow and just add the extra necessary `GetConnection()` to the `connectHandler` in order to pass down the stack the correct `targetId`. Also please add a comment why this is necessary for DATs, otherwise someone will definitely mess this up again as I did, there is no indication this should be avoided.

Nabeelperson (approved):
Please bump version.

tflpd (approved):

PR #124, opened 2021-07-15
Add Kube Daemon/Agent to Zli

sidpremkumar:
## Description of the change

This adds the kube Daemon and Agent to the Zli. Please see the linked Jira ticket for more information. 

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-907

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


bctl-go-daemon/bctl/daemon/daemon.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		os.Exit(1)
+	}
+	logger.AddDaemonVersion(version)
+	dcLogger := logger.GetDatachannelLogger()
+
+	logger.Info(fmt.Sprintf("Opening websocket to Bastion: %s", serviceUrl))
+	startDatachannel(dcLogger)
+
+	select {} // sleep forever?
```
ymarcus93, line 45: So how is daemon killed? Does `zli` just `kill -9` the process if the user runs `disconnect`?

ymarcus93: We might want to refactor this to use `context` or https://pkg.go.dev/gopkg.in/tomb.v2 so that if an interrupt signal is passed, we gracefully shut down the daemon by cancelling the context or `tomb`--this allows goroutines listening on the context done channel or tomb, to do some cleanup.

If we just `kill -9` from the `zli`, we're not giving the chance for graceful cleanup...but maybe that's what we want for the daemon? Not sure...

bctl-go-daemon/bctl/daemon/daemon.go
```diff
@@ -0,0 +1,127 @@
...
+	params := make(map[string]string)
+	params["session_id"] = sessionId
+	params["assume_role"] = assumeRole
+	params["assume_cluster_id"] = assumeClusterId
+	params["environment_id"] = environmentId
+
+	dataChannel, _ := dc.NewDataChannel(logger, configPath, assumeRole, serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect)
+
+	// TODO: Integrate this with existing messaging
+	time.Sleep(3 * time.Second)
```
ymarcus93, line 63: What race condition are we running into without this `sleep`?

bctl-go-daemon/bctl/daemon/daemon.go
```diff
@@ -0,0 +1,127 @@
...
+
+	dc "bastionzero.com/bctl/v1/bctl/daemon/datachannel"
+	wsmsg "bastionzero.com/bctl/v1/bzerolib/channels/message"
+	lggr "bastionzero.com/bctl/v1/bzerolib/logger"
+)
+
+// Declaring flags as package-accesible variables
+var (
+	sessionId, authHeader, assumeRole, assumeClusterId, serviceUrl           string
+	daemonPort, localhostToken, environmentId, certPath, keyPath, configPath string
```
ymarcus93, line 19: How is `cert` generated? From the `zli`?

bctl-go-daemon/bctl/daemon/plugin/kube/kube.go
```diff
@@ -0,0 +1,277 @@
...
+	rest "bastionzero.com/bctl/v1/bctl/daemon/plugin/kube/actions/restapi"
+	lggr "bastionzero.com/bctl/v1/bzerolib/logger"
+	plgn "bastionzero.com/bctl/v1/bzerolib/plugin"
+	smsg "bastionzero.com/bctl/v1/bzerolib/stream/message"
+
+	"github.com/google/uuid"
+)
+
+const (
+	securityToken = "++++"
```
ymarcus93, line 24: Please add a comment here explaining what this is used for

bctl-go-daemon/bctl/daemon/plugin/kube/kube.go
```diff
@@ -0,0 +1,277 @@
...
+func (k *KubeDaemonPlugin) InputMessageHandler(action string, actionPayload []byte) (string, []byte, error) {
+	if len(actionPayload) > 0 {
+		// Get just the request ID so we can associate it with the previously started action object
+		var d JustRequestId
+		if err := json.Unmarshal(actionPayload, &d); err != nil {
+			rerr := fmt.Errorf("could not unmarshal json: %s", err)
+			k.logger.Error(rerr)
+			return "", []byte{}, rerr
+		} else {
+			if act, ok := k.actions[d.RequestId]; ok {
```
ymarcus93, line 154: Not thread safe because you're not locking on the mutex

bctl-go-daemon/bctl/daemon/plugin/kube/kube.go
```diff
@@ -0,0 +1,277 @@
...
+
+		// TODO: Figure out what to do with this
+		log.Fatal(http.ListenAndServeTLS(":"+plugin.daemonPort, plugin.certPath, plugin.keyPath, nil))
+	}()
+
+	return &plugin, nil
+}
+
+func (k *KubeDaemonPlugin) handleStreamMessage(smessage smsg.StreamMessage) error {
+	if act, ok := k.actions[smessage.RequestId]; ok {
```
ymarcus93, line 126: Not thread safe because you're not locking on the mutex

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
```diff
@@ -0,0 +1,154 @@
...
+			// Verify the Signature
+			if err := ksMessage.VerifySignature(certMetadata.Cert.ClientPublicKey); err != nil {
+				return err
+			}
+		}
+
+		// Verify recieved hash pointer matches expected
+		if dataPayload.HPointer != k.expectedHPointer {
+			return fmt.Errorf("data's hash pointer did not match expected")
+		}
```
ymarcus93, line 105: @lgmugnier, @tflpd, @sidpremkumar 

Isn't there a race condition here? Please help me understand the following scenario and correct me if I'm wrong.

Assumptions:
- Each message that comes from the `InputChannel` (i.e. the websocket) are handled on different goroutines:

https://github.com/bastionzero/zli/blob/f0a6b68349a4e152e465d022a768f4c950c1cb7b/bctl-go-daemon/bctl/agent/datachannel/datachannel.go#L72-L78

https://github.com/bastionzero/zli/blob/f0a6b68349a4e152e465d022a768f4c950c1cb7b/bctl-go-daemon/bctl/agent/datachannel/datachannel.go#L110-L123

https://github.com/bastionzero/zli/blob/f0a6b68349a4e152e465d022a768f4c950c1cb7b/bctl-go-daemon/bctl/agent/datachannel/datachannel.go#L130-L133

- The call to `Validate()` checks that the received `HPointer` in the `DATA` matches the `expectedHPointer`:

https://github.com/bastionzero/zli/blob/62dd20f8292df943434c1a5765107d7982a4f2f5/bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go#L92-L95

- When we `BuildResponse` and create the `DATAACK`, the `expectedHPointer` is set to equal the hash of the `DATAACK` that is being sent:

https://github.com/bastionzero/zli/blob/62dd20f8292df943434c1a5765107d7982a4f2f5/bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go#L128-L136


Scenario:

- A `DATA` msg is sent with `HPointer` == daemon's last received H(`DATAACK`). Let's call this data msg `DATA_1`.
- Before receiving the `DATAACK` to the `DATA` above, another `DATA` msg is sent with `HPointer` == daemon's last received H(`DATAACK`). Let's call this data msg `DATA_2`.

Let `HPointer` == H(`DATAACK`) == `foo` for both `DATA_1` and `DATA_2`.

- Agent receives `DATA_1` first. It starts processing it on a separate goroutine as shown above:
   - `Validate()` is successful because the `HPointer` is correct from the agent's current perspective
   - `DATA_1`'s payload is passed to the correct plugin's `InputMessageHandler`.
   - Response comes back from plugin and datachannel now builds a `DATAACK` by calling `BuildResponse()`
   - **Point C:** `expectedHPointer` is updated and set to the hash of the `DATAACK` that is acking `DATA_1`.
- Agent receives `DATA_2` at **Point C**. It starts processing it on a separate goroutine:
  - Assume **Point C** has finished. `expectedHPointer` has been updated.
  - `DATA_2`'s `HPointer` will not match the `expectedHPointer` because it's still using `foo` instead of the new `expectedHPointer` which was not known by the daemon at the time `DATA_2` was sent.
  - ERROR. `DATA_2` cannot be processed...unexpected HPointer.

bctl-go-daemon/bzerolib/channels/message/signalr.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Hub name aka "Target" from "SignalRWrapper".  These are like API
+// endpoint that the message will hit
+type Hub string
+
+const (
+	RequestBastionToAgent  Hub = "RequestBastionToAgent"
+	ResponseAgentToBastion Hub = "ResponseAgentToBastion"
+)
```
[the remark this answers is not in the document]
lgmugnier: This has been removed c08c2b6

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+
+type IWebsocket interface {
+	Connect(serviceUrl string, hubEndpoint string, headers map[string]string, params map[string]string) error
+	Send(agentMessage wsmsg.AgentMessage) error
+	// We should probably also have a close method here too
+}
+
+// This will be the client that we use to store our websocket connection
+type Websocket struct {
+	Client  *websocket.Conn
```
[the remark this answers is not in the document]
lgmugnier: why? I think it's pretty common to refer to this as `client`

bctl-go-daemon/bctl/daemon/daemon.go
lgmugnier: do you mean certPath? Yes, it's passed in as a flag on daemon startup

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
lgmugnier: I believe you're right and we should be processing the messages in series.  I think @sidpremkumar put them into separate routines because of some race condition he encountered with exec?

bctl-go-daemon/bctl/daemon/daemon.go
lgmugnier: https://github.com/bastionzero/zli/blob/41b193bb70bfc9808dd65322bd02ddeee0d699a4/src/handlers/generate-kubeconfig.handler.ts#L31

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+	Cancel context.CancelFunc
+	Closed bool
+
+	// These are the channels for recieving and sending messages and done
+	InputChannel  chan wsmsg.AgentMessage
+	OutputChannel chan wsmsg.AgentMessage
+	DoneChannel   chan string
+
+	// Function for figuring out correct Target SignalR Hub
+	targetSelectHandler func(msg wsmsg.AgentMessage) (string, error)
```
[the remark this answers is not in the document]
lgmugnier: Not that you're wrong, but it might be more work/complexity than its worth.  Ideally, in the near future this concept should be somewhat-ish obsolete and we'll properly use your suggestion with Hubs at that point

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
lgmugnier: Let me make sure I'm understanding you correct @ymarcus93.  What you're saying is that when we are sending a bunch of DATA messages (let's forget for a second that reponses are sent in the DATAACKs and the daemon therefore has the ability to pre-calculate the responses to it's messages and send an infinite number of DATA messages).  And each of those messages are essentially processed in parallel, then the correct hpointer is going to get all messed up because the agent does need to process all incoming messages in order.

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
ymarcus93: Ok. I agree. We can push this to some later date. I'm resolving this related comment as well: https://github.com/bastionzero/zli/pull/124#discussion_r701329807

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
lgmugnier: These lines mean that for every time the `InputMessageHandler` is called there is a single response.  

https://github.com/bastionzero/zli/blob/41b193bb70bfc9808dd65322bd02ddeee0d699a4/bctl-go-daemon/bctl/daemon/plugin/kube/kube.go#L168-L180

The response comes from a channel which is buffered and can have up to 100 messages in it, but if there has been no DATAACK there will be no DATA sent

https://github.com/bastionzero/zli/blob/41b193bb70bfc9808dd65322bd02ddeee0d699a4/bctl-go-daemon/bctl/daemon/datachannel/datachannel.go#L208

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+// Constructor to create a new common websocket client object that can be shared by the daemon and server
+func NewWebsocket(serviceUrl string, hubEndpoint string, params map[string]string, headers map[string]string, targetSelectHandler func(msg wsmsg.AgentMessage) (string, error), autoReconnect bool, getChallenge bool) (*Websocket, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	ret := Websocket{
+		Cancel: cancel,
+		Closed: false,
+
+		InputChannel:        make(chan wsmsg.AgentMessage, 200),
+		OutputChannel:       make(chan wsmsg.AgentMessage, 200),
```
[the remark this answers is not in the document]
lgmugnier: This sounds like just another layer of abstraction?

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+	// 	return &ret, fmt.Errorf("Error connecting to websocket: %s", err.Error())
+	// }
+
+	ret.Connect(serviceUrl, hubEndpoint, headers, params)
+
+	go func() {
+		for {
+			select {
+			case msg := <-ret.OutputChannel:
+				ret.Send(msg)
```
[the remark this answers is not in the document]
lgmugnier: this has been solved in one of my current branches, we hadn't properly implemented all error handling at this point in the code

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+
+				if err != nil && ret.Closed == false {
+					// If we see an error, and we are not trying to close the connection
+					// TODO: Handle this error better
+					log.Println("Error in websocket, will attempt to reconnect: ", err)
+					ret.IsReady = false
+					if err.Error() == "websocket: close 1000 (normal)" {
+						log.Println("Normal closure in websocket. Ending websocket connection")
+						return
+					}
```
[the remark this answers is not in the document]
lgmugnier: I'll check it out and see if that works

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+					if ret.autoReconnect {
+						ret.Connect(serviceUrl, hubEndpoint, headers, params)
+					} else {
+						log.Println("Auto-recoonect disabled, returning")
+						return
+					}
+				} else if ret.Closed == true {
+					// If we are trying to close to connection, end the goroutine
+					return
```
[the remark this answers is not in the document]
lgmugnier: this has already been taken care of

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+
+	// Extract the challange
+	responseDecoded := GetChallengeResponse{}
+	json.NewDecoder(response.Body).Decode(&responseDecoded)
+
+	// Solve Challenge
+	return SignChallenge(privateKey, responseDecoded.Challenge)
+}
+
+func SignChallenge(privateKey string, challange string) (string, error) {
```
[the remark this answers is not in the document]
lgmugnier: already fixed

bctl-go-daemon/bzerolib/keysplitting/util/util.go
```diff
@@ -0,0 +1,45 @@
...
+	"bytes"
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+
+	"golang.org/x/crypto/sha3"
+)
+
+func HashPayload(payload interface{}) ([]byte, bool) {
```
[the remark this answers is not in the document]
lgmugnier: instead of an `err`, I wanted to implement an `ok` here because there is very little information we get from an error here and I realized in the code that I was mostly ignoring the error because it was unlikely to show up (since the payload we're hashing are also the ones we're literally building)

bctl-go-daemon/bzerolib/keysplitting/bzcert/bzcert.go
```diff
@@ -0,0 +1,58 @@
...
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"bastionzero.com/bctl/v1/bzerolib/keysplitting/util"
+)
+
+const (
+	bzecertLifetime = time.Hour * 24 * 365 * 5 // 5 years
+)
```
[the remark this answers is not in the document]
lgmugnier: https://github.com/bastionzero/zli/blob/41b193bb70bfc9808dd65322bd02ddeee0d699a4/bctl-go-daemon/bzerolib/keysplitting/bzcert/bzcertverifier.go#L126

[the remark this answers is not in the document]
lgmugnier: This is a holdover from before we had implemented keysplitting correctly

bctl-go-daemon/bzerolib/plugin/plugin.go
```diff
@@ -0,0 +1,23 @@
+package plugin
+
+import (
+	smsg "bastionzero.com/bctl/v1/bzerolib/stream/message"
+)
+
+type PluginName string
+
+const (
+	Kube       PluginName = "kube"
```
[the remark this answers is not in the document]
lgmugnier: technically it's kinda both.  They both need to be able to say that the plugin is "kube"

bctl-go-daemon/bctl/agent/datachannel/datachannel.go
```diff
@@ -0,0 +1,202 @@
...
+
+type IDataChannel interface {
+	Send(messageType wsmsg.MessageType, messagePayload interface{}) error
+	Receive(agentMessage wsmsg.AgentMessage) error
+}
+
+type DataChannel struct {
+	websocket *ws.Websocket
+	logger    *lggr.Logger
+	ctx       context.Context
```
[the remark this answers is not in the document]
lgmugnier: I'm not going to say that the way I implemented contexts is the best.  But it does work for the use case we have and for that same reason it doesn't fall victim to the reasons doing it the way we did it are bad.  I think tomb is probably the right way to go but I'm going to make a ticket for this and delay it for now as I don't see this as a current priority.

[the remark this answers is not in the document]
lgmugnier: https://commonwealthcrypto.atlassian.net/browse/CWC-1058

bctl-go-daemon/bzerolib/keysplitting/keysplitting.go
```diff
@@ -0,0 +1,284 @@
...
+	Exp  time.Time
+}
+
+type IKeysplitting interface {
+	BuildSyn(action string, payload []byte) (ksmsg.KeysplittingMessage, error)
+	Validate(ksMessage *ksmsg.KeysplittingMessage) error
+	BuildResponse(ksMessage *ksmsg.KeysplittingMessage, action string, actionPayload []byte) (ksmsg.KeysplittingMessage, error)
+}
+
+type Keysplitting struct {
```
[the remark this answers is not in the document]
lgmugnier: this has been removed in an incoming pr

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
```diff
@@ -0,0 +1,144 @@
...
+	Exp  time.Time
+}
+
+type IKeysplitting interface {
+	Validate(ksMessage *ksmsg.KeysplittingMessage) error
+	BuildResponse(ksMessage *ksmsg.KeysplittingMessage, action string, actionPayload []byte) (ksmsg.KeysplittingMessage, error)
+}
+
+type Keysplitting struct {
+	hPointer         string
```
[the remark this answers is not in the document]
lgmugnier: It's used in error messages and that an incoming PR

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
```diff
@@ -0,0 +1,144 @@
...
+	"fmt"
+	"time"
+
+	bzcrt "bastionzero.com/bctl/v1/bzerolib/keysplitting/bzcert"
+	ksmsg "bastionzero.com/bctl/v1/bzerolib/keysplitting/message"
+	"bastionzero.com/bctl/v1/bzerolib/keysplitting/util"
+)
+
+const (
+	schemaVersion = "1.0"
```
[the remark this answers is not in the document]
lgmugnier: no where, it's been removed

bctl-go-daemon/bctl/agent/plugin/kube/kube.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func (k *KubePlugin) GetName() plgn.PluginName {
+	return plgn.Kube
+}
+
+func (k *KubePlugin) PushStreamInput(smessage smsg.StreamMessage) error {
+	return fmt.Errorf("")
+}
```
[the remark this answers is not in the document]
lgmugnier: So that Kube plugin can fully implement `IPlugin`

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
ymarcus93: It is another layer of abstraction, but it prevents future developers from making a mistake, e.g. sending on a channel that shouldn't be sent on outside of the `websocket.go` struct, or receiving on a channel that shouldn't be received on outside of the `websocket.go` struct.

If someone tries to receive on a `chan<-`, the Go compiler will fail to compile the code. If someone tries to send on a `<-chan`, the Go compiler will fail to compile the code.

We don't have to do this now as it's not super important. But I like when we can use type safety/compiler to help prevent these issues from arising.

bctl-go-daemon/bzerolib/keysplitting/util/util.go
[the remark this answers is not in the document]
ymarcus93: Ok

bctl-go-daemon/bctl/agent/datachannel/datachannel.go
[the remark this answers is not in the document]
ymarcus93: Yup. Better to use context than no context, otherwise we run into leaky goroutines that can never be cancelled. I agree that we don't need to look into this right now. Thanks for making the ticket to track this.

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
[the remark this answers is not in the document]
ymarcus93: Ok

bctl-go-daemon/bctl/agent/keysplitting/keysplitting.go
ymarcus93: Ok this makes sense now. I can help write some comments as I was confused at first glance.

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+	SocketLock sync.Mutex
+
+	// These objects are used for closing the websocket
+	Cancel context.CancelFunc
+	Closed bool
+
+	// These are the channels for recieving and sending messages and done
+	InputChannel  chan wsmsg.AgentMessage
+	OutputChannel chan wsmsg.AgentMessage
+	DoneChannel   chan string
```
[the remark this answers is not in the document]
lgmugnier: eabdb87

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+type Websocket struct {
+	Client  *websocket.Conn
+	IsReady bool
+
+	// Ref: https://github.com/gorilla/websocket/issues/119#issuecomment-198710015
+	SocketLock sync.Mutex
+
+	// These objects are used for closing the websocket
+	Cancel context.CancelFunc
+	Closed bool
```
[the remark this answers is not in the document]
lgmugnier: 7bd3302

bctl-go-daemon/bctl/daemon/daemon.go
lgmugnier: this has been fixed in my `error-messages` pr.  We are no longer sleeping and instead waiting for the bastion to tell us the communication pathway is open

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
lgmugnier: okay lets revisit this during our code structure discussions

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+
+// Function to write signalr message to websocket
+func (w *Websocket) Send(agentMessage wsmsg.AgentMessage) error {
+	if !w.IsReady {
+		return fmt.Errorf("Websocket not ready to send yet")
+	}
+
+	// Lock our mutex and setup the unlock
+	w.SocketLock.Lock()
+	defer w.SocketLock.Unlock()
```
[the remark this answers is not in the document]
lgmugnier: yes, you are right.  this was removed in 58b376f

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
lgmugnier: d2159af

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+
+			// Add the connection id to the list of params
+			params["id"] = connectionId.(string)
+			params["clientProtocol"] = "1.5"
+			params["transport"] = "WebSockets"
+
+			// Make an interrupt channel
+			// TODO: Think this can be removed
+			interrupt := make(chan os.Signal, 1)
+			signal.Notify(interrupt, os.Interrupt)
```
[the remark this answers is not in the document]
lgmugnier: 6dd5d2b

bctl-go-daemon/bzerolib/keysplitting/bzcert/bzcert.go
```diff
@@ -0,0 +1,58 @@
...
+
+type BZCert struct {
+	InitialIdToken  string `json:"initialIdToken"`
+	CurrentIdToken  string `json:"currentIdToken"`
+	ClientPublicKey string `json:"clientPublicKey"`
+	Rand            string `json:"rand"`
+	SignatureOnRand string `json:"signatureOnRand"`
+}
+
+func (b *BZCert) Verify() (string, time.Time, error) {
```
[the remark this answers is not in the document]
lgmugnier: fa5e63c

bctl-go-daemon/bzerolib/keysplitting/bzcert/bzcert.go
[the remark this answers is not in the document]
lgmugnier: e8ff3ce

bctl-go-daemon/bzerolib/stream/stdwriter/stdwriter.go
```diff
@@ -0,0 +1,45 @@
+package stdwriter
+
+import (
+	"encoding/base64"
+
+	smsg "bastionzero.com/bctl/v1/bzerolib/stream/message"
+)
+
+var (
+	notreadybytes = [2]byte{64, 94}
```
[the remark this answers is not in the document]
lgmugnier: f471197

bctl-go-daemon/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -0,0 +1,306 @@
...
+					return
+				}
+			}
+		}
+	}()
+	return &control, nil
+}
+
+func (c *ControlChannel) Receive(agentMessage wsmsg.AgentMessage) error {
+	switch wsmsg.MessageType(agentMessage.MessageType) {
```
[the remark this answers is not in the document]
lgmugnier: 7caadac

bctl-go-daemon/bctl/agent/datachannel/datachannel.go
```diff
@@ -0,0 +1,202 @@
...
+				case <-d.ctx.Done():
+					return
+				case streamMessage := <-ch:
+					d.Send(wsmsg.Stream, streamMessage)
+				}
+			}
+		}()
+
+		subLogger := d.logger.GetPluginLogger(plugin)
+		d.plugin = kube.NewPlugin(d.ctx, subLogger, ch, d.role)
```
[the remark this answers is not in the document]
lgmugnier: this has already been fixed.  If a new syn is received and there's already a plugin started, no new plugin in started

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/exec.go
```diff
@@ -0,0 +1,178 @@
...
+const (
+	EscChar = "^[" // ESC char
+)
+
+type ExecAction struct {
+	serviceAccountToken string
+	kubeHost            string
+	impersonateGroup    string
+	role                string
+	logId               string
```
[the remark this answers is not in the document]
lgmugnier: I actually want to revisit this for our code structure tear-down, so I'm pushing this one til then

bctl-go-daemon/bctl/daemon/daemon.go
lgmugnier: this is another one we should revisit later

bctl-go-daemon/bctl/agent/controlchannel/controlchannelmessages.go
```diff
@@ -0,0 +1,30 @@
+package controlchannel
+
+type NewDatachannelMessage struct {
+	ConnectionId string `json:"connectionId"`
+	Role         string `json:"role"`
```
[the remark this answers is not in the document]
sidpremkumar: So after seeing this comment, I looked into refactoring everything to `User`. But I noticed that things quickly get confusing, especially in the bastion. Since in bastion we have a concept of a user/subject, you can see how changing this from Role -> User might get confusing really easily: 
```
KubeLogsRepository.Add(new Database.Schema.KubeLogs
                {
                    CreationDate = DateTimeOffset.Now,
                    Id = message.LogId,
                    OrgId = message.OrgId,
                    UserId = message.UserId,
                    User = message.User,
                    Endpoints = endpoints,
                    ExecCommands = execCommands,
                    KubeEnglishCommand = message.KubeEnglishCommand,
                    ClusterName = message.ClusterName,
                    ClusterId = message.ClusterId
                });
```
Is `User=message.User` talking about Kube user, or Bastion user?. It could just be me, but it seems that even tho "Role" is technically incorrect, it makes it easier to see the difference between Kube users/Bastion users. 

But I'm open to making this change if you think it's worth it?

src/handlers/bctl.handler.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        if (code != 0) {
+            // Check to ensure they are using the right context
+            const currentContext = await execPromise('kubectl config current-context ');
+
+            if (currentContext.stdout != 'bctl-server') {
+                logger.warn('Make sure you using the correct kube config!');
+            }
+        }
+    });
```
ymarcus93, line 51: I'm kind of confused how this works. Isn't there a chance that the process spawns and exits with non-zero exit code before `kubeCommandProcess.on` can register a handler on `close`?

Also, shouldn't we `await kubeCommandProcess`? How does `bctlHandler` know when the spawned process (`kubectl`)  has finished? Because right now without the `await`, doesn't the handler function just return and `zli` exits?

bctl-go-daemon/bctl/daemon/plugin/kube/kube.go
sidpremkumar: Please see https://github.com/bastionzero/zli/pull/183

src/handlers/delete-target-user.handler.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+        // Then remove the role from the policy if it exists
+        delete kubernetesPolicyContext.clusterUsers[targetUserName];
+
+        // And finally update the policy
+        policy.context = kubernetesPolicyContext;
+        break;
+    case PolicyType.TargetConnect:
+        const targetConnectContext = policy.context as TargetConnectContext;
+        const targetUsers = targetConnectContext.targetUsers as {[targetUser: string]: TargetUser};
```
ymarcus93, line 37: As I said above, I think there may be a future refactor where we can have typescript infer what type `policy.context` is so we don't have to do these type assertions if we follow [this](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions).

src/handlers/delete-user-policy.handler.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Logger } from '../logger.service/logger';
+import { ClusterSummary } from '../types';
+import { cleanExit } from './clean-exit.handler';
+
+export async function deleteUserFromPolicyHandler(userEmail: string, policyName: string, clusterTargets: Promise<ClusterSummary[]>, configService: ConfigService, logger: Logger) {
+    // First ensure we can lookup the user
+    const kubeService = new KubeService(configService, logger);
+    const userInfo = await kubeService.GetUserInfoFromEmail(userEmail);
+
+    if (userInfo.email == 'unknown') {
```
ymarcus93, line 13: Backend stores literal string `unknown` in this property? Can we add a comment here?

bctl-go-daemon/bctl/daemon/plugin/kube/kube.go
lgmugnier: 094651e

bctl-go-daemon/bctl/daemon/plugin/kube/kube.go
lgmugnier: 094651e

bctl-go-daemon/bctl/agent/plugin/kube/kube.go
```diff
@@ -0,0 +1,161 @@
...
+	} else {
+		rid = justrid.RequestId
+
+		subLogger := k.logger.GetActionLogger(action)
+		subLogger.AddRequestId(rid)
+	}
+
+	// Interactive commands like exec and log need to be able to recieve multiple inputs, so we start them and track them
+	// and send any new messages with the same request ID to the existing action object
+	if act, ok := k.actions[rid]; ok {
```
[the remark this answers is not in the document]
lgmugnier: 094651e

bctl-go-daemon/bctl/agent/plugin/kube/kube.go
```diff
@@ -0,0 +1,161 @@
...
+		action, payload, err := act.InputMessageHandler(action, actionPayloadSafe)
+
+		// Check if that last message closed the action, if so delete from map
+		if act.Closed() {
+			delete(k.actions, rid)
+		}
```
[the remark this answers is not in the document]
lgmugnier: 094651e

bctl-go-daemon/bctl/agent/plugin/kube/kube.go
```diff
@@ -0,0 +1,161 @@
...
+	RestApi KubeAction = "restapi"
+)
+
+type KubePlugin struct {
+	role                string
+	streamOutputChannel chan smsg.StreamMessage
+	serviceAccountToken string
+	kubeHost            string
+	actions             map[string]IKubeAction
+	actionsMapLock      sync.RWMutex
```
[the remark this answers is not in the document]
lgmugnier: 094651e

src/handlers/describe-cluster.handler.ts
```diff
@@ -0,0 +1,64 @@
+import { Logger } from '../logger.service/logger';
+import { ConfigService } from '../config.service/config.service';
+import { cleanExit } from './clean-exit.handler';
+import * as types from '../types';
+import { EnvironmentDetails } from '../http.service/http.service.types';
+import { PolicyQueryService } from '../http.service/http.service';
+
+
+export async function describeClusterHandler(
```
ymarcus93, line 9: Another example where a helper function would be nice.

For example, create a type that represents all the information here:

```
logger.info(`Cluster information for: ${clusterName}`);
    logger.info(`    - Environment Name: ${environment.name}`);
    logger.info(`    - Policies using this cluster: ${policiesString}`);
    logger.info(`    - Valid Cluster Users: ${validUserString}`);
```

Then, the helper function can return that type. This allows others to use the same functionality in some other command later. `describeClusterHandler` will use it and just `log.info` as you're doing right now.

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/exec.go
```diff
@@ -0,0 +1,178 @@
...
+	impersonateGroup string,
+	role string,
+	ch chan smsg.StreamMessage) (*ExecAction, error) {
+
+	return &ExecAction{
+		serviceAccountToken: serviceAccountToken,
+		kubeHost:            kubeHost,
+		impersonateGroup:    impersonateGroup,
+		role:                role,
+		closed:              false,
```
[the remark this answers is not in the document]
lgmugnier: 1bcb650

src/handlers/generate-kube-yaml.handler.ts
```diff
@@ -0,0 +1,64 @@
...
+    const labels: { [index: string ]: string } = {};
+    if (argv.labels != []) {
+        for (const keyValueString of argv.labels) {
+            const key = keyValueString.split(':')[0];
+            const value = String(keyValueString.split(':')[1]);
+            labels[key] = value;
+        }
+    }
+
+    // If environemtn has been passed, ensure its a valid envId
```
ymarcus93, line 39: typo:

```suggestion
    // If environment has been passed, ensure it's a valid envId
```

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/exec.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Add our impersonation information
+	config.Impersonate = rest.ImpersonationConfig{
+		UserName: e.role,
+		Groups:   []string{e.impersonateGroup},
+	}
+	config.BearerToken = e.serviceAccountToken
+
+	kubeExecApiUrl := e.kubeHost + startExecRequest.Endpoint
+	kubeExecApiUrlParsed, _ := url.Parse(kubeExecApiUrl)
```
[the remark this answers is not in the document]
lgmugnier: c5e637e

src/handlers/generate-kube-yaml.handler.ts
```diff
@@ -0,0 +1,64 @@
...
+        (await envs).forEach(env => {
+            if (env.id == argv.environmentId) {
+                validEnv = true;
+            }
+        });
+        if (validEnv == false) {
+            logger.error('The environment Id you passed is invalid.');
+            await cleanExit(1, logger);
+        }
+    }
```
ymarcus93, line 51: Also, I feel like I've seen this code before. `autodiscovery-script` does it. And I think my new `generate-bash` handler does it. Perhaps that justifies making a common function that can check if an envId is valid by querying the `EnvironmentService`?

bctl-go-daemon/bctl/agent/plugin/kube/actions/logs/logs.go
```diff
@@ -0,0 +1,215 @@
...
+
+type LogAction struct {
+	RequestId           string
+	serviceAccountToken string
+	kubeHost            string
+	impersonateGroup    string
+	role                string
+	streamOutputChannel chan smsg.StreamMessage
+	closed              bool
+	doneChannel         chan bool
```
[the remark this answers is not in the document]
lgmugnier: context can only be cancelled by datachannel.  This structure will need to be revisited tho

src/handlers/generate-kubeconfig.handler.ts
```diff
@@ -0,0 +1,116 @@
...
+) {
+    // Check if we already have generated a cert/key
+    let kubeConfig = configService.getKubeConfig();
+
+    if (kubeConfig == undefined) {
+        logger.info('No KubeConfig has been generated before, generating key and cert for local daemon...');
+
+        // Create and save key/cert
+        const createCertPromise = new Promise<void>(async (resolve, reject) => {
+            pem.createCertificate({ days: 999, selfSigned: true }, async function (err: any, keys: any) {
```
ymarcus93, line 23: @sidpremkumar So somewhere in `go` do we have some flag that says "accept self-signed cert"? I couldn't find that when reviewing the go stuff.

src/handlers/generate-kubeconfig.handler.ts
```diff
@@ -0,0 +1,116 @@
...
+                        logger.error('Error writing key to file!');
+                        reject();
+                        return;
+                    }
+                    logger.debug('Generated and saved key file');
+                });
+                await fs.writeFile(pathToCert, keys.certificate, function (err: any) {
+                    if (err) {
+                        logger.error('Error writing cert to file!');
+                        reject();
```
ymarcus93, line 45: Why not include the `err` in `reject()`?

src/handlers/generate-kubeconfig.handler.ts
```diff
@@ -0,0 +1,116 @@
...
+                // Get the path of where we want to save
+                const pathToConfig = path.dirname(configService.configPath());
+                const pathToKey = `${pathToConfig}/kubeKey.pem`;
+                const pathToCert = `${pathToConfig}/kubeCert.pem`;
+
+                // Now save the key and cert
+                await fs.writeFile(pathToKey, keys.serviceKey, function (err: any) {
+                    if (err) {
+                        logger.error('Error writing key to file!');
+                        reject();
```
ymarcus93, line 37: Why not include the `err` in `reject()`?

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
```diff
@@ -0,0 +1,372 @@
...
+
+								// Deserialize the argument
+								var closeMessage wsmsg.CloseMessage
+								if err := json.Unmarshal(wrappedMessage.Arguments[0].MessagePayload, &closeMessage); err != nil {
+									log.Printf("Error unmarshalling SignalR Close Message from Bastion: %v", string(wrappedMessage.Arguments[0].MessagePayload))
+									break
+								}
+
+								// Send an alert on our done channel for our datachannel
+								ret.DoneChannel <- closeMessage.Message
```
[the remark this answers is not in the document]
lgmugnier: I don't particularly think it likely that we would have multiple objects receiving on that channel ever, but comment noted

src/handlers/generate-kubeconfig.handler.ts
```diff
@@ -0,0 +1,116 @@
...
+                const token = randtoken.generate(128) + '++++';
+
+                // Find an open port, define it here as if the config has already been created, this codeblock will never be executed
+                const findPort = require('find-open-port');
+                const localPort = new Promise<number>(async (resolve, _) => {
+                    findPort().then((port: any) => {
+                        resolve(port);
+                    });
+                });
+
```
ymarcus93, line 62: Can't this be simplified to?:

```suggestion
                const localPort = await findPort();

```

Because it looks like `findPort` returns a promise.

src/handlers/list-targets.handler.ts
```diff
@@ -5,20 +5,43 @@ import {
...
+    configService: ConfigService,
     logger: Logger,
     argv: any,
     dynamicConfigs: Promise<TargetSummary[]>,
     ssmTargets: Promise<TargetSummary[]>,
+    clusters: Promise<ClusterSummary[]>,
     envsPassed: Promise<EnvironmentDetails[]>) {
-    // await and concatenate
+
+    const clusterTargets = (await clusters).map<TargetSummary>((cluster, _index, _array) => {
```
ymarcus93, line 25: We aren't using `_index` and `_array`:

```suggestion
    const clusterTargets = (await clusters).map<TargetSummary>((cluster) => {
```

src/handlers/start-kube-daemon.handler.ts
```diff
@@ -0,0 +1,210 @@
...
+                process.exit();
+            });
+        }
+    } catch (error) {
+        logger.error(`Something went wrong starting the Kube Daemon: ${error}`);
+        await cleanExit(1, logger);
+    }
+}
+
+async function getClusterInfoFromName(clusterTargets: ClusterSummary[], clusterName: string, logger: Logger): Promise<ClusterSummary> {
```
ymarcus93, line 134: @tflpd In your refactor PR, should we move these functions to separate file to keep the `handler` file cleaner? This one especially looks like it can be reused elsewhere, so maybe `export` this one as well?

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/terminalsizequeue.go
```diff
@@ -0,0 +1,28 @@
...
+
+func NewTerminalSizeQueue(requestId string, execResizeChannel chan KubeExecResizeActionPayload) *TerminalSizeQueue {
+	return &TerminalSizeQueue{
+		execResizeChannel: execResizeChannel,
+		RequestId:         requestId,
+	}
+}
+
+func (t *TerminalSizeQueue) Next() *remotecommand.TerminalSize {
+	tsMessage := <-t.execResizeChannel
```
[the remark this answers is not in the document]
lgmugnier: I'm not sure I buy that we're responsible for that, that they don't have a way of cleaning up the go routines that they're spawning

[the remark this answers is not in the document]
lgmugnier: but I guess I can implement some `more` logic just in case

src/handlers/start-kube-daemon.handler.ts
ymarcus93: `copyExecutableToTempDir` and `deleteIfExists` are unique to this file's handler so probably don't `export` those.

src/handlers/start-kube-daemon.handler.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    // Copy over our executable to a temp file
+    let finalDaemonPath = '';
+    if (process.env.ZLI_CUSTOM_BCTL_PATH) {
+        // If we set a custom path, we will try to start the daemon from the source code
+        cwd = process.env.ZLI_CUSTOM_BCTL_PATH;
+        finalDaemonPath = 'go';
+        args = ['run', 'daemon.go'].concat(args);
+    } else {
+        finalDaemonPath = await copyExecutableToTempDir(logger, configService.configPath());
```
ymarcus93, line 77: @sidpremkumar This is probably why my computer always pops up that firewall question. We're creating a new executable in the temp dir each time this handler is called.

ymarcus93: I believe a new executable doesn't inherit the same firewall exceptions

src/handlers/start-kube-daemon.handler.ts
```diff
@@ -0,0 +1,210 @@
...
+                    shell: true,
+                    detached: true,
+                    stdio: 'inherit'
+                });
+            });
+
+            daemonProcess.on('exit', function () {
+                // Whenever the daemon exits, exit
+                process.exit();
+            });
```
ymarcus93, line 126: Does this do anything? Because shouldn't the handler code just fallthrough after running `await spawn` and exit? If we want the `zli` process to not exit until the daemon exits, we should convert this `else` branch to a promise that doesn't `resolve` until `exit` handler is hit. And then `await` on that promise.

src/handlers/start-kube-daemon.handler.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    // Create our executable file
+    fs.writeFileSync(finalDaemonPath, "");
+
+    // Copy the file to the computers file system
+    await copy(daemonExecPath, finalDaemonPath);
+
+    // Grant execute permission
+    // TODO: See if this is the right level of permission
+    await chmod(finalDaemonPath, 0o765);
```
ymarcus93, line 197: https://chmodcommand.com/chmod-0765/:

> Chmod 0765 (chmod a+rwx,g-x,o-w,ug-s,-t) sets permissions so that, (U)ser / owner can read, can write and can execute. (G)roup can read, can write and can't execute. (O)thers can read, can't write and can execute.

Is this what we want? Shouldn't the group be able to execute as well?

src/http.service/http.service.ts
```diff
@@ -124,6 +169,74 @@ export class HttpService
...
+                const requestStream = this.httpClient.stream.post(
+                    route,
+                    {
+                        isStream: true,
+                        body: formBody
+                    }
+                );
+
+                // Buffer is returned by 'data' event
+                requestStream.on('data', (response: Buffer) => {
```
ymarcus93, line 225: So `data` event is triggered at end of stream, or on every received chunk in the download?

ymarcus93 (commented):
`zli` comments:

bctl-go-daemon/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
ymarcus93: Ok. Resolving this then.

bctl-go-daemon/bzerolib/stream/stdreader/stdreader.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"bytes"
+	"io"
+
+	smsg "bastionzero.com/bctl/v1/bzerolib/stream/message"
+)
+
+var (
+	EndStreamBytes = []byte{0x62, 0x61, 0x73, 0x74, 0x69, 0x6f, 0x6e, 0x7a, 0x65, 0x72, 0x6f}
```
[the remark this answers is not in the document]
ymarcus93: @lgmugnier Can you link to commit where you added comment?

bctl-go-daemon/bctl/agent/controlchannel/controlchannelmessages.go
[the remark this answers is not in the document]
ymarcus93: > ... groups that the same column names would work etc.

If we want to support impersonating a group, won't we need a new column like `ImpersonatedKubeGroup`? Because it looks like there is a separate header for it: `Impersonate-Group`. There may be a scenario where we need to impersonate as a user and group whose names aren't equal. Also, if I recall right now we assume `impersonateGroup == system:authenticated`--there is no OPA policy option for it yet so it's probably not important right now to have this distinction.

> Is User=message.User talking about Kube user, or Bastion user?. It could just be me, but it seems that even tho "Role" is technically incorrect, it makes it easier to see the difference between Kube users/Bastion users. 

We can change that column to `ImpersonatedKubeUser`, and then I think the ambiguity between Bastion users and Kube users is removed.

> , that when we eventually support impersonating Service account / ...

But even in the case of service accounts, Kube uses the terminology `User`. See: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-subjects

> 
    Note:
        - system:serviceaccount: (singular) is the prefix for service account usernames.
        - system:serviceaccounts: (plural) is the prefix for service account groups.

It seems that service accounts are automatically mapped to the `system:serviceaccount:<service-account-name>` user and `system:serviceaccounts:<service-account-name>` group.

> But I'm open to making this change if you think it's worth it?

The change is probably not worth it at the current moment. Let's just keep everything as is.

bctl-go-daemon/bctl/agent/datachannel/datachannel.go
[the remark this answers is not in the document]
ymarcus93: Can you link to the commit or the file?

bctl-go-daemon/bctl/agent/plugin/kube/kube.go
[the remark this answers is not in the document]
ymarcus93: @lgmugnier We should revisit this at some point. I think there is a way where we don't have to have this boilerplate function. We just have to scope down our interfaces. And use different ones for agent plugin vs. daemon plugin--if one doesn't need this function but the other does.

.gitignore
```diff
@@ -338,3 +338,9 @@ ASALocalRun/
 
 # Local History for Visual Studio
 .localhistory/
+
+# Our built agent 
+bctl-go-daemon/bctl/daemon/daemon-linux
+bctl-go-daemon/bctl/daemon/daemon-macos
+bctl-go-daemon/bctl/daemon/daemon-windows
+package-lock.json
```
lipmas, line 346: We should not be adding package-lock.json to .gitignore. It is intended for this file to be checked in to source control to keep versions consistent across dev/build machines.

Also side note if its annoying that the package-lock.json keeps updating when doing `npm install`  we should using `npm ci` instead which wont mutate the lock file. We should also have our CI/CD commands use this version. See https://stackoverflow.com/a/48524475/9186330

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/terminalsizequeue.go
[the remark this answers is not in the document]
ymarcus93: I think we are responsible for that. The code I just referenced handles `nil` explicitly by exiting from the goroutine to clean it up. And the public comment here for the interface we're implicitly implementing on our `TerminalSizeQueue` struct: https://pkg.go.dev/k8s.io/client-go/tools/remotecommand?utm_source=gopls#TerminalSizeQueue references how `Next()` should be implemented:

> It returns nil when monitoring has been stopped.

Feel free to push this to another date. I don't think this is blocking for release. Also, I can't find the original code for this file, so maybe this comment isn't even applicable anymore?

CC: @sidpremkumar

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/exec.go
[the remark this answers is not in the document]
ymarcus93: Ok

src/cli-driver.ts
```diff
@@ -43,16 +59,94 @@ export class CliDriver
...
     private ssmTargets: Promise<TargetSummary[]>;
     private dynamicConfigs: Promise<TargetSummary[]>;
+    private clusterTargets: Promise<ClusterSummary[]>;
     private envs: Promise<EnvironmentDetails[]>;
 
     // use the following to shortcut middleware according to command
-    private noOauthCommands: string[] = ['config', 'login', 'logout'];
-    private noMixpanelCommands: string[] = ['config', 'login', 'logout'];
-    private noFetchCommands: string[] = ['ssh-proxy-config', 'config', 'login', 'logout'];
+    private oauthCommands: string[] = [
```
lipmas, line 66: I think `kube` should now be added to both this and mixpanelCommands

src/cli-driver.ts
```diff
@@ -146,14 +250,96 @@ export class CliDriver
...
+                            default: -1,
+                            demandOption: false
+                        })
+                        .example('$0 tunnel admin@neat-cluster', 'Connect to neat-cluster as the admin Kube RBAC role');
+                },
+                async (argv) => {
+                    if (argv.tunnelString) {
+                        // TODO make this smart parsing
+                        const connectUser = argv.tunnelString.split('@')[0];
+                        const connectCluster = argv.tunnelString.split('@')[1];
```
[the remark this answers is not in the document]
lipmas: you can use [assignment destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to do this:

```
const [connectUser, connectCluster] = argv.tunnelString.split('@');
```

src/cli-driver.ts
```diff
@@ -146,14 +250,96 @@ export class CliDriver
...
+                            type: 'string',
+                        })
+                        .example('$0 status test-cluster', '');
+                },
+                async (argv) => {
+                    await describeClusterHandler(argv.clusterName, this.configService, this.logger, this.clusterTargets, this.envs);
+                }
+            )
+            .command(
+                'disconnect',
```
lipmas, line 324: Might be too late to change this now but this is a bit confusing with the existence of `list-connections`/`close`/`attach-to-connections` commands for connections. Might be worth renaming `close` => `close-connection` and then this to `close-tunnel`.

sidpremkumar: We can bring it up with Mike at the POC, i think this would be an easy thing to change last minute. Does that work?

.gitignore
sidpremkumar: https://github.com/bastionzero/zli/pull/124/commits/c0b64ef2a4a57a9e5c0d25003e94f5a8477f2701

src/cli-driver.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/124/commits/c0b64ef2a4a57a9e5c0d25003e94f5a8477f2701

bctl-go-daemon/bctl/agent/plugin/kube/actions/exec/terminalsizequeue.go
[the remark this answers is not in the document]
sidpremkumar: yes I believe Yuval might be right. We should call a `Close()` or something on the terminal size queue, so that way it will return nil. Not 100% sure if its not automatically cleaned up by 
```go
err := exec.Stream(remotecommand.StreamOptions{
			Stdin:             stdinReader,
			Stdout:            stdoutWriter,
			Stderr:            stderrWriter,
			TerminalSizeQueue: terminalSizeQueue,
			Tty:               true, // TODO: We dont always want tty
		})
```

Its very possible that once the `Stream` ends the terminalresizequeue will get cleaned up. Not sure if its worth adding right now, but yes something we should look into.

src/handlers/add-group-policy.handler.ts
```diff
@@ -0,0 +1,59 @@
...
+        await cleanExit(1, logger);
+    }
+
+    if (policy.type !== PolicyType.KubernetesTunnel && policy.type !== PolicyType.TargetConnect){
+        logger.error(`Adding group to policy ${policyName} failed. Adding groups to ${policy.type} policies is not currently supported.`);
+        await cleanExit(1, logger);
+    }
+
+    // If this group exists already
+    const group = groups.find(g => g.name == groupSummary.name);
```
lipmas, line 40: @tflpd Shouldnt this be checking `policy.groups` to see if the group exists in the policy? Wouldnt this always be true otherwise?

src/handlers/bctl.handler.ts
sidpremkumar: So `spawn` is async : https://stackoverflow.com/questions/58570325/how-to-turn-child-process-spawns-promise-syntax-to-async-await-syntax. See image too. 

I'm not sure if that clears things up. We just wait for it to exit, if its a non-zero exit code we just ensure they are using the right kube config file. And bubble up that alert back to the user. 

2021-09-07 at 4 46 PM

src/handlers/describe-cluster.handler.ts
sidpremkumar: What is the advantage of this though? sure its nice and we can reuse it, but right now we are only using it one place. Nice future enchantment if we decide to use this log again, but right now I dont think its worth a refactor just to use it in one place.

src/cli-driver.ts
lipmas: Yep for sure lets bring it up with mike no need to change for this PR

src/handlers/generate-kube-yaml.handler.ts
sidpremkumar: In the autodiscovery-script we search based on name, here we are searching based on Id

src/handlers/generate-kubeconfig.handler.ts
sidpremkumar: I do not think so, we have this in the kube config
```
insecure-skip-tls-verify: true
```

src/handlers/generate-kubeconfig.handler.ts
sidpremkumar: bc we do nothing with the return

src/handlers/generate-kubeconfig.handler.ts
sidpremkumar: bc we do nothing with the return

src/handlers/start-kube-daemon.handler.ts
sidpremkumar: Please take a look at `copyExecutableToTempDir`. We do not copy it to a temp dir anymore, I've update the function name

src/handlers/start-kube-daemon.handler.ts
sidpremkumar: It does do something. It waits until the daemon process has finished before leaving the zli process. this allows us to stream stdout from the daemon in debug mode.

src/handlers/start-kube-daemon.handler.ts
sidpremkumar: Updated to 755

sidpremkumar: https://chmodcommand.com/chmod-775/

src/cli-driver.ts
```diff
@@ -279,12 +465,204 @@ export class CliDriver
...
+                        .example('$0 user --json', 'List all users, output as json, pipeable')
+                        .example('$0 user --add test-cluster test@test.com ', 'Adds the test@test.com IDP user to test-cluster policy')
+                        .example('$0 user -d test-cluster test@test.com ', 'Removes the test@test.com IDP user from test-cluster policy');
+                },
+                async (argv) => {
+                    if (!! argv.add) {
+                        await addUserToPolicyHandler(argv.idpEmail, argv.policyName, this.clusterTargets, this.configService, this.logger);
+                    } else if (!! argv.delete) {
+                        await deleteUserFromPolicyHandler(argv.idpEmail, argv.policyName, this.clusterTargets, this.configService, this.logger);
+                    } else if (!(!!argv.add && !!argv.delete)) {
```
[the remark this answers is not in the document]
tflpd: We don't need it, no. It is a good practice, adds to readability; however it is also debatable. I have seen people strongly supporting it and people saying that it just add to confusion. I am open to change it, @lipmas wdyt?

src/handlers/add-user-policy.handler.ts
```diff
@@ -0,0 +1,62 @@
...
+    const policyService = new PolicyService(configService, logger);
+    const policies = await policyService.ListAllPolicies();
+
+    // Loop till we find the one we are looking for
+    for (const policy of policies) {
+
+        if (policy.name == policyName) {
+            if (policy.type !== PolicyType.KubernetesTunnel && policy.type !== PolicyType.TargetConnect){
+                logger.error(`Adding user to policy ${policyName} failed. Adding users to ${policy.type} policies is not currently supported.`);
+                await cleanExit(1, logger);
```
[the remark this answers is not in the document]
tflpd: While this is a good practice, since at the moment we have exactly zero pieces of code we want to reuse I don't believe it makes sense to make this refactor right now. We can make it easily as we go wherever it is needed - or at once when it is proven that we need it in many places

src/handlers/describe-cluster.handler.ts
tflpd: I agree. I see the point but I feel it should be an as you go refactor - unless we expect the majority of code to need this in the near future

src/handlers/start-kube-daemon.handler.ts
tflpd: Yes, I will adjust it in the refactor

src/handlers/add-group-policy.handler.ts
tflpd: Nice catch, thanks a lot!
01872f83a3b3639defe4d759936d38f44edd2f5f

lipmas (approved):

PR #156, opened 2021-08-26
Create new command generate-bash

ymarcus93:
## Description of the change

Creates a new command `generate-bash`. This command replaces the `autodiscovery-script` command; in a future release, `autodiscovery-script` will be removed (i.e. `autodiscovery-script` is deprecated).

This command simplifies the user flow for SSH onboarding as one can call this command with no flags at all. The new script, found in backend PR bastionzero/webshell-backend#811, has been made universal--the script will detect the target machine's os.

The replacement command is also more on par with the webapp compared to the previous command; a `zli` user can specify a target name schema if they don't want to name the target explicitly/manually.

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1028

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

- [x] Do all tests related to the changed code pass in development?

### DIFF
--- a/src/handlers/generate-bash/generate-bash.command-builder.test.ts
+++ b/src/handlers/generate-bash/generate-bash.command-builder.test.ts
@@ -1,0 +1,40 @@
+import yargs from "yargs";
+import { generateBashCmdBuilder } from "./generate-bash.command-builder";
+
+// Tests that code in .check() in generateBashCmdBuilder() does not mess up
+// mutual exclusion check on --targetName and --targetNameScheme flags
+test.each([
+    [["--targetName", "foo", "--targetNameScheme", "time"]],
+    [["--targetName", "foo", "       --targetNameScheme", "time"]],
+    [["--targetName", "foo", "--targetNameScheme=time"]],
+    [["--targetName", "foo", "       --targetNameScheme=time"]],
+])('check mutually exclusive error is thrown with processArgs: %s', (processArgs) => {
+    // Simulate passing in arguments for yargs and store validation error
+    let validationErr: Error;
+    yargs.command('generate-bash', '', (yargs) => generateBashCmdBuilder(processArgs, yargs), () => { })
+        .parse(['generate-bash', ...processArgs], {}, (err) => {
+            validationErr = err;
+        });
+
+    // Assert that mutual exclusion error is thrown
+    expect(validationErr).not.toBeNull();
+    expect(validationErr.message).toEqual('Arguments targetNameScheme and targetName are mutually exclusive');
+});
+
+// Tests that code in .check() in generateBashCmdBuilder() correctly permits the
+// --targetName flag to be passed by itself without any validation errors (e.g.
+// mutual exclusion error)
+test.each([
+    [["--targetName", "foo"]],
+    [["-n", "foo"]],
+])('check no validation error is thrown with processArgs: %s', (processArgs) => {
+    // Simulate passing in arguments for yargs and store validation error
+    let validationErr: Error;
+    yargs.command('generate-bash', '', (yargs) => generateBashCmdBuilder(processArgs, yargs), () => { })
+        .parse(['generate-bash', ...processArgs], {}, (err) => {
+            validationErr = err;
+        });
+
+    // Assert that no validation error is thrown
+    expect(validationErr).toBeNull();
+});
--- a/src/handlers/generate-bash/generate-bash.command-builder.test.ts
+++ b/src/handlers/generate-bash/generate-bash.command-builder.test.ts
@@ -1,13 +1,13 @@
-import yargs from "yargs";
-import { generateBashCmdBuilder } from "./generate-bash.command-builder";
+import yargs from 'yargs';
+import { generateBashCmdBuilder } from './generate-bash.command-builder';
 
 // Tests that code in .check() in generateBashCmdBuilder() does not mess up
 // mutual exclusion check on --targetName and --targetNameScheme flags
 test.each([
-    [["--targetName", "foo", "--targetNameScheme", "time"]],
-    [["--targetName", "foo", "       --targetNameScheme", "time"]],
-    [["--targetName", "foo", "--targetNameScheme=time"]],
-    [["--targetName", "foo", "       --targetNameScheme=time"]],
+    [['--targetName', 'foo', '--targetNameScheme', 'time']],
+    [['--targetName', 'foo', '       --targetNameScheme', 'time']],
+    [['--targetName', 'foo', '--targetNameScheme=time']],
+    [['--targetName', 'foo', '       --targetNameScheme=time']],
 ])('check mutually exclusive error is thrown with processArgs: %s', (processArgs) => {
     // Simulate passing in arguments for yargs and store validation error
     let validationErr: Error;
@@ -25,8 +25,8 @@ test.each([
 // --targetName flag to be passed by itself without any validation errors (e.g.
 // mutual exclusion error)
 test.each([
-    [["--targetName", "foo"]],
-    [["-n", "foo"]],
+    [['--targetName', 'foo']],
+    [['-n', 'foo']],
 ])('check no validation error is thrown with processArgs: %s', (processArgs) => {
     // Simulate passing in arguments for yargs and store validation error
     let validationErr: Error;
--- a/src/handlers/generate-bash/generate-bash.command-builder.ts
+++ b/src/handlers/generate-bash/generate-bash.command-builder.ts
@@ -10,7 +10,7 @@ export type generateBashArgs = { environment: string } &
 { targetName: string } &
 { outputFile: string }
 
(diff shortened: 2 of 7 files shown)


src/handlers/generate-bash-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+TARGET_NAME=$(curl -s -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id)`;
+                break;
+            case 'time':
+                targetNameScript = "TARGET_NAME=target-$(date +\"%m%d-%H%M%S\")";
+                break;
+            case 'hostname':
+                targetNameScript = "TARGET_NAME=$(hostname)";
+                break;
+            default:
+                throw new Error(`Unhandled target name scheme option: ${targetNameSchemeArg}`);
```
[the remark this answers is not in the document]
ymarcus93: This is no longer applicable as I'm now using union types with a compile-time exhaustive check:

https://github.com/bastionzero/zli/blob/380ad53123b2f3fc535fb9b85cb474465686d7cc/src/handlers/generate-bash/generate-bash.handler.ts#L19-L39

src/handlers/generate-bash/generate-bash.command-builder.test.ts
```diff
@@ -0,0 +1,40 @@
+import yargs from "yargs";
+import { generateBashCmdBuilder } from "./generate-bash.command-builder";
+
+// Tests that code in .check() in generateBashCmdBuilder() does not mess up
+// mutual exclusion check on --targetName and --targetNameScheme flags
+test.each([
+    [["--targetName", "foo", "--targetNameScheme", "time"]],
+    [["--targetName", "foo", "       --targetNameScheme", "time"]],
+    [["--targetName", "foo", "--targetNameScheme=time"]],
+    [["--targetName", "foo", "       --targetNameScheme=time"]],
```
lipmas, line 10: Cool unit test...but what is the difference between these 4 test inputs they look the same to me...

src/handlers/generate-bash/generate-bash.command-builder.ts
```diff
@@ -0,0 +1,98 @@
+import yargs from 'yargs';
+
+const targetNameSchemes = ["do", "aws", "time", "hostname"] as const;
+export type TargetNameScheme = typeof targetNameSchemes[number];
```
lipmas, line 4: huh this is interesting syntax havent seen before. So what is the type of `TargetNameScheme` in this case- is it a string?

src/handlers/generate-bash/generate-bash.handler.ts
```diff
@@ -0,0 +1,61 @@
...
+            case 'time':
+                targetNameScript = "TARGET_NAME=target-$(date +\"%m%d-%H%M%S\")";
+                break;
+            case 'hostname':
+                targetNameScript = "TARGET_NAME=$(hostname)";
+                break;
+            default:
+                // Compile-time exhaustive check
+                // See: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking
+                const _exhaustiveCheck: never = argv.targetNameScheme;
```
lipmas, line 37: So this fails to compile if you added another value to the `targetNameSchemes` const?

src/handlers/generate-bash/generate-bash.handler.ts
```diff
@@ -0,0 +1,61 @@
...
+                targetNameScript = String.raw`
+TOKEN=$(curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
+TARGET_NAME=$(curl -s -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id)`;
+                break;
+            case 'time':
+                targetNameScript = "TARGET_NAME=target-$(date +\"%m%d-%H%M%S\")";
+                break;
+            case 'hostname':
+                targetNameScript = "TARGET_NAME=$(hostname)";
+                break;
```
lipmas, line 33: Might be worth moving this to the common websocket library for now so we could refactor the webapp at a later point to reuse the same function.

lipmas (approved):

src/handlers/generate-bash/generate-bash.handler.ts
tflpd: Yes! It is very cool! You create your own type and it understands it in compilation time!

src/handlers/generate-bash/generate-bash.command-builder.test.ts
ymarcus93: The difference is really on how the third argument `--targetNameScheme` is inputted to yargs:

- Standard: `--targetNameScheme time`
- With extra spaces at beginning: `<leading whitespace>--targetNameScheme time`
- Using `=` aka one string flag: `--targetNameScheme=time`
- With extra spaces at beginning (with one string flag): `<leading whitespace>--targetNameScheme=time`

The portion of code this unit test is really checking is this code here in `check()` for the command builder:
https://github.com/bastionzero/zli/blob/3777e4a94e46a84e8c69cf00e2add0aef32c8d06/src/handlers/generate-bash/generate-bash.command-builder.ts#L74-L75

Basically, I was running into the mutual exclusion issue linked in the comment:

https://github.com/bastionzero/zli/blob/3777e4a94e46a84e8c69cf00e2add0aef32c8d06/src/handlers/generate-bash/generate-bash.command-builder.ts#L85-L91

And that `if` statement above used to just check for exact match using `==` rather than regex. The regex makes it so it matches cases like the one string flag (cases 3 and 4). If it did not match on cases 3 and 4, then the code for `check()` makes it so `--targetNameScheme=<schemeName>` format cannot be passed in **by itself**; because the first conditional would be true (`targetNameScheme` could not be found as it was an exact case match) and the option for `targetNameScheme` would incorrectly be set to `undefined`. Therefore, those cases are critical to check in case someone changes the code not to do the regex check.

src/handlers/generate-bash/generate-bash.handler.ts
ymarcus93: Yes :) Only possible because `TargetNameScheme` is a string literal (union type).

Ref: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking

src/handlers/generate-bash/generate-bash.command-builder.ts
ymarcus93: https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

It's a string literal that can only take on the value "do", "aws", "time", "hostname". Essentially, it's an enum that is more powerful because (1) I can do exhaustiveness checking at compile-time and (2) do things like this: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions

The union/sum type Typescript has (taken from FP) is pretty cool because we can create complex types/interfaces that have different properties, collapse them all into a single type using a union `type1 | type2 | type3`, and then do special logic on the different types after discriminating the correct one via a string literal property that `type1` and `type2` and `type3` all have defined differently. No typecasts are needed for this too.

I use it again in Quickstart.

src/handlers/generate-bash/generate-bash.handler.ts
ymarcus93: Can we delay this to the quickstart PR? I change this function once more in that PR

src/handlers/generate-bash/generate-bash.command-builder.ts
```diff
@@ -0,0 +1,98 @@
...
+                default: 'latest',
+                describe: 'Use a specific version of the agent',
+            }
+        )
+        .option(
+            'os',
+            {
+                type: 'string',
+                demandOption: false,
+                choices: ['centos', 'ubuntu', 'universal'],
```
tflpd, line 49: Should we abstract them as a const on top of the file?

tflpd (approved):

src/handlers/generate-bash/generate-bash.command-builder.ts
ymarcus93: I make that enhancement in #166

src/handlers/generate-bash/generate-bash.handler.ts
lipmas: Yup we can delay this until quickstart PR its just a nice to have refactor whenever possible.

src/handlers/generate-bash/generate-bash.command-builder.test.ts
lipmas: Ok thanks for the explanation...it makes sense to test the different whitespace variants because you specifically rely on them in the regex i was missing that.

src/handlers/generate-bash/generate-bash.command-builder.ts
lipmas: That is really cool! Didnt realize the potential of typescript string literals

PR #159, opened 2021-08-31
Small fixes aug-31

sidpremkumar:
## Description of the change

* Extract get-challenge logic to inside `websocket.go`
* Update the agent version schema 
* Fix logs so we actually can stream logs for a specific container

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1041

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


bctl-go-daemon/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -72,7 +65,7 @@ func NewControlChannel(serviceUrl string,
 
 	log.Printf("\nserviceURL: %v, \nhubEndpoint: %v, \nparams: %v, \nheaders: %v", serviceUrl, hubEndpoint, params, headers)
 
-	wsClient, err := ws.NewWebsocket(serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect)
+	wsClient, err := ws.NewWebsocket(serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect, config.Data.PrivateKey)
```
lgmugnier, line 68: I'm thinking that maybe we should have it so the websocket grabs the privatekey from config whenever it needs it, as opposed to passing it in here and possibly using a bool flag?

sidpremkumar: Yeah we can definitely do that. Do you see any advantages to doing it like that?

sidpremkumar: Actually thinking about it ur right that makes more sense (if the key changes in the background, if we have to pass more info in the future ). So will make that change.

lgmugnier: I think it's just that it's cleaner and it helps clarify what the variable is for for ease-of-reading and stuff.  I don't think we should be passing around the private key and it should always come from the config aka the trusted source.

tflpd: Could caching be a possible answer to why we would want to avoid rereading it every time?

bctl-go-daemon/bctl/agent/datachannel/datachannel.go
```diff
@@ -37,7 +37,7 @@ func NewDataChannel(role string,
 	targetSelectHandler func(msg wsmsg.AgentMessage) (string, error),
 	autoReconnect bool) (*DataChannel, error) {
 
-	wsClient, err := ws.NewWebsocket(serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect)
+	wsClient, err := ws.NewWebsocket(serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect, "")
```
tflpd, line 40: Maybe `nil` instead?

bctl-go-daemon/bctl/daemon/datachannel/datachannel.go
```diff
@@ -41,7 +41,7 @@ func NewDataChannel(configPath string,
 	targetSelectHandler func(msg wsmsg.AgentMessage) (string, error),
 	autoReconnect bool) (*DataChannel, error) {
 
-	wsClient, err := ws.NewWebsocket(serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect)
+	wsClient, err := ws.NewWebsocket(serviceUrl, hubEndpoint, params, headers, targetSelectHandler, autoReconnect, "")
```
tflpd, line 44: same^

tflpd (approved):

bctl-go-daemon/bctl/agent/controlchannel/controlchannel.go
lgmugnier: Reading from config should be happening very rarely:
1. Once on Startup
2. Once on starting a new Datachannel
3. Every time we renew challenge (which should be rarely but there's also nothing else going on if that's happening so there's no worries with it interfering with user processes)

lgmugnier (approved):
Love the spelling fixes!

PR #160, opened 2021-08-31
search for open port instead of using 1234 lol

sidpremkumar:
## Description of the change

> Description here (why is it needed, what does it do)

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1045

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/src/handlers/generate-kubeconfig.handler.ts
+++ b/src/handlers/generate-kubeconfig.handler.ts
@@ -5,6 +5,7 @@ import { ConfigService } from '../config.service/config.service';
 
 const path = require('path');
 const fs = require('fs');
+const pem = require('pem');
 
 export async function generateKubeconfigHandler(
     argv: any,
@@ -20,7 +21,6 @@ export async function generateKubeconfigHandler(
         // Create and save key/cert
         const createCertPromise = new Promise<void>(async (resolve, reject) => {
             // Define pem here as if the config has already been created, this codeblock will never be executed
-            const pem = require('pem');
             pem.createCertificate({ days: 999, selfSigned: true }, async function (err: any, keys: any) {
                 if (err) {
                     throw err;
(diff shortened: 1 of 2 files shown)


src/handlers/generate-kubeconfig.handler.ts
```diff
@@ -21,6 +19,8 @@ export async function generateKubeconfigHandler(
 
         // Create and save key/cert
         const createCertPromise = new Promise<void>(async (resolve, reject) => {
+            // Define pem here as if the config has already been created, this codeblock will never be executed
+            const pem = require('pem');
```
tflpd, line 23: I believe it is not a good TS coding principle to import things here, probably everything should go on top (?)

tflpd (approved):

src/handlers/generate-kubeconfig.handler.ts
sidpremkumar: so it used to be there, but my thinking was that since this codeblock will realistically only be executed ~one time on a customer's machine. we shouldn't import it every time we generate a kubeconfig. But maybe that's bad practice, @lipmas advice?

ymarcus93: I think some packages are targeted for ES6 and some are targeted for CommonJS, so they have different ways of importing them. I tried importing a package once using `import` but then it complained and I had to use `require`. CC: @chad-cwc 

https://stackoverflow.com/questions/31354559/using-node-js-require-vs-es6-import-export

lipmas: Yeah i agree with @tflpd here its better practice to include all imports at the top for readability also see [this](https://stackoverflow.com/a/43311154/9186330).

Also the performance impact of loading the module would only matter if the user is using the zli installed through npm. If they brew install/download the executable everything is already precompiled into an executable by pkg.

sidpremkumar: https://github.com/bastionzero/zli/pull/160/commits/0b05bf4708f20ee24f0b173a553756e136269f53

PR #161, opened 2021-08-31
Wrap kube commands under kube

sidpremkumar:
## Description of the change

Couple of changes here:
* Talked with mike and the kube commands should go under `zli kube get pods`
* Thanos requested that we have a way to build the zli without building the agent , added that under `npm run release-dev`
* Realised we were not using `$0` in our `.example()` (ref: https://yargs.js.org/docs/#api-reference-examplecmd-desc), so I fixed that. See image as that adds the `zli` to the example:
2021-08-31 at 4 12 PM

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1008

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/src/handlers/bctl.handler.ts
+++ b/src/handlers/bctl.handler.ts
@@ -1,11 +1,11 @@
 import { ConfigService } from '../../src/config.service/config.service';
 import { Logger } from '../../src/logger.service/logger';
 import { cleanExit } from './clean-exit.handler';
+import util from 'util';
+import { spawn, exec } from 'child_process';
 
 const { v4: uuidv4 } = require('uuid');
-const spawn = require('child_process').spawn;
-const { promisify } = require('util');
-const exec = promisify(require('child_process').exec)
+const execPromise = util.promisify(exec)
 
 
 export async function bctlHandler(configService: ConfigService, logger: Logger) {
@@ -43,9 +43,9 @@ export async function bctlHandler(configService: ConfigService, logger: Logger) 
 
         if (code != 0) {
             // Check to ensure they are using the right context
-            const currentContext = await exec('kubectl config current-context ');
+            const currentContext = await execPromise('kubectl config current-context ');
 
-            if (currentContext != 'bctl-server') {
+            if (currentContext.stdout != 'bctl-server') {
                 logger.warn('Make sure you using the correct kube config!');
             }
         }
--- a/src/handlers/bctl.handler.ts
+++ b/src/handlers/bctl.handler.ts
@@ -24,7 +24,7 @@ export async function bctlHandler(configService: ConfigService, logger: Logger) 
     const logId = uuidv4();
 
     // Now build our token
-    const kubeArgsRaw = process.argv.splice(2);
+    const kubeArgsRaw = process.argv.splice(3);
     const kubeArgsString = kubeArgsRaw.join(' ');
     const formattedToken = `${token}zli ${kubeArgsString}++++${logId}`;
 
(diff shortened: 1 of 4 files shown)


src/handlers/bctl.handler.ts
```diff
@@ -34,5 +34,22 @@ export async function bctlHandler(configService: ConfigService, logger: Logger)
...
-    spawn('kubectl', kubeArgs, { stdio: [process.stdin, process.stdout, process.stderr] });
+    const kubeCommandProcess = spawn('kubectl', kubeArgs, { stdio: [process.stdin, process.stdout, process.stderr] });
+
+    kubeCommandProcess.on('close', async (code: number) => {
+        logger.debug(`Kube command process exited with code ${code}`);
+
+        if (code != 0) {
+            // Check to ensure they are using the right context
+            // Imported here since the code should usually be 0 
+            const { promisify } = require('util');
```
sidpremkumar, line 45: this needs to move

src/handlers/bctl.handler.ts
```diff
@@ -4,6 +4,8 @@ import { cleanExit } from './clean-exit.handler';
 
 const { v4: uuidv4 } = require('uuid');
 const spawn = require('child_process').spawn;
+const { promisify } = require('util');
+const exec = promisify(require('child_process').exec)
```
tflpd, line 8: I am quite convinced that using `require` is not the TS ES6 way, it is the JS way - maybe I am wrong tho

tflpd (approved):

src/handlers/bctl.handler.ts
sidpremkumar: It seems to be okay? According to this at least: https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

I could be wrong too. @lipmas or @chad-cwc would either of you know if this is okay?

chad-cwc: I'm not an expert on this either.

To me, it reads like it really depends on how the module you're using is exported....maybe?

I didn't even know syntax like this is supported:
`import zip = require("./ZipCodeValidator");`

ymarcus93: Similar discussion: https://github.com/bastionzero/zli/pull/160#discussion_r700200458

lipmas: I am also not an expert on this but I think chad is right it does depend on the module is exported but we should try and use the import syntax if its possible. In some cases it might not be possible and using `require` in those cases is fine.

But that said i think for all the standard nodejs modules we should be able to use import. Specifically for `util.Promisify` i used this in the ssm tunnel code in the zli before see below:

https://github.com/bastionzero/zli/blob/83894bc6235ac832d672cb5aa2d7aff03f120a6e/src/ssm-tunnel/ssm-tunnel.service.ts#L1
https://github.com/bastionzero/zli/blob/83894bc6235ac832d672cb5aa2d7aff03f120a6e/src/ssm-tunnel/ssm-tunnel.service.ts#L96

sidpremkumar: https://github.com/bastionzero/zli/pull/161/commits/6a2a835829342effc5a2abebb0e41207ff3a37b8

I've updated to this: 
```
import { ConfigService } from '../../src/config.service/config.service';
import { Logger } from '../../src/logger.service/logger';
import util from 'util';
import { spawn } from 'child_process';

const { v4: uuidv4 } = require('uuid');
const exec = util.promisify(require('child_process').exec)
```

PR #165, opened 2021-09-01
Ensure all vars are stored in vault

sidpremkumar:
## Description of the change

Ensure that we are using `os.Getenv()` as little as possible, and always pulling from a source of truth. We store the org info in the vault when we generate a new pub/private key. And then pull from there. We do not pull every time we need to make a message request (as that would not really be sustainable), rather we pull once when we build out key splitting object when we build the datachannel/plugin and pass those values to the function that needs them. 

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1030

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/bctl-go-daemon/bctl/agent/agent.go
+++ b/bctl-go-daemon/bctl/agent/agent.go
@@ -14,8 +14,9 @@ import (
 )
 
 var (
-	serviceUrl, orgId, clusterName string
-	environmentId, activationToken string
+	serviceUrl, orgId, clusterName   string
+	environmentId, activationToken   string
+	idpProvider, namespace, idpOrgId string
 )
 
 const (
@@ -46,7 +47,7 @@ func main() {
 	}
 
 	// Connect to the control channel
-	control, err := cc.NewControlChannel(ccLogger, serviceUrl, activationToken, orgId, clusterName, environmentId, version, controlchannelTargetSelectHandler)
+	control, err := cc.NewControlChannel(ccLogger, serviceUrl, activationToken, orgId, clusterName, environmentId, version, idpProvider, idpOrgId, namespace, controlchannelTargetSelectHandler)
 	if err != nil {
 		select {} // TODO: Should we be trying again here?
 	}
@@ -148,6 +149,9 @@ func parseFlags() error {
 	orgId = os.Getenv("ORG_ID")
 	clusterName = os.Getenv("CLUSTER_NAME")
 	environmentId = os.Getenv("ENVIRONMENT")
+	idpProvider = os.Getenv("IDP_PROVIDER")
+	idpOrgId = os.Getenv("IDP_ORG_ID")
+	namespace = os.Getenv("NAMESPACE")
 
 	// Ensure we have all needed vars
 	missing := []string{}
@@ -161,8 +165,6 @@ func parseFlags() error {
 	case clusterName == "":
 		missing = append(missing, "clusterName")
 		fallthrough
-	// case environmentId == "":
-	// 	missing = append(missing, "environmentId")
 	case activationToken == "":
 		missing = append(missing, "activationToken")
 	}
--- a/bctl-go-daemon/bctl/agent/agent.go
+++ b/bctl-go-daemon/bctl/agent/agent.go
@@ -61,7 +61,7 @@ func main() {
 	}
 
 	// Connect to the control channel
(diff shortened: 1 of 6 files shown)


bctl-go-daemon/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -221,6 +228,11 @@ func newAgent(serviceUrl string, activationToken string, agentVersion string, or
 				log.Printf("Error making post request to register agent. Error: %s. Response: %s", err, response)
 				return nil, err
 			}
+
+			// If the registration went ok, save the config
```
sidpremkumar, line 232: Feel into an issue where registration went bad but the config was already updated. So I had to manually clear the config. This change would prevent something like that from happening.

bctl-go-daemon/bzerolib/keysplitting/bzcert/bzcertverifier.go
```diff
@@ -42,25 +41,24 @@ const (
...
+		iss = getMicrosoftIssuerUrl(idpOrgId)
 	// case Custom:
 	// 	iss = customIss // Any valid iss requires a discovery document
 	default:
 		return &BZCertVerifier{} // return error
 	}
 
 	return &BZCertVerifier{
-		orgId:       orgId,
+		orgId:       idpOrgId,
```
tflpd, line 61: Why do we change this to `idpOrgId`?

sidpremkumar: because before we were using `orgId` like: `orgId := os.Getenv("IDP_ORG_ID")`. But now we save it in the config (vault) and pass it into the function as `idpOrgId`. You can see here: https://github.com/bastionzero/zli/pull/165/files/79dbf3db46321ccdcd587bfc3a1275742e7992c2#diff-57a36cc101c7b10a7b0f3ce211b49e05d1a120e2db37f98806b7fe0a553c7728R135

bctl-go-daemon/bctl/agent/agent.go
```diff
@@ -130,6 +131,9 @@ func parseFlags() {
 	orgId = os.Getenv("ORG_ID")
 	clusterName = os.Getenv("CLUSTER_NAME")
 	environmentId = os.Getenv("ENVIRONMENT")
+	idpProvider = os.Getenv("IDP_PROVIDER")
+	idpOrgId = os.Getenv("IDP_ORG_ID")
+	namespace = os.Getenv("NAMESPACE")
```
lgmugnier, line 136: I think we'll also want to remove these.  Is there a way to pass a config file to the kube deployment instead of env variables?  We can also delay that endeavor for another time and write a ticket for now

sidpremkumar: yeah we can do that, but there some pros/cons to that , and would like to discuss with you first prior. So maybe lets leave that for another day. I've created a ticket to track the discussion.

lgmugnier (approved):

bctl-go-daemon/bctl/agent/agent.go
```diff
@@ -32,7 +33,7 @@ func main() {
 	version := getAgentVersion()
 
 	// Connect to the control channel
-	control, err := cc.NewControlChannel(serviceUrl, activationToken, orgId, clusterName, environmentId, version, controlchannelTargetSelectHandler)
+	control, err := cc.NewControlChannel(serviceUrl, activationToken, orgId, clusterName, environmentId, version, idpProvider, idpOrgId, namespace, controlchannelTargetSelectHandler)
```
lgmugnier, line 36: I feel like we're working backwards with this.  Shouldn't this data be grabbed by whatever component is using it as opposed to being passed down the line?

sidpremkumar: so we pass it down here bc it could not be in the component yet, since this could be a brand new agent. So we need to pass it down from the flag -> into the `NewControlChannel` which will pass it down to `NewAgent`

lgmugnier: I see now, I missed that newAgent was in the control channel.  Although, why is it there, it seems like something the agent should be tackling and not the control channel? It seems like the only thing the control channel needs from the vault is the public key.  Would we be able to put newAgent() in agent.go, and then have a separate function which loads the vault and grabs the public key in the control channel?

sidpremkumar: https://github.com/bastionzero/zli/pull/165/commits/6d660c91080d7cc6ab9054874669c7ef19368160

PR #182, opened 2021-09-07
Fix stdout stream

sidpremkumar:
## Description of the change

This pr defaults to all logs streaming to stdout. This allows us to get nice formatted log in the dockerhub image and in our dev tools: 
2021-09-07 at 2 08 PM
2021-09-07 at 1 59 PM

I left the functionality so that we can disable this behavior in the future. But tested it by building a new image + deploying a new dev agent and it was working. 

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-XXX

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/bctl-go-daemon/bctl-agent.service
+++ b/bctl-go-daemon/bctl-agent.service
@@ -3,8 +3,8 @@ Description=Bctl-Agent
 
 [Service]
 ExecStart=/bctl-agent-files/start-dev.sh
-# StandardOutput=file:/var/log/bzero/bctl-agent.log
-# StandardError=file:/var/log/bzero/bctl-agent.log
+StandardOutput=file:/var/log/bzero/bctl-agent.log
+StandardError=file:/var/log/bzero/bctl-agent.log
 
 [Install]
 WantedBy=multi-user.target
--- a/bctl-go-daemon/bctl/agent/agent.go
+++ b/bctl-go-daemon/bctl/agent/agent.go
@@ -31,7 +31,6 @@ const (
 
 	// Disable auto-reconnect
 	autoReconnect = false
-	logFilePath   = "/var/log/cwc/bctl-agent-logs.log"
 )
 
 func main() {
@@ -39,7 +38,7 @@ func main() {
 	agentVersion := getAgentVersion()
 
 	// setup our loggers
-	logger, err := lggr.NewLogger(lggr.Debug, logFilePath)
+	logger, err := lggr.NewLogger(lggr.Debug, "")
 	if err != nil {
 		return
 	}
--- a/bctl-go-daemon/bctl/agent/agent.go
+++ b/bctl-go-daemon/bctl/agent/agent.go
@@ -31,15 +31,22 @@ const (
 
 	// Disable auto-reconnect
 	autoReconnect = false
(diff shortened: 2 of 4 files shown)


bctl-go-daemon/bctl/agent/agent.go
```diff
@@ -31,15 +31,15 @@ const (
...
+	logFilePath   = "/var/log/cwc/bctl-agent-logs.log"
 )
 
 func main() {
 	// Get agent version
 	agentVersion := getAgentVersion()
 
 	// setup our loggers
-	logger, err := lggr.NewLogger(lggr.Debug, logFilePath, false)
+	logger, err := lggr.NewLogger(lggr.Debug, logFilePath, true)
```
lipmas, line 42: Something to be aware of that if we do keep this file logger as well then its going to fill up space in the agent pod in the customers cluster. If `/var/log/cwc/` isnt on a mounted volume this will get counted as memory usage by the pod and the pod will get OOM and killed if we exceed the memory request- which i believe is currently set to 1GB

lipmas: So maybe it makes sense to not write to the log file at all and only write to stdout? Or only write to log file when we run this in dev if thats an easy switch to add?

sidpremkumar: Hmm thats a really good point

sidpremkumar: Talked with @lgmugnier and decided to just go with stdout.

lipmas (approved):

bctl-go-daemon/bctl-agent.service
```diff
@@ -3,8 +3,8 @@ Description=Bctl-Agent
 
 [Service]
 ExecStart=/bctl-agent-files/start-dev.sh
-# StandardOutput=file:/var/log/bzero/bctl-agent.log
-# StandardError=file:/var/log/bzero/bctl-agent.log
+StandardOutput=file:/var/log/bzero/bctl-agent.log
```
lgmugnier, line 6: Does this stream all standard output or just the service's output

bctl-go-daemon/bzerolib/logger/logger.go
```diff
@@ -25,28 +25,31 @@ type Logger struct {
...
 		return &Logger{
 			logger: zerolog.New(multi).With().Timestamp().Logger(),
 		}, nil
 	} else {
+		consoleWriter := zerolog.ConsoleWriter{Out: os.Stdout}
+		multi := zerolog.MultiLevelWriter(consoleWriter)
+
 		return &Logger{
-			logger: zerolog.New(logFile).With().Timestamp().Logger(),
+			logger: zerolog.New(multi).With().Timestamp().Logger(),
```
lgmugnier, line 52: If we want just json output, I think these lines become:
`return &Logger {
    logger: zerolog.New(os.Stdout).With().Timestamp.Logger(),
}`

bctl-go-daemon/bctl-agent.service
sidpremkumar: just the services

lgmugnier (approved):

PR #187, opened 2021-09-07
Add alert on creating namespace

sidpremkumar:
## Description of the change

Alert on namespace creation when generating yaml

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-XXX

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/src/handlers/generate-kube-yaml.handler.ts
+++ b/src/handlers/generate-kube-yaml.handler.ts
@@ -60,5 +60,9 @@ export async function generateKubeYamlHandler(
     } else {
         logger.info(kubeYaml.yaml);
     }
+
+    // Always let the user know they need to create the namespace
+    logger.info(`Please make sure you create the kube namespace: ${argv.namespace}`)
+
     await cleanExit(0, logger);
 }
--- a/src/cli-driver.ts
+++ b/src/cli-driver.ts
@@ -784,7 +784,7 @@ export class CliDriver
                             default: null
                         }).option('namespace', {
                             type: 'string',
-                            default: '',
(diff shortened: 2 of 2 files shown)


lgmugnier (approved):

src/handlers/generate-kube-yaml.handler.ts
```diff
@@ -60,5 +60,9 @@ export async function generateKubeYamlHandler(
     } else {
         logger.info(kubeYaml.yaml);
     }
+
+    // Always let the user know they need to create the namespace
+    logger.info(`Please make sure you create the kube namespace: ${argv.namespace}`)
```
lipmas, line 65: Any reason not to just add this to the kube yaml file directly from the backend? It should work according to https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#creating-a-new-namespace

```
apiVersion: v1
kind: Namespace
metadata:
  name: bastionzero
```

sidpremkumar: We do, but we also allow the user the option of passing `--namespace` so they can deploy the agent in any namespace they want.

lipmas: So couldnt we just use the injected parameter in that case. I mean add it to the kube yaml template: https://github.com/bastionzero/webshell-backend/blob/master/Webshell.WebApp/Configuration/kube.xml

```
apiVersion: v1
kind: Namespace
metadata:
  name: $NAMESPACE 
```

PR #193, opened 2021-09-09
Ensure that close message gets bubbled up to Daemon

sidpremkumar:
## Description of the change

Ensure that when we receive a close message from the bastion, it gets bubbled up to the daemon. We had this functionality at some point, but looks like it was lost in the many rebase's refactors that occured

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1077

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/kube-go-assets/bctl/daemon/datachannel/datachannel.go
+++ b/kube-go-assets/bctl/daemon/datachannel/datachannel.go
@@ -98,15 +98,17 @@ func NewDataChannel(logger *lggr.Logger,
 						ret.logger.Error(err)
 					}
 				}()
-			case <-ret.websocket.DoneChan:
+			case message := <-ret.websocket.DoneChan:
 				// The websocket has been closed
-				msg := "Websocket has been closed, closing datachannel"
+				msg := fmt.Sprintf("Websocket has been closed, closing datachannel: %s", message)
 				ret.logger.Info(msg)
-				cancel()
 
 				// Send a message to our done channel so kubectl can display it
-				ret.doneChannel <- msg
-				return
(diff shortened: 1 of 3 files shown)


bctl-go-daemon/bctl/daemon/datachannel/datachannel.go
```diff
@@ -98,15 +98,17 @@ func NewDataChannel(logger *lggr.Logger,
...
-				cancel()
 
 				// Send a message to our done channel so kubectl can display it
-				ret.doneChannel <- msg
-				return
+				ret.doneChannel <- message
+
+				// Do not return, so the user can see the error, just sleep forever
+				// We also do not cancel the context for the same reason
+				select {}
```
ymarcus93, line 111: So if we did `return` here the message doesn't reach the user/terminal/kubectl? Why is that? Shouldn't there be someone listening on the `doneChannel` and then displaying that to the user?

I only ask because `select {}` is essentially blocking this goroutine from ever being terminated (unless of course the daemon is killed).

sidpremkumar: We want to sleep forever so that nothing can ever get to the other code block. We want to just block this go routine till someone kills the daemon, as that is what the error message should say/that is what they will need to do. 

As for why it did not reach the kube plugin, I'm not sure. I believe it might be because the `cancel()` was being called, and that would cause the context to be canceled in the plugin as well.

tflpd (approved):

lgmugnier (approved):

PR #194, opened 2021-09-14
Require node v14 and npm v7. Update all deps

ymarcus93:
## Description of the change

+ Ran npm update to update all dependencies (respects our version restrictions in package.json)
+ Added .npmrc file. This file contains a strict check on engine which
  means if a developer tries using a node version or npm version not
  specified in package.json, then they will get a fatal error when
  running npm install
+ Remove explicit dependency on normalize-url. It's only a
  transient dependency for some other package we use.
+ Update codebuild yamls to use node v14. Use npm ci instead of npm i
  (for more deterministic builds)

Note: It would be nice to require npm 7 but our brew release process
only has access to npm 6.

## Type of change

- [x] Refactor (code change with no functionality change)

## Related JIRA tickets

Relates to JIRA: CWC-XXX

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -12,10 +12,10 @@ env:
 phases:
   install:
     runtime-versions:
-      nodejs: 12
+      nodejs: 14
     commands:
       - echo Install started on `date`
-      - npm i
+      - npm ci
       - npm i -g semver
   pre_build:
     commands:
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bastionzero/zli",
-  "version": "5.1.14",
+  "version": "5.1.15",
   "description": "BastionZero cli",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
   "types": "dist/src/index.d.ts",
   "bin": "dist/src/index.js",
   "engines": {
-    "node": ">0.12.0",
-    "npm": ">7.0.0"
+    "node": "^14",
+    "npm": ">=6"
   },
   "scripts": {
     "start": "ts-node src/index.ts",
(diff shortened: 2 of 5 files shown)


buildspec.yml
```diff
@@ -12,7 +12,7 @@ env:
 phases:
   install:
     runtime-versions:
-      nodejs: 12
+      nodejs: 14
     commands:
       - echo Install started on `date`
       - npm i
```
lipmas, line 18: Hey while we are making changes to the CI/CD to bump the node version can we also switch the `npm i` commands to `npm ci` which will ensure we are installing packages according to package-lock.json. https://docs.npmjs.com/cli/v7/commands/npm-ci

ymarcus93: Yes that sounds good. We should be respecting the package-lock so that we get more reproducible builds

ymarcus93: Fixed in 404df93

ymarcus93:
I tested all the zli commands. Everything seems to be working with node 14. The addition of the `.npmrc` file will make it so `npm` complains if using a non-supported `node` or `npm` version (as specified in `engine` block of `package.json`)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bastionzero/zli",
-  "version": "5.1.3",
+  "version": "5.1.4",
```
sidpremkumar, line 3: Really should be a major version bump, but since this would just force everyone using the zli to upgrade for no real reason, I think its okay to make it minor

sidpremkumar (approved):

lipmas (approved):

ymarcus93:
Made small fixup to `package-lock`

lipmas (approved):

ymarcus93:
Just tested everything again on b908c4d. Looks good

lipmas (approved):
LGTM

PR #199, opened 2021-09-20
Always perform an undefined check for kubeConfig

sidpremkumar:
## Description of the change

Had a bug where we would throw an error if the kubeconfig was not created in the zli config. This PR fixes that by checking if kubeConfig != undefined first. 

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1101

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development

### DIFF
--- a/src/handlers/bctl.handler.ts
+++ b/src/handlers/bctl.handler.ts
@@ -11,7 +11,7 @@ const execPromise = util.promisify(exec);
 export async function bctlHandler(configService: ConfigService, logger: Logger, listOfCommands: string[]) {
     // Check if daemon is even running
     const kubeConfig = configService.getKubeConfig();
-    if (kubeConfig !== undefined && kubeConfig['localPid'] === null) {
+    if (kubeConfig !== null && kubeConfig['localPid'] === null) {
         logger.warn('No Kube daemon running');
         await cleanExit(1, logger);
     }
--- a/src/handlers/bctl.handler.ts
+++ b/src/handlers/bctl.handler.ts
@@ -11,7 +11,7 @@ const execPromise = util.promisify(exec);
 export async function bctlHandler(configService: ConfigService, logger: Logger, listOfCommands: string[]) {
     // Check if daemon is even running
     const kubeConfig = configService.getKubeConfig();
-    if (kubeConfig['localPid'] == null) {
+    if (kubeConfig !== undefined && kubeConfig['localPid'] === null) {
         logger.warn('No Kube daemon running');
         await cleanExit(1, logger);
     }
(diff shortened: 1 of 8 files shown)


src/handlers/bctl.handler.ts
```diff
@@ -11,7 +11,7 @@ const execPromise = util.promisify(exec);
 export async function bctlHandler(configService: ConfigService, logger: Logger, listOfCommands: string[]) {
     // Check if daemon is even running
     const kubeConfig = configService.getKubeConfig();
-    if (kubeConfig['localPid'] == null) {
+    if (kubeConfig !== undefined && kubeConfig['localPid'] === null) {
```
sidpremkumar, line 14: 2021-09-20 at 10 52 AM

src/handlers/logout.handler.ts
```diff
@@ -12,7 +12,7 @@ export async function logoutHandler(configService: ConfigService, logger: Logger
 
     logger.info('Closing any existing Kube Proxy Connections');
     const kubeConfig = configService.getKubeConfig();
-    if (kubeConfig !== null && kubeConfig['localPid'] !== null) {
+    if (kubeConfig !== undefined && kubeConfig['localPid'] !== null) {
```
lipmas, line 15: So now that there is a default value for the kubeConfig you should be able to simplify and remove everywhere you check `kubeConfig !== undefined`. kubeConfig will always be defined now.

sidpremkumar: Yes you are right, sorry I was unsure if we could remove them if people had already generated a config.

sidpremkumar: https://github.com/bastionzero/zli/pull/199/commits/a407ea80c06aaa0be3d4267855441e44f478b29d

lipmas: If people already generated a config then it wont be null yeah. However they might not get the new (null) default values for everything. So like any values of the config that were unset  will still be undefined for these users.

I think this is ok since you were explicitly setting `localPid` to null already and the only other null check you do is for `keyPath` but that should always be set for users who already generated a kubeConfig right?

sidpremkumar: yes keyPath will be set for anyone who has generate a kubeConfig

sidpremkumar: I'll just make it undefined in my config and test one sec

sidpremkumar: 2021-09-20 at 11 48 AM
2021-09-20 at 11 48 AM

It looks like since its undefined not "not set" if that makes sense. 
:/ looks like the null checks might be needed.

sidpremkumar: 2021-09-20 at 12 13 PM
Latest commit seems to fix this

sidpremkumar: Alternatively, we can look into https://www.npmjs.com/package/conf `migrations`

lipmas (approved):

PR #200, opened 2021-09-20
Remove idp [CWC-1090]

tflpd:
## Description of the change

Removed the idp positional argument from the zli login. Now the user gets redirected to the webapp in order to choose their provider

In order to test execute `zli login`. Please test with login with MFA as well

Related PR: https://github.com/bastionzero/webshell-backend/pull/844

## Type of change

- [X] New feature (non-breaking change that adds functionality)

## Related JIRA tickets

Relates to JIRA: CWC-1090

Release Notes:

## Potential Security Impacts

This essentially just sends user's idp choice from the webapp to the zli, there should not be any security issues related

### Development

### DIFF
--- a/src/services/oauth/oauth.service.ts
+++ b/src/services/oauth/oauth.service.ts
@@ -9,27 +9,32 @@ import { loginHtml } from './templates/login';
 import { logoutHtml } from './templates/logout';
 import { cleanExit } from '../../handlers/clean-exit.handler';
 import { parse as QueryStringParse } from 'query-string';
+import { idpRedirect } from './templates/idpRedirect';
+import { parseIdpType } from '../../utils';
 
 export class OAuthService implements IDisposable {
     private server: http.Server; // callback listener
     private host: string = 'localhost';
     private logger: Logger;
+    private oidcClient: Client;
+    private codeVerifier: string;
+    private nonce: string;
 
     constructor(private configService: ConfigService, logger: Logger) {
         this.logger = logger;
     }
 
     private setupCallbackListener(
-        client: Client,
-        codeVerifier: string,
         callback: (tokenSet: TokenSet) => void,
         onListen: () => void,
-        resolve: (value?: void | PromiseLike<void>) => void,
-        expectedNonce?: string
+        resolve: (value?: void | PromiseLike<void>) => void
     ): void {
 
         const requestListener: RequestListener = async (req, res) => {
-            res.writeHead(200, { 'content-type': 'text/html' });
+            res.writeHead(200, {
+                'Access-Control-Allow-Origin': '*',
+                'content-type': 'text/html',
+              });
 
             // Example of request url string
             // /login-callback?param=...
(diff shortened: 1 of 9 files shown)


src/utils.ts
```diff
@@ -49,6 +49,23 @@ export function parsePolicyType(policyType: string) : PolicyType
     }
 }
 
+export function parseIdpType(idp: string) : IdP
```
lipmas, line 52: Maybe its simpler if you just define a enum (with string values) in `webshell-common-ts` and use it both here and when sending the idp type in the webapp

tflpd: So, while it is indeed good to have the idp as a shared type - added that - I needed this to parse the query param and get either `undefined` or `IdP`. Is there a better way of getting `undefined` when the passed param is not part of the `enum`?

lipmas: Yeah i think you can just make the `default` case in the switch block return undefined in this case.

lipmas: And have this function take the idp type `idp: IdentityProvider`

src/services/oauth/oauth.service.ts
```diff
@@ -46,13 +51,45 @@ export class OAuthService implements IDisposable {
             }
 
             switch (urlParts[0]) {
+            case '/webapp-callback':
+                res.write(idpRedirect);
+                res.end();
+
+                // Prepare config for a new login
+                const provider = parseIdpType(queryParams.idp as string);
```
lipmas, line 59: This can cast the query para to `IdentityProvider` type

src/utils.ts
tflpd: Oh, yeah okok so we have to keep the utils function that was my point. okok perf

lipmas: oh yup sorry if that wasnt clear but yeah keep this utils function but use the type instead of a regex

src/services/oauth/oauth.service.ts
tflpd: 42f1ca2

src/services/oauth/oauth.service.ts
```diff
@@ -46,13 +51,45 @@ export class OAuthService implements IDisposable {
...
+                    await cleanExit(1, this.logger);
+                }
+                await this.configService.loginSetup(provider);
+
+                // Setup the oidc client for a new login
+                await this.setupClient();
+                this.codeVerifier = generators.codeVerifier();
+                const code_challenge = generators.codeChallenge(this.codeVerifier);
+
+                await open(this.getAuthUrl(code_challenge));
```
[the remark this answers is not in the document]
tflpd: 42f1ca2

src/utils.ts
tflpd: 42f1ca2

src/utils.ts
```diff
@@ -49,18 +50,13 @@ export function parsePolicyType(policyType: string) : PolicyType
...
-
     switch (idp.toLowerCase()) {
-    case IdP.Google.toLowerCase():
-        return IdP.Google;
-    case IdP.Microsoft.toLowerCase():
-        return IdP.Microsoft;
+    case IdentityProvider.Google.toLowerCase():
+        return IdentityProvider.Google;
+    case IdentityProvider.Microsoft.toLowerCase():
+        return IdentityProvider.Microsoft;
```
chad-cwc: Do we want to keep the `toLowerCase` here?

src/utils.ts
```diff
@@ -49,6 +50,18 @@ export function parsePolicyType(policyType: string) : PolicyType
     }
 }
 
+export function parseIdpType(idp: IdentityProvider) : IdentityProvider
+{
+    switch (idp.toLowerCase()) {
+    case IdentityProvider.Google.toLowerCase():
+        return IdentityProvider.Google;
+    case IdentityProvider.Microsoft.toLowerCase():
```
chad-cwc, line 58: Do we want to keep the `toLowerCase` here?

tflpd: 2517d19

chad-cwc (approved):

PR #212, opened 2021-09-28
Update error messages on bctl handler

sidpremkumar:
## Description of the change

Couple of things here:
* First adds a check if the command failed to see if the user has kubectl installed
* Then updates the kube config check to `.trim()` which I believe was causing that block to always execute
* Finally add a check to see if the kube daemon quit unexpectedly

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1106

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


src/handlers/bctl.handler.ts
```diff
@@ -43,11 +43,30 @@ export async function bctlHandler(configService: ConfigService, logger: Logger,
         logger.debug(`Kube command process exited with code ${code}`);
 
         if (code != 0) {
-            // Check to ensure they are using the right context
-            const currentContext = await execPromise('kubectl config current-context ');
+            // Check if the daemon has quit
+            if (!require('is-running')(kubeConfig['localPid'])) {
```
lipmas, line 47: also check if `kubeConfig['localPid'] == null` and dont try to check if its running in this case.

lipmas: nit: move require import to top of file

lipmas (approved):

src/handlers/bctl.handler.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/212/commits/6123db2f39da843f27dbded4da8d7573e2bd3e90

PR #213, opened 2021-09-30
add refresh id token logic to zli

lgmugnier:
## Description of the change

The daemon did not refresh the id token, so when there was no zli activity, the id token would expire without any way to refresh it.  Now, the daemon refreshes the token whenever it sends a new Syn, which is when a datachannel is opened and then whenever it receives a keysplitting error from the agent (up to a certain number of times).

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-912

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


kube-go-assets/bctl/daemon/keysplitting/keysplitting.go
```diff
@@ -179,6 +182,11 @@ func (k *Keysplitting) BuildSyn(action string, payload []byte) (ksmsg.Keysplitti
 }
 
 func (k *Keysplitting) buildBZCert() (bzcrt.BZCert, error) {
+	// update the id token by calling the passed in zli command
+	if err := exec.Command("bash", "-c", k.refreshTokenCommand).Run(); err != nil {
```
sidpremkumar, line 186: So this assumes that the shell is bash, can we remove "bash" and "-c" and will the command still work? What if the user only has `sh` installed or something

lgmugnier: testing...

src/handlers/tunnel/start-kube-daemon.handler.ts
```diff
@@ -160,6 +164,21 @@ async function copyExecutableToLocalDir(logger: Logger, configPath: string): Pro
     // Helper function to copy the Daemon executable to a local dir on the file system
     // Ref: https://github.com/vercel/pkg/issues/342
 
+    const WINDOWS_DAEMON_PATH : string = 'kube-go-assets/bctl/daemon/daemon-windows';
+    const LINUX_DAEMON_PATH   : string = 'kube-go-assets/bctl/daemon/daemon-linux';
+    const MACOS_DAEMON_PATH   : string = 'kube-go-assets/bctl/daemon/daemon-macos';
+
+    let prefix = '';
```
sidpremkumar, line 171: So this is a common patter we follow, can we make a helper function called ~ `isPkgProcess()` that will return the result of `pkgProcess`

```
function isPkgProcess(): bool {
const process1 = <any>process;
return process1.pkg;
}
```

lipmas: yes sorry i was planning on doing that then forgot :facepalm: :

kube-go-assets/bctl/daemon/keysplitting/keysplitting.go
lgmugnier: experimenting...

lgmugnier: 2384f65

lipmas (approved):

PR #224, opened 2021-10-06
Add access token header in groups/list endpoint for microsoft users

lipmas:
## Description of the change

This is needed in order for the microft groups integration to workproperly.  The backend needs an access token with the required scopes for fetching group information from microsoft graph API. This only applies to users who have gone through the azure groups integration flow in the webapp and whose access token contains the scopes User.Read.All and Group.Read.All. If the access token does not contain these scopes we can skip passing the token to the backend and the groups/list endpoint will simply return an empty list of groups.

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1145

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


src/services/groups/groups.service.ts
```diff
@@ -12,7 +16,32 @@ export class GroupsService extends HttpService
...
+            const scopes = tokenSet.scope.split(' ');
+            const adminConsentAcquired = MicrosoftAdminScopes.every(scope => scopes.includes(scope));
+
+            if(adminConsentAcquired) {
+                const accessToken = tokenSet.access_token;
+
+                if(accessToken) {
+                    extraHeaders['AccessToken'] = accessToken;
+                } else {
+                    this.logger.warn('No access token in token set cannot fetch microsoft groups');
```
tflpd, line 39: we may want to rephrase this? I am not sure I follow the message

tflpd (approved):
tested quickly, I think it is okay

src/services/groups/groups.service.ts
lipmas: yeah honestly this branch should never really happen just wanted to add a log for completeness...since as far as im aware the access token should always be included in the token set from the oidc client and if the token set was not set  we would have failed much earlier in the oauth middleware.

PR #229, opened 2021-10-13
Registration Debugability

lgmugnier:
## Description of the change

We are constantly running into issues with registration.  errors aren't being reported correctly and it's very difficult to debug because we have very little visibility into what's failing on the agent.

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1138, CWC-1149, CWC-1151

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


kube-go-assets/bctl/agent/agent.go
```diff
@@ -198,12 +213,82 @@ func getAgentVersion() string {
...
+		},
+	}
+
+	verbs := make(map[string]bool)
+	rules, rerr := clientset.AuthorizationV1().SelfSubjectRulesReviews().Create(context.TODO(), sar, metav1.CreateOptions{})
+	if rerr != nil {
+		return fmt.Errorf("could not review service account permissions: %s", rerr)
+	}
+
+	for _, rule := range rules.Status.ResourceRules {
```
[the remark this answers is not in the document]
lgmugnier: oh we should, I totally missed those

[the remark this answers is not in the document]
lgmugnier: 16f6536

kube-go-assets/bctl/agent/agent.go
```diff
@@ -45,30 +44,38 @@ func main() {
 		return
 	}
 	logger.AddAgentVersion(agentVersion)
+	logger.Info(fmt.Sprintf("BastionZero Agent version %s starting up...", agentVersion))
```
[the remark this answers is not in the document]
sidpremkumar: Similar comment across this PR

kube-go-assets/bctl/agent/vault/vault.go
```diff
@@ -81,6 +82,8 @@ func LoadVault() (*Vault, error) {
 					}, nil
 				}
 			} else {
+				secretsClient := clientset.CoreV1().Secrets(os.Getenv("NAMESPACE"))
+				log.Printf("secretsClient: %+v", secretsClient)
```
sidpremkumar, line 86: Do we need this log statement?

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -332,7 +332,7 @@ func (w *Websocket) Connect() {
 			w.logger.Info("Error when trying to agree on version for SignalR!")
 			w.client.Close()
 		} else {
-			w.logger.Infof("Connection Successful!")
+			w.logger.Infof("Connection successful!")
```
sidpremkumar, line 335: We don't need to use `Infof` here

kube-go-assets/bctl/agent/rbac/rbac.go
```diff
@@ -0,0 +1,146 @@
...
+						missing[apiGroupReq][resourceReq][verbReq] = false
+					}
+				}
+			}
+		}
+	}
+
+	// iterate through the list of provided rules and requirements, and delete from the dictionary of missing permissions.
+	for _, rule := range rules.Status.ResourceRules {
+		for _, apiGroup := range rule.APIGroups {
```
sidpremkumar, line 94: How come we need this double for loop, isn't the chain `ApiGroups` -> `Resources` -> `Verbs`?

I believe this is the object we are using: https://pkg.go.dev/k8s.io/api/authorization/v1#ResourceRule

So can we loop over all those object like you did: 
```go
for _, apiGroup := range rule.APIGroups { // Loop over all api groups the SA has
   for _, resource := range rule.Resources { // Loop over every resource in that API group
    // Here just ensure that, rule.Verbs == missing[apiGroup][resourceReq]
  }
}
```

Totally could be misunderstanding something

kube-go-assets/bctl/agent/rbac/rbac.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// all required rules
+	reqs, err := loadConfig()
+	if err != nil {
+		return err
+	}
+
+	// create a dictionary so we can be specific about the permissions they're missing
+	missing := make(map[string]map[string]map[string]bool)
```
sidpremkumar, line 73: Nit: We could extract these complicated for loops to helper functions, to clean up the main logic of this function,

kube-go-assets/bctl/agent/rbac/rbac.go
lgmugnier: So the ResourceRules is made up of several objects of type ResourceRule:
`for _, rule := range rules.Status.ResourceRules {`
then we need to grab a singular apigroup, resource, and verb.  Each rule can have multiple for each so we need every permutation.  Keep in mind that there is no nice way to check for membership in go:
`for _, apiGroup := range rule.APIGroups {
	for _, resource := range rule.Resources {
		for _, verb := range rule.Verbs {`
Also keep in mind that a rule may have more or fewer verbs than what we're looking for so we want to match each.  That way we'll know if our list is a subset and if it isn't, we'll be able to output exactly which ones are missing.

kube-go-assets/bzerolib/channels/websocket/websocket.go
lgmugnier: 18d0484

kube-go-assets/bctl/agent/agent.go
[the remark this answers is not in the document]
lgmugnier: 18d0484

kube-go-assets/bctl/agent/rbac/rbac.go
lgmugnier: f798d3c

sidpremkumar:
Tested and ran into some issues: 
2021-12-09 at 1 05 PM

Seems like this line isn't working: https://github.com/bastionzero/zli/blob/reg-debugability/kube-go-assets/bctl/agent/rbac/rbac.go#L150

lgmugnier:
> Tested and ran into some issues: 2021-12-09 at 1 05 PM
> 
> Seems like this line isn't working: https://github.com/bastionzero/zli/blob/reg-debugability/kube-go-assets/bctl/agent/rbac/rbac.go#L150

This was due to an ordering bug in how I was deleting items from the dictionary, mainly that I would check for an empty apigroup dictionary before deleting the resource, so some dictionaries were obviously not being deleted. d326e2c

sidpremkumar:
Tested and all working! 
2021-12-10 at 2 23 PM

sidpremkumar (approved):

PR #232, opened 2021-10-14
Quickstart refresh

ymarcus93:
## Description of the change

**Summary:** Clean up quickstart user experience

**Notable changes**:
- Removal of resolving invalid SSH configs at start
- Removal of "Validating SSH keys" messages. Additionally, we don't even let the user pick the host if the key path does not exist. Note: We still have the additional prompt where the SSH key is encrypted on-disk--but that's about it from the user's perspective.
- Assume `Port 22` if not specified in config
- Add value proposition messages throughout the interaction: e.g. welcome message, tips on using zli, links to documentation
- Clear indication of remaining steps. Clear console on each new step (with scrollback still possible). New progress bar during SSH to run autodiscovery+poll for online step.
  - Adds `ora` as dependency
- Login is now part of `quickstart`. New users do not need to run `zli login`. `zli quickstart` will handle the case where they have not logged in yet.
- Optional transcript/log of quickstart actions is saved as a separate file. This is done by new `consoleAndTranscript` service.
- MFA in `login` has been made fully interactive. We still support the old behavior of passing the `--mfa` flag though.
  - This change is needed in case the user is running `quickstart` as a logged in user of an org which as MFA. The old `login` code would fatally exit.

**Not related to quickstart but in this PR:**
- Handle case where `string` is returned in catch-all `fail()`. See CWC-1158 for more information.

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1134

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


src/handlers/quickstart/quickstart-handler.ts
```diff
@@ -37,94 +97,152 @@ async function validateQuickstartArgs(argv: yargs.Arguments<quickstartArgs>) {
...
+            await cleanExit(1, logger);
         }
     }
 
+    // New step so clear screen
+    clearScreen();
+
+    // We cannot create the MixpanelService until the user has logged in
+    const mixpanelService = new MixpanelService(configService);
+    mixpanelService.TrackCliCommand(version, 'quickstart', [])
```
ymarcus93, line 213: Probably should pass process args here (instead of empty `[]`) to keep consistent with old behavior

src/handlers/quickstart/quickstart-handler.ts
```diff
@@ -2,19 +2,79 @@ import { ConfigService } from '../../services/config/config.service';
...
+Our quickstart installer is a fast and easy method for you to try BastionZero using your existing SSH configuration.
+We will use the following information supplied by your SSH configuration file to onboard the targets:\n
+Host ${chalk.bold.yellowBright('your_host_name')}
+    HostName ${chalk.bold.yellowBright('your.ip.address')}
+    User ${chalk.bold.yellowBright('your_username')}
+    IdentityFile ${chalk.bold.yellowBright('path/to/your/.ssh/.pem/file')}
+    Port ${chalk.bold.yellowBright('specify_port_number_if_desired')}
+`;
+
+const loginMessage = `Below, when you press the key to continue a browser window will appear.\nAuthenticate yourself with Google or Microsoft based on your identity provider. Doing so will automatically create your BastionZero account.\nPress any key to continue...`;
```
lipmas, line 35:
```suggestion
const loginMessage = `Below, when you press the key to continue a browser window will appear.\nLogin with your SSO provider. Doing so will automatically create your BastionZero account.\nPress any key to continue...`;
```

Not a huge deal but maybe we should future-proof the wording here to not be specific to google/microsoft

src/services/consoleWithTranscript/consoleWithTranscript.service.ts
```diff
@@ -0,0 +1,79 @@
...
+        this.transcript = [];
+    }
+
+    public getTranscript(): readonly TranscriptMessage[] {
+        return this.transcript;
+    }
+
+    public info(msg: string, pushToTranscript: boolean = true): void {
+        if (this.defaultColor)
+            msg = this.defaultColor(msg);
```
lipmas, line 22: Im not following how this is working. It seems you do set a default color [here](https://github.com/bastionzero/zli/blob/a884b0a5b229ce72101d028a8aac457468880afe/src/handlers/quickstart/quickstart-handler.ts#L142) but then in most references to this function you also pass in a msg that is wrapped in a color (for example [here](https://github.com/bastionzero/zli/blob/a884b0a5b229ce72101d028a8aac457468880afe/src/handlers/quickstart/quickstart-handler.ts#L68)). What happens when we wrap the message with multiple chalk calls? Does only the inner one get displayed?

src/services/consoleWithTranscript/consoleWithTranscript.service.ts
```diff
@@ -0,0 +1,79 @@
...
+            case 'warn':
+                this.attachedLogger.warn(strippedMsg);
+                break;
+            case 'error':
+                this.attachedLogger.error(strippedMsg);
+                break;
+            default:
+                const exhaustiveCheck: never = msgType;
+                throw new Error(`Unhandled type: ${exhaustiveCheck}`);
+            }
```
lipmas, line 76: I know this is currently unused but what is the intended purpose of this? Is it in case we wanted to log the quickstart console/transcript messages to the zli log file as well? This just seems overly convoluted and not that useful since we have a transcript of the quickstart output anyway. IMO we can probably just remove this and if we did it would also simplify this service since we wouldnt have to care about the log level of the messages.

src/services/consoleWithTranscript/consoleWithTranscript.service.ts
ymarcus93: Yes the inner one gets displayed and overwrites the outer one. `chalk` supports nested color decorations. There is a default color of `magenta` that is set in the `consoleWithTranscript` service, as you pointed out, but others can override that.

In the case of `welcomeMessage` (the first message quickstart users see), the default color is overridden to `blue` because the entire string is wrapped in a `chalk.blue`:

https://github.com/bastionzero/zli/blob/a884b0a5b229ce72101d028a8aac457468880afe/src/handlers/quickstart/quickstart-handler.ts#L161-L162

However, this too has nested `chalk` calls. For example, we highlight some parts of the message with `chalk.bold.yellowBright`:

https://github.com/bastionzero/zli/blob/a884b0a5b229ce72101d028a8aac457468880afe/src/handlers/quickstart/quickstart-handler.ts#L25-L33

So it looks something like this when it reaches the `consoleWithTranscript` service: `chalk.magenta(chalk.blue(welcomeMsg + chalk.bold.yellowBright(some text) + rest of welcome msg + ...)))`

src/services/oauth/oauth.service.ts
```diff
@@ -240,44 +241,58 @@ export class OAuthService implements IDisposable {
...
+     * function will exit the running process if any error occurs or if the user
+     * is not logged in (i.e. tokenSet not found in config).
+     * @returns The current OIDC id_token
+     */
+    public async getIdTokenAndExitOnError(): Promise<string> {
+        let idToken: string;
+        try {
+            idToken = await this.getIdToken();
+        } catch (e) {
+            if (e instanceof errors.RPError || e instanceof errors.OPError) {
```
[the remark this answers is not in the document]
ymarcus93: How about this?: 3628f538a46d58363cafff6bcfab7187f6eb481f

src/handlers/quickstart/quickstart-handler.ts
ymarcus93: ea4c2f79e5b8e0be225950fc01ed85ad70f93470

src/services/consoleWithTranscript/consoleWithTranscript.service.ts
ymarcus93: Fixed: 6ae064fabe832fe8d36fa9edbd0ee481bd074ec6

PR #234, opened 2021-10-20
Datachannel Parallelization

lgmugnier:
## Description of the change

We need to decouple the datachannel from the websocket so that both can be scaled up/down independently of each other.  It will also make it so that if a websocket breaks, it doesn't immediately destroy the keysplitting session in case the client wants to reconnect.  This requires a major refactor of the current kube agent which includes:

1. [CWC-1170](https://commonwealthcrypto.atlassian.net/browse/CWC-1170) Moving the HTTP Server out of the Kube Daemon plugin and into a higher-level place so that we can add in logic to create, destroy, and manage websockets and datachannels.
2. [CWC-1171](https://commonwealthcrypto.atlassian.net/browse/CWC-1171) We need to do something similar with the Control Channel for the agent so that it has a parallel ability to create, destroy, and manage websockets and datachannels.
3. [CWC-1172](https://commonwealthcrypto.atlassian.net/browse/CWC-1172) Tombs are a better way for us to internally manage go routines so we should include that library in this refactor.
4. [CWC-1173](https://commonwealthcrypto.atlassian.net/browse/CWC-1173) We need to formalize websocket and datachannel creation and destruction so that we can start them up and shut them down (and recover from unexpected closes) without memory leaks and without inconveniencing the user.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related JIRA tickets

Relates to JIRA: CWC-1169, CWC-1170, CWC-1171, CWC-1172, CWC-1173

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,243 @@ package controlchannel
...
-		return &ControlChannel{}, err
+		connections: make(map[string]wsMeta),
 	}
 
-	control := ControlChannel{
-		websocket:          wsClient,
-		NewDatachannelChan: make(chan NewDatachannelMessage),
-		logger:             logger,
-	}
+	websocket.Subscribe("", control) // this is a bit of a hack, but bastion does not currently have the concept of a channel id
```
sidpremkumar, line 73: So bit confused, currently are we setting it to `""` since there will only be one datachannel and websocket?

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,259 @@ const (
...
-		// Update our object to let others know we have subscribed
-		w.subscribed = true
 
-		// Listener for any messages that need to be sent
-		go func() {
-			for {
+	// Listener for any messages that need to be sent
+	go func() {
+		for {
+			for ws.ready {
```
sidpremkumar, line 117: Do we need the double for here?

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,259 @@ const (
...
-		if rawMessage[len(rawMessage)-1] == signalRMessageTerminatorByte {
-			rawMessage = rawMessage[0 : len(rawMessage)-1]
+		if messages, err := w.unwrapSignalR(rawMessage); err != nil {
+			return err
+		} else {
+			for _, message := range messages {
+				switch message.Target {
+				case "CloseConnection":
+					rerr := errors.New("closing message received; websocket closed")
+					w.shabtis.Kill(rerr) // TODO: implement reconnect flow
```
sidpremkumar, line 184: important todo that needs to get done before merge, + check against `autoReconnect`

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
```diff
@@ -16,56 +17,65 @@ import (
...
-	r.RequestChannel <- wrapStartPayload(isTty, r.requestId, r.logId, request.URL.Query()["command"], request.URL.String())
+	e.outputChan <- wrapStartPayload(isTty, e.requestId, e.logId, request.URL.Query()["command"], request.URL.String())
 
 	// Define our function to process each message
 	seqNumber := 0
 	processMessage := func(content []byte) bool {
 		// Check for agent-initiated end e.g. user typing 'exit'
 		if string(content) == kubeexec.EscChar {
-			r.logger.Info("stream ended")
+			e.logger.Info("stream ended")
```
sidpremkumar, line 78: maybe exec stream ended so we know what type of stream we are talking about

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
```diff
@@ -76,35 +86,45 @@ func (r *ExecAction) InputMessageHandler(writer http.ResponseWriter, request *ht
...
 		for {
 			select {
-			case <-r.ctx.Done():
+			case <-tmb.Dying():
 				return
-			case streamMessage := <-r.streamChannel:
+			case streamMessage := <-e.streamInputChan:
 				contentBytes, _ := base64.StdEncoding.DecodeString(streamMessage.Content)
 
+				// Check for agent-initiated end e.g. user typing 'exit'
```
sidpremkumar, line 100: this is handled inside `processMessage` I believe now

kube-go-assets/bctl/daemon/plugin/kube/actions/restapi/restapi.go
```diff
@@ -14,37 +15,45 @@ import (
...
-	streamResponseChannel chan smsg.StreamMessage
-	logger                *lggr.Logger
-	ctx                   context.Context
+	logger *lggr.Logger
+
+	requestId       string
+	logId           string
+	commandBeingRun string
+
+	// channels for sending and recieving messages
```
sidpremkumar, line 24: receiving

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
```diff
@@ -16,56 +17,65 @@ import (
...
+
+	// input and output channels relative to this plugin
+	outputChan      chan plgn.ActionWrapper
+	streamInputChan chan smsg.StreamMessage
+	ksInputChan     chan plgn.ActionWrapper
 }
 
-func NewExecAction(ctx context.Context,
-	logger *lggr.Logger,
+func NewExecAction(logger *lggr.Logger,
```
sidpremkumar, line 32: we should keep this consistent with the agent code and call this function `New`

kube-go-assets/bctl/daemon/plugin/kube/actions/stream/stream.go
```diff
@@ -20,43 +21,59 @@ const (
...
+	ksInputChan     chan plgn.ActionWrapper
+
 	expectedSequenceNumber int
 	outOfOrderMessages     map[int]smsg.StreamMessage
 	writer                 http.ResponseWriter
 }
 
-func NewStreamAction(ctx context.Context,
-	logger *lggr.Logger,
+func NewStreamAction(logger *lggr.Logger,
```
sidpremkumar, line 40: similar this should read `New`

kube-go-assets/bctl/daemon/plugin/kube/actions/restapi/restapi.go
```diff
@@ -14,37 +15,45 @@ import (
...
+	outputChan chan plgn.ActionWrapper
+	inputChan  chan plgn.ActionWrapper
 }
 
-func NewRestApiAction(ctx context.Context,
-	logger *lggr.Logger,
+func NewRestApiAction(logger *lggr.Logger,
```
sidpremkumar, line 29: similar this should read `New`

kube-go-assets/bctl/daemon/plugin/kube/utils/utils.go
```diff
@@ -6,15 +6,12 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+
+	"github.com/google/uuid"
 )
 
-// Helper function to extract headers from a http request
-func GetHeaders(headers http.Header) map[string][]string {
```
sidpremkumar, line 12: This I believe is still used, so dont wanna delete that.

kube-go-assets/bctl/daemon/plugin/kube/kube.go
```diff
@@ -35,118 +28,191 @@ type JustRequestId struct {
...
+	var actOutputChan chan plgn.ActionWrapper
+
+	switch KubeDaemonAction(action) {
+	case Exec:
+		act, actOutputChan = exec.NewExecAction(actLogger, requestId, logId, command)
+	case Stream:
+		act, actOutputChan = stream.NewStreamAction(actLogger, requestId, logId, command)
+	case RestApi:
+		act, actOutputChan = rest.NewRestApiAction(actLogger, requestId, logId, command)
+	default:
```
sidpremkumar, line 175: going to need portforward somewhere here

kube-go-assets/bzerolib/channels/agentmessage/agentmessage.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+type AgentMessage struct {
+	ChannelId      string `json:"ChannelId"` // acts like a session id to tie messages to a keysplitting hash chain
+	MessageType    string `json:"messageType"`
+	SchemaVersion  string `json:"schemaVersion" default:"1.0"`
+	MessagePayload []byte `json:"messagePayload"`
+}
+
+// func New(channelId string, messageType string, messagePayload []byte) AgentMessage {
```
sidpremkumar, line 19: can we remove this?

kube-go-assets/bzerolib/logger/logger.go
```diff
@@ -61,25 +59,25 @@ func (l *Logger) AddDaemonVersion(version string) {
 
 // TODO: instead of assigning random uuid, the control channel and data channel should both use
```
sidpremkumar, line 60: Can we remove this TODO now? We still have to use the connection Id, but in terms of the work here this todo might not be relevent anymore.

sidpremkumar: Also there is some code at the bottom of this file that I believe can be removed,

```go
// ??
// func (l *Logger) Fatal(err error) {
// 	l.logger.Fatal().
// 		Msg(err.Error())
// }
```

since we are not using Fatel? Otherwise we can keep it if we think we will use it in the future

sidpremkumar (changes requested):

kube-go-assets/bzerolib/channels/websocket/websocket.go
lgmugnier: yup. the idea is that the first for is to loop infinitely and the second for is to loop finitely only when the websocket is ready otherwise, it will stop listening to the channels and only start up again when the websocket is ready.  At least that's what I was going for, maybe I was wrong.

kube-go-assets/bctl/agent/agent.go
```diff
@@ -45,93 +48,94 @@ func main() {
...
 
 	// Sleep forever because otherwise kube will endlessly try restarting
 	// Ref: https://stackoverflow.com/questions/36419054/go-projects-main-goroutine-sleep-forever
 	select {}
 }
 
-func startDatachannel(logger *lggr.Logger, message cc.NewDatachannelMessage) {
+func startControlChannel(logger *lggr.Logger, agentVersion string) error {
+	// Load in our saved config
+	config, _ := vault.LoadVault()
```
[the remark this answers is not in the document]
lgmugnier: this is addressed in my other PR for reg-debuggability https://github.com/bastionzero/zli/pull/229

kube-go-assets/bctl/agent/agent.go
```diff
@@ -45,93 +48,94 @@ func main() {
...
-	// Add our token to our params
-	params := make(map[string]string)
-	params["daemon_connection_id"] = message.ConnectionId
-	params["token"] = message.Token
+	// Make and add our params
+	params := map[string]string{
+		"public_key":    config.Data.PublicKey,
+		"agent_version": agentVersion,
+
+		// Why do we need these?  Can we remove them?
```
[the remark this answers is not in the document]
lgmugnier: let's do a full walk through of these at a later date

kube-go-assets/bctl/agent/agent.go
```diff
@@ -45,93 +48,94 @@ func main() {
...
+		"environment_id": environmentId,
+	}
 
-	// Create our response channels
-	// TODO: WE NEED TO SEND AN INTERRUPT CHANNEL TO DATACHANNEL FROM CONTROL
-	// or pass a context that we can cancel from the control channel??
-	dc.NewDataChannel(logger, message.TargetUser, message.TargetGroups, serviceUrl, hubEndpoint, params, headers, datachannelTargetSelectHandler, autoReconnect)
+	// create a websocket
+	wsId := uuid.New().String()
+	wsLogger := logger.GetWebsocketLogger(wsId) // TODO: replace with actual connectionId
```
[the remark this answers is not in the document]
lgmugnier: yes.  This ConnectionId should be provided by the bastion when it connects I think.  This is an issue with the ordering the logger in that I do it before the websocket even reaches out to Bastion, so the logger will be inaccurate.

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: Yes.  The ChannelId is mostly for distinguishing multiple channels over a single websocket but the control channel has its own dedicated websocket.  This also makes it so there can only ever be one control channel associated with a given websocket at any time.  Updated comments on this line to reflect this.

kube-go-assets/bzerolib/channels/websocket/websocket.go
lgmugnier: This was a comment for the reconnect flow that would only be possible for connection service.  Aka if bastion/agent connection breaks, we can use the reconnect flow but in the future we're probably going to want a formal message maybe

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,259 @@ const (
...
 	defer w.socketLock.Unlock()
 
-	if !w.IsReady {
-		return fmt.Errorf("Websocket not ready to send yet")
+	if !w.ready {
+		w.logger.Info("websocket not ready to send yet. waiting...")
+
+		// wait until websocket is ready
+		ticker := time.NewTicker(10 * time.Millisecond)
+		for {
```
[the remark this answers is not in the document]
lgmugnier: maybe? I think it's fine for now but once we integrate with connection service and have a better reconnect flow, then for sure

kube-go-assets/bctl/daemon/plugin/kube/utils/utils.go
lgmugnier: it's still there

kube-go-assets/bzerolib/channels/agentmessage/agentmessage.go
lgmugnier: waiting to remove this

kube-go-assets/bctl/agent/datachannel/datachannel.go
```diff
@@ -1,113 +1,123 @@
...
-				}()
-			case <-ret.websocket.DoneChan:
-				// The websocket has been closed
-				ret.logger.Info("Websocket has been closed, closing datachannel")
-				cancel()
+			case <-parentTmb.Dying(): // daemon is dying
+				datachannel.tmb.Kill(errors.New("daemon killed the radio star"))
+				return
+			case <-datachannel.tmb.Dying():
+				// parentDoneChan <- tmb.Err().Error() // report death reason to parent
```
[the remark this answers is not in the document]
lgmugnier: yes but we have nothing to do with it right now.  Maybe later when we have a better control channel.  Printing it to logs for now

kube-go-assets/bctl/agent/plugin/kube/actions/stream/stream.go
```diff
@@ -155,11 +159,11 @@ func (s *StreamAction) StartStream(streamActionRequest KubeStreamActionPayload,
 				// Check the errors
 				if err == io.EOF {
 					// This means we are done
-					s.logger.Info("Received  EOF error on Watch stream")
+					s.logger.Info("Received EOF on stream")
```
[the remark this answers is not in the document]
lgmugnier: Ah yeah I was thinking that it's the stream action but you are correct, this just makes it super generic.  Since it could also be a Follow stream and this could become confusing, could I change it to "Received EOF on Stream action stream?" or something (open to alts)

kube-go-assets/bctl/agent/agent.go
```diff
@@ -45,93 +48,94 @@ func main() {
...
-	err = newAgent(logger, serviceUrl, activationToken, agentVersion, orgId, environmentId, clusterName, clusterId, idpProvider, idpOrgId, namespace)
+	err = newAgent(logger, agentVersion)
 	if err != nil {
 		logger.Error(err)
 		return
 	}
 
-	// Connect to the control channel
-	control, err := cc.NewControlChannel(ccLogger, serviceUrl, activationToken, orgId, clusterName, clusterId, environmentId, agentVersion, controlchannelTargetSelectHandler)
+	// profiling code
```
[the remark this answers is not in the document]
lgmugnier: Nope we don't 698faa3

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,243 @@ package controlchannel
...
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-const (
-	hubEndpoint       = "/api/v1/hub/kube-control"
-	challengeEndpoint = "/api/v1/kube/get-challenge"
-	autoReconnect     = true
-)
+// const (
```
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: 698faa3

kube-go-assets/bzerolib/channels/websocket/websocket.go
lgmugnier: 698faa3

kube-go-assets/bctl/agent/datachannel/datachannel.go
```diff
@@ -1,113 +1,123 @@
...
+
+const (
+	Kube PluginName = "kube"
+)
 
 type DataChannel struct {
 	websocket *ws.Websocket
 	logger    *lggr.Logger
-	ctx       context.Context
+	tmb       tomb.Tomb
```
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/agent/datachannel/datachannel.go
```diff
@@ -1,113 +1,123 @@
...
 		targetUser:   targetUser,
 		targetGroups: targetGroups,
-		logger:       logger, // TODO: get debug level from flag
-		ctx:          ctx,
+		inputChan:    make(chan am.AgentMessage, 50),
+		id:           id,
 	}
 
-	// Subscribe to our input channel
+	// register with websocket so datachannel can send a recieve messages
```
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/agent/datachannel/datachannel.go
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/agent/datachannel/datachannel.go
```diff
@@ -159,21 +173,21 @@ func (d *DataChannel) handleKeysplittingMessage(keysplittingMessage *ksmsg.Keysp
 			}
 
 			// Start plugin
-			if err := d.startPlugin(plgn.PluginName(x[0])); err != nil {
+			if err := d.startPlugin(PluginName(x[0])); err != nil {
```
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/agent/plugin/kube/kube.go
```diff
@@ -64,28 +72,20 @@ func NewPlugin(ctx context.Context, logger *lggr.Logger, ch chan smsg.StreamMess
...
-
-func (k *KubePlugin) InputMessageHandler(action string, actionPayload []byte) (string, []byte, error) {
+func (k *KubePlugin) Receive(action string, actionPayload []byte) (string, []byte, error) {
 	// Get the action so we know where to send the payload
-	msg := fmt.Sprintf("Plugin received Data message with %v action", action)
-	k.logger.Info(msg)
+	k.logger.Info(fmt.Sprintf("Plugin received Data message with %v action", action))
 
+	// parse action
 	x := strings.Split(action, "/")
```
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/agent/datachannel/datachannel.go
```diff
@@ -1,113 +1,123 @@
...
-				// Handle each message in its own thread
-				go func() {
-					ret.Receive(agentMessage)
-				}()
-			case <-ret.websocket.DoneChan:
-				// The websocket has been closed
-				ret.logger.Info("Websocket has been closed, closing datachannel")
-				cancel()
+			case <-parentTmb.Dying(): // daemon is dying
+				datachannel.tmb.Kill(errors.New("daemon killed the radio star"))
```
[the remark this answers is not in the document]
lgmugnier: 698faa3

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
lgmugnier: 698faa3

kube-go-assets/bctl/daemon/plugin/kube/actions/restapi/restapi.go
lgmugnier: 698faa3

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
lgmugnier: 698faa3

kube-go-assets/bctl/daemon/plugin/kube/actions/stream/stream.go
lgmugnier: 698faa3

kube-go-assets/bctl/daemon/plugin/kube/actions/restapi/restapi.go
lgmugnier: 698faa3

kube-go-assets/bzerolib/logger/logger.go
lgmugnier: 698faa3

kube-go-assets/bctl/daemon/plugin/kube/kube.go
lgmugnier: d1d03af

kube-go-assets/bctl/daemon/plugin/kube/utils/utils.go
lgmugnier: d1d03af

kube-go-assets/bctl/agent/datachannel/datachannel.go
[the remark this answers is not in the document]
lgmugnier: yes but we have nothing to do with it right now. Maybe later when we have a better control channel. Printing it to logs for now

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
lgmugnier: sorry must have rebased wrong

kube-go-assets/bctl/agent/plugin/kube/actions/exec/exec.go
```diff
@@ -144,9 +142,10 @@ func (e *ExecAction) InputMessageHandler(action string, actionPayload []byte) (s
 }
 
 func (e *ExecAction) validateRequestId(requestId string) error {
-	if err := kubeutils.ValidateRequestId(requestId, e.requestId); err != nil {
-		e.logger.Error(err)
-		return err
+	if requestId != e.requestId {
```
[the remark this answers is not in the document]
lgmugnier: 8c487fd

kube-go-assets/bctl/agent/plugin/kube/actions/stream/stream.go
```diff
@@ -75,14 +83,18 @@ func (s *StreamAction) InputMessageHandler(action string, actionPayload []byte)
 			s.logger.Error(rerr)
 			return action, []byte{}, rerr
 		}
-		if err := s.validateRequestId(streamActionRequest.RequestId); err != nil {
-			return "", []byte{}, err
+
+		// check requestid matches
+		if streamActionRequest.RequestId != s.requestId {
```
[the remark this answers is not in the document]
lgmugnier: 8c487fd

kube-go-assets/bctl/agent/plugin/kube/actions/stream/stream.go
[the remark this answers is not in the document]
lgmugnier: 8c487fd

kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go
lgmugnier: 8c487fd

kube-go-assets/bctl/agent/agent.go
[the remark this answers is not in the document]
sidpremkumar: for now I believe its possible to remove environment_id, so if we can we should do that now and remove this comment since we do know we need them.

kube-go-assets/bzerolib/channels/agentmessage/agentmessage.go
sidpremkumar: ? not sure I follow

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
 	w.socketLock.Lock()
 	defer w.socketLock.Unlock()
 
-	if !w.IsReady {
-		return fmt.Errorf("Websocket not ready to send yet")
+	if !w.ready {
+		w.logger.Info("websocket not ready to send yet. waiting...")
+
+		// wait until websocket is ready
+		ticker := time.NewTicker(10 * time.Millisecond)
```
sidpremkumar, line 246: Can we extract this function (or at least `10 * time.Millisecoond`) to a common util, we also share the same logic in `datachannel.go`

kube-go-assets/bzerolib/channels/websocket/websocket.go
sidpremkumar: no that makes sense, so I assume then `for ws.ready {}` if `ws.ready` was false we would block there correct?

lgmugnier: yes, exactly

kube-go-assets/bzerolib/channels/agentmessage/agentmessage.go
lgmugnier: Technically this is the right way to do this.  I want the agentmessage package to be in control of inserting its own `SchemaVersion` so none of the other packages needs to think about it.  rn I have it as the default json value, but I think that if we want to make this more "pro", we have to use this function.  But I'm a coward and am holding out until the last second in case there's another option

kube-go-assets/bzerolib/channels/websocket/websocket.go
lgmugnier: I think since it's only in those two places for now, it's better just to keep it transparent and not obscure it.  It might change too when we integrate with the connection service, so it's probably not worth putting the effort in now.

ymarcus93 (changes requested):
I like the refactors done in this PR. I like the separation between DCs+CCs and websockets which is nice.

**Open, general questions**:

1) Does this PR have it so CC is on its own separate websocket, and DCs are all managed on a single, separate websocket (by using `ChannelId` to let `websocket.go` distinguish which DC to send the incoming message to). That is, are there two websockets in total? Or is there still 1 websocket per DC with the option of allowing more than one DC to exist on a single WS if we want for certain plugins?
2) How have we tested that everything is working as before? Have we tested all actions and use-cases manually? Let me know if I should run some manual tests myself on this branch.

Also, haven't confirmed yet because a lot has been refactored, but I think my comment here: https://github.com/bastionzero/zli/pull/207#discussion_r715700978 has been addressed. I don't think we're handling each incoming message on the WS/DC on its own goroutine anymore, so we've now prevented some of the data races explained in that comment.

To elaborate,

https://github.com/bastionzero/zli/blob/0bb94f06132154dfc04992ff15df1f9e4d4cb0ff/kube-go-assets/bctl/daemon/plugin/kube/actions/exec/exec.go#L110-L121

unless there is some strange reordering going on in the SignalR Hub (have not confirmed), I'm not sure we need `streamQueue` anymore in the code block above (and any variants of it in other actions--e.g. perhaps `port.go`?). Don't remove `streamQueue` in this PR, but we should write a ticket to investigate if it is needed anymore. CC @sidpremkumar 

----
Comments that may apply to other parts of the codebase:
- `doneChan` usage. Do plugins/actions need `doneChan` if `tomb` is being used?
- Use `tomb.Go` so that `tomb` can track the goroutines and `Wait` can wait for tracked goroutines to terminate.
- Any pattern where we have a `for` loop constantly checking for a conditon to be true, even if done in a separate goroutine, should use either a channel or `sync.Cond` to be signaled when there is work to do (because the condition has changed). Good article on `sync.Cond`: https://lukechampine.com/cond.html

kube-go-assets/bctl/daemon/datachannel/datachannel.go
```diff
@@ -1,272 +1,317 @@
...
-	doneChannel chan string
+	// channels for incoming and outgoing messages
+	inputChan chan am.AgentMessage
 
 	// If we need to send a SYN, then we need a way to keep
 	// track of whatever message that triggered the send SYN
-	onDeck      plgn.ActionWrapper
-	lastMessage plgn.ActionWrapper
+	onDeck      plugin.ActionWrapper
+	lastMessage plugin.ActionWrapper
```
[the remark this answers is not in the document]
lgmugnier: You're correct.  If we get an error on a keysplitting message, then we use these to make sure we're not dropping any messages.

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,241 @@ package controlchannel
...
+		} else {
+			return c.openWebsocket(owRequest)
+		}
+	case am.CloseWebsocket:
+		var cwRequest CloseWebsocketMessage
+		if err := json.Unmarshal(agentMessage.MessagePayload, &cwRequest); err != nil {
+			return fmt.Errorf("malformed close websocket request")
+		} else {
+			if websocket, ok := c.connections[cwRequest.ConnectionId]; ok {
+				websocket.Client.Close(errors.New("websocket closed on daemon"))
```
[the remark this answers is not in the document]
lgmugnier: 71168ef

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,241 @@ package controlchannel
...
-	hubEndpoint       = "/api/v1/hub/kube-control"
-	challengeEndpoint = "/api/v1/kube/get-challenge"
-	autoReconnect     = true
-)
-
+type wsMeta struct {
+	Client       *websocket.Websocket
+	DataChannels map[string]websocket.IChannel
+}
 type ControlChannel struct {
```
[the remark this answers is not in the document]
lgmugnier: 71168ef

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
+		for {
+			<-ticker.C
+			if w.ready {
+				w.logger.Info("websocket ready")
+				break
+			}
+		}
```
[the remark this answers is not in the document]
lgmugnier: I think you're right.  I thought I had put this in here for a good reason, but I can't think of it

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
+func (w *Websocket) Close(reason error) {
+	// close all of our existing datachannels
+	for _, channel := range w.channels {
+		channel.Close(reason)
 	}
+
+	// tell our tmb to clean up after us
+	w.logger.Infof("websocket closed because: %s", reason)
+	w.tmb.Kill(reason)
+	w.tmb.Wait()
```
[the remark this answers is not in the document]
lgmugnier: 71168ef

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
-		// Listener for any messages that need to be sent
-		go func() {
-			for {
+	// Listener for any messages that need to be sent
+	go func() {
+		for {
+			for ws.ready {
 				select {
-				case <-w.ctx.Done():
+				case <-ws.tmb.Dying():
```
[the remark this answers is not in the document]
lgmugnier: 71168ef

kube-go-assets/bctl/agent/agent.go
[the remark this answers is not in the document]
lgmugnier: 86d2676

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
 }
 
 // This will be the client that we use to store our websocket connection
 type Websocket struct {
-	client  *websocket.Conn
-	logger  *lggr.Logger
-	IsReady bool
+	client     *websocket.Conn
+	logger     *logger.Logger
+	ready      bool
```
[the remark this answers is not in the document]
lgmugnier: > Does this PR have it so CC is on its own separate websocket, and DCs are all managed on a single, separate websocket (by using ChannelId to let websocket.go distinguish which DC to send the incoming message to). That is, are there two websockets in total? Or is there still 1 websocket per DC with the option of allowing more than one DC to exist on a single WS if we want for certain plugins?

Yes.  For right now we only have 2 websockets: one for the control channel (bastion <-> agent) and another for all of our datachannels (daemon <-> agent). In the future, once we've integrated with the connection service, we'll have the freedom to scale the websockets more.

> How have we tested that everything is working as before? Have we tested all actions and use-cases manually? Let me know if I should run some manual tests myself on this branch.

We are testing this manually right now, feel free to test yourself.

> unless there is some strange reordering going on in the SignalR Hub (have not confirmed), I'm not sure we need streamQueue anymore in the code block above (and any variants of it in other actions--e.g. perhaps port.go?). Don't remove streamQueue in this PR, but we should write a ticket to investigate if it is needed anymore.

I think we should keep this in for now just because I feel like there's always a risk of things being out of order and it's always better to be safe than sorry because we do have two websockets instead of one.

> Comments that may apply to other parts of the codebase:
> 
> doneChan usage. Do plugins/actions need doneChan if tomb is being used?

It's essentially being used as a local tomb and a way for these local tombs to close each other without having to kill the big daddy tomb.  I understand your point but I didn't think it really needed changing, feel free to let me know if you disagree.

> Use tomb.Go so that tomb can track the goroutines and Wait can wait for tracked goroutines to terminate.

Yup, figured this one out recently and it's been committed.

> Any pattern where we have a for loop constantly checking for a conditon to be true, even if done in a separate goroutine, should use either a channel or sync.Cond to be signaled when there is work to do (because the condition has changed). Good article on sync.Cond: https://lukechampine.com/cond.html

I think `sync.Cond` is a great idea.

> Is it fine to set it to true only in case (1)? Is it safe to assume ReadyBastionToClient is the first message the Bastion sends to the other party when the connection is ready? Or is ready supposed to indicate that the websocket connection is established (i.e. Connect returns successfully)?

In regards to below, the websocket is doing a little heavy lifting here because it's handling cases for both the daemon and the agent which do this ready dance differently.  So, for right now, it will have to stay like this to avoid unnecessary complexity in the bastion.

> Shouldn't it be retrying using exponential backoff if some error occurs in Connect? There was a PR by @sidpremkumar which added this somewhere using some standard backoff library. Will this be added here?

Sid did make a comment about this [above](https://github.com/bastionzero/zli/pull/234#discussion_r745067507), we haven't done it yet but you're correct.

> Also, haven't confirmed yet because a lot has been refactored, but I think my comment here: #207 (comment) has been addressed. I don't think we're handling each incoming message on the WS/DC on its own goroutine anymore, so we've now prevented some of the data races explained in that comment.

Actually, we still need some async-ness here because of how keysplitting works vs. streams but I've addressed your issues by pushing keysplitting to a queue which gets processed in series while still allowing the datachannel to receive streams.

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
+					w.logger.Info("Connection Ready")
+					w.ready = true
+				default:
+					w.ready = true
+
+					// push to the right channel
+					agentMessage := message.Arguments[0]
+
+					if channel, ok := w.getChannel(agentMessage.ChannelId); ok {
+						channel.Receive(agentMessage)
```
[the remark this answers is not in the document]
lgmugnier: But the websocket is not in charge of creating and destroying the datachannels, that's the responsibility of the httpserver and the control channel.  So we have a different pubsub setup than in the blog post.  One entity creates and manages, while the datachannel is responsible for subbing and unsubbing.  

> we can use a select statement on publishing a message to a channel and adding a <-ws.tmb.Dying(): so that websocket.go does not block on sending a channel if the tmb has been told to die.

I might be wrong but because we don't technically close any of the channels, websocket would never block on this, the message would just be put into a channel and never processed and eventually would be garbage collected.

Also I absolutely think the client should be the one setting the buffer size since it's the client that knows what size buffer it should have based on its own use case.

kube-go-assets/bctl/agent/plugin/kube/actions/stream/stream.go
```diff
@@ -2,29 +2,31 @@ package stream
...
+	"gopkg.in/tomb.v2"
 )
 
 type StreamAction struct {
+	logger *logger.Logger
+	tmb    *tomb.Tomb
+	closed bool
+
+	streamOutputChan chan smsg.StreamMessage
+	doneChan         chan bool
```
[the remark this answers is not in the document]
lgmugnier: These `doneChan`s are used as essentially local tombs.  I see you're point (they could be tombs) I just didn't think it was all that important but that's maybe something we'll think is important later

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,259 @@ const (
...
 
+// Function to write signalr message to websocket
+func (w *Websocket) Send(agentMessage am.AgentMessage) {
+	w.sendQueue <- agentMessage
+}
+
 func (w *Websocket) Connect() {
-	for !w.IsReady {
+	for {
+		time.Sleep(time.Second * sleepIntervalInSeconds)
```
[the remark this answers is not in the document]
lgmugnier: b23a82e

kube-go-assets/bzerolib/channels/websocket/websocket.go
lgmugnier: Removed b23a82e

kube-go-assets/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
lgmugnier: b23a82e

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
+	w.addChannel(id, channel)
+}
+
+// remove channel from channel dictionary
+func (w *Websocket) Unsubscribe(id string) {
+	w.deleteChannel(id)
+}
+
+func (w *Websocket) SubscriberCount() int {
+	return len(w.channels)
```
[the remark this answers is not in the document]
lgmugnier: b23a82e

kube-go-assets/bzerolib/channels/websocket/websocket.go
```diff
@@ -32,204 +33,257 @@ const (
...
 }
 
 // Returns error on websocket closed
 func (w *Websocket) Receive() error {
```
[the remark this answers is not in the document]
lgmugnier: b23a82e

kube-go-assets/bctl/daemon/httpserver/httpserver.go
```diff
@@ -0,0 +1,233 @@
...
+				}
+			}
+		}()
+		return datachannel, nil
+	}
+}
+
+func (h *HTTPServer) rootCallback(logger *lggr.Logger, w http.ResponseWriter, r *http.Request) {
+	h.logger.Info(fmt.Sprintf("Handling %s - %s\n", r.URL.Path, r.Method))
+
```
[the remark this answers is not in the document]
lgmugnier: b23a82e

kube-go-assets/bctl/agent/agent.go
```diff
@@ -39,95 +40,84 @@ func main() {
...
-				startDatachannel(dcLogger, message)
-			}
-		}
-	}()
+	startControlChannel(logger, agentVersion)
 
 	// Sleep forever because otherwise kube will endlessly try restarting
 	// Ref: https://stackoverflow.com/questions/36419054/go-projects-main-goroutine-sleep-forever
 	select {}
 }
```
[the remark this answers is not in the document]
lgmugnier: > Also, listening for SIGINT should also be done in daemon.go. Right now, I think we do kill -9 in zli disconnect. But instead if we changed it to kill -SIGTERM we could also give the chance for the daemon to gracefully clean up.

> Oh also, we should give daemon a grace period to clean up similar to how Kube gives a grace period before killing the agent.

Good idea but out of scope for this PR, could you write a ticket for this please @ymarcus93

kube-go-assets/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
ymarcus93: > I think sync.Cond is a great idea but I also think that it's not the easiest to understand on first glance. I understand the benefit, but until we start doing some heavy optimization I think this way makes it easier to understand which I think should be a higher priority atm. Although I did add a comment to remind us if/when the time comes in 450aec4

Fair point. If we don't want to use `sync.Cond` we could also use a channel that we just `close` to signal readiness. But the problem with this is you can only close a channel once. That's why I was asking in a different thread if we could simplify how many times `ready` flips on and off.

But given the following:

> In regards to below, the websocket is doing a little heavy lifting here because it's handling cases for both the daemon and the agent which do this ready dance differently. So, for right now, it will have to stay like this to avoid unnecessary complexity in the bastion.

It seems we can't use a channel unless we split `ready` into two different things for agent vs daemon so we can just `close` once per. We can revisit `sync.Cond` at a later point.

kube-go-assets/bctl/agent/agent.go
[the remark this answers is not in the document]
ymarcus93: Done. See: CWC-1319

kube-go-assets/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
ymarcus93: > But the websocket is not in charge of creating and destroying the datachannels, that's the responsibility of the httpserver and the control channel. So we have a different pubsub setup than in the blog post. One entity creates and manages, while the datachannel is responsible for subbing and unsubbing.

Sorry if I was unclear. I didn't mean to change the responsibility concerns. Currently, `websocket.go` exposes `Subscribe` and `Unsubscribe` which is good. I didn't mean to remove that--although we could move it to some other struct like `PubSub` that only deals with channels and the sub map if we want to keep the structs separate (and possibly reuse PubSub logic somewhere else).

> Also I absolutely think the client should be the one setting the buffer size since it's the client that knows what size buffer it should have based on its own use case.

Yeah I'm okay with this too. We could do something like this:

```
func (ps *Pubsub) Subscribe(topic string, bufferSize int) <-chan string {
  ps.mu.Lock()
  defer ps.mu.Unlock()

  ch := make(chan string, bufferSize)
  ps.subs[topic] = append(ps.subs[topic], ch)
  return ch
}
```

But that would require to also do the other refactor which would be having access to the channel type in `websocket.go` instead of calling `IChannel.Receive()`.

> I might be wrong but because we don't technically close any of the channels, websocket would never block on this, the message would just be put into a channel and never processed and eventually would be garbage collected.

In theory, the call to `Receive()` can block if the underlying channel buffer is full. Looks like it has a capacity of 25 right now, at least in the CC.

https://github.com/bastionzero/zli/blob/450aec470e2b0f2d25556cfaf892cfc3615b1494/kube-go-assets/bctl/agent/controlchannel/controlchannel.go#L64

This buffer will be full if `control.process(agentMessage)` (or `dataChannel.process()` for that matter) is taking too long to return, OR there is an influx of CC messages on the websocket _after_ `CloseWebsocketMessage` was processed by the CC. The processing of `CloseWebsocketMessage` causes the goroutine that is listening on `inputChan` to stop listening. If there are > 25 messages received after the goroutine that listens on `inputChan` closes, then `channel.Receive(agentMessage)` blocks forever meaning this goroutine never returns:

https://github.com/bastionzero/zli/blob/450aec470e2b0f2d25556cfaf892cfc3615b1494/kube-go-assets/bzerolib/channels/websocket/websocket.go#L93-L104

and as a side-effect, this one doesn't either

https://github.com/bastionzero/zli/blob/450aec470e2b0f2d25556cfaf892cfc3615b1494/kube-go-assets/bctl/agent/controlchannel/controlchannel.go#L202-L205

That's why I wanted to have a `select` statement so the `websocket` can stop waiting to send on `Receive()`/`inputCh` if the websocket tomb itself has been told to die.

kube-go-assets/bctl/daemon/datachannel/datachannel.go
```diff
@@ -89,18 +94,30 @@ func New(logger *logger.Logger,
...
+			}
+		}
+	})
+
+	// listen and process keysplitting messages in series
+	go func() {
+		for {
+			ksMessage := <-dc.ksInputChan
+			if err := dc.handleKeysplitting(ksMessage); err != nil {
+				dc.logger.Error(err)
```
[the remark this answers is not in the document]
ymarcus93: Also, isn't this goroutine untracked by tomb?

kube-go-assets/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
ymarcus93: But I guess actually for the latter case, the influx of messages won't be processed as this goroutine

https://github.com/bastionzero/zli/blob/450aec470e2b0f2d25556cfaf892cfc3615b1494/kube-go-assets/bzerolib/channels/websocket/websocket.go#L93-L104

won't call `receive()` if the tomb is dying. So maybe we're actually okay.

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,247 @@ package controlchannel
...
-	// Make and add our params
-	params := map[string]string{
-		"public_key":    config.Data.PublicKey,
-		"agent_version": agentVersion,
+	// Set up our handler to deal with incoming messages
+	control.tmb.Go(func() error {
+		defer websocket.Unsubscribe(id)
+		for {
+			select {
+			// case <-parentTmb.Dying():
```
sidpremkumar, line 79: This should be uncommented?

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,247 @@ package controlchannel
...
+	// Push message to websocket channel output
+	c.websocket.Send(agentMessage)
+	return nil
+}
+
+func (c *ControlChannel) openWebsocket(message OpenWebsocketMessage) error {
+	subLogger := c.logger.GetWebsocketLogger(message.ConnectionId)
+
+	// Create our headers and params, headers are empty
+	// TODO: We need to drop this session id auth header req and move to a token based system
```
sidpremkumar, line 123: Not sure I understand this comment?

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
```diff
@@ -3,142 +3,247 @@ package controlchannel
...
+	return nil
 }
 
-func (c *ControlChannel) Receive(agentMessage wsmsg.AgentMessage) error {
-	switch wsmsg.MessageType(agentMessage.MessageType) {
-	case wsmsg.NewDatachannel:
-		var dataMessage NewDatachannelMessage
-		if err := json.Unmarshal(agentMessage.MessagePayload, &dataMessage); err != nil {
-			return fmt.Errorf("error unmarshalling new controlchannel request: %v", err.Error())
+func (c *ControlChannel) process(agentMessage am.AgentMessage) error {
```
sidpremkumar, line 171: Can we add a comment here giving some context to this function? What's its purpose what's it doing?

Only asking since this is a decently important function and I think that'll clarify confusion later down the line

kube-go-assets/bctl/agent/plugin/kube/actions/exec/exec.go
```diff
@@ -30,53 +30,53 @@ const (
...
 	}, nil
 }
 
 func (e *ExecAction) Closed() bool {
 	return e.closed
 }
 
-func (e *ExecAction) InputMessageHandler(action string, actionPayload []byte) (string, []byte, error) {
+func (e *ExecAction) Receive(action string, actionPayload []byte) (string, []byte, error) {
+	// TODO: Check request ID matches from startexec
```
sidpremkumar, line 79: We have helper functions for this, and I believe we are already doing this on line 103?

sidpremkumar:
lgtm, just need to do a in depth test tomorrow

sidpremkumar:
Logs + exec are both not working: 
2021-11-23 at 9 45 AM

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: controlchannel never receives a `parentTmb` because we don't currently have any logic in the agent code that would require that.  This is here mostly as a place holder, I can remove it if it bothers you

sidpremkumar: Yeah lets remove it if we are not using it, Or lets leave a comment here explaining why it might be useful? Just thinking it might give the next person who takes a look more context.

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: change name `process` -> `processInput`, can you specify exactly what is unclear/what kind of context you needed to figure out that we can put in there?

kube-go-assets/bctl/agent/plugin/kube/actions/exec/exec.go
lgmugnier: this was a leftover old comment, deleted.

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
sidpremkumar: If you don't think its necessary feel free to ignore this comment !

kube-go-assets/bctl/daemon/datachannel/datachannel.go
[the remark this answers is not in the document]
lgmugnier: oh good catch

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: 42cfdae

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: 42cfdae

kube-go-assets/bctl/agent/controlchannel/controlchannel.go
lgmugnier: 42cfdae

kube-go-assets/bctl/daemon/datachannel/datachannel.go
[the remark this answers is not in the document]
lgmugnier: 42cfdae

kube-go-assets/bctl/agent/plugin/kube/actions/exec/exec.go
lgmugnier: 42cfdae

lgmugnier:
> Logs + exec are both not working: 2021-11-23 at 9 45 AM

exec seems to be working, looks like stream is having difficulties...debugging

sidpremkumar:
Also make sure to update the `VERSION` file for the agent to be 3.x.x. This should be a major version bump. (not sure if you did I couldn't find it but GitHub is hard to scroll with so many files 😓 )

sidpremkumar (approved):
Tested with Lens + k9s and was not able to break anything Nice! 🚢 Tested groups as well. 

Just make sure to take care of this https://github.com/bastionzero/zli/pull/234#issuecomment-978060548

kube-go-assets/bzerolib/channels/websocket/websocket.go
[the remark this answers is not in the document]
lgmugnier: So I think what you're saying is that there is a problem if the datachannel's channels are full, we'll lose the messages if the websocket keeps sending stuff anyways.  I don't disagree and this is a known problem, but I don't understand the solution you're proposing.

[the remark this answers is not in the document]
ymarcus93: yeah actually i think it's fine. its just normal blocking/throttling until old messages get processed.

ymarcus93 (approved):

PR #242, opened 2021-11-01
Add dynamic buffer time for stdin

sidpremkumar:
## Description of the change
I added a dynamic buffer time, this means if the buffer sees a lot of stdin coming very quickly, it will increase the buffer time to allow for more input to be sent (i.e. in the event of a copy and paste). Taking a look at the latency we get ~51.7ms:  
2021-11-02 at 9 47 AM
(connection id: 42e3cbd4-8d26-4f22-b7b0-3668e919a350)

Comparing that to no buffer at all, and stdin -> terminal input we get an average of ~49.1: 
2021-11-02 at 9 48 AM
(connection Id: 838a23c2-2b01-4acc-8394-2aeb080b4038)

So this buffer is not really impacting latency a lot, and significantly less than before.

Now moving on to the copy-paste, this is much much faster than what I had previously. What I had previously worked, and did the job for small - medium text sizes. But feel short on super (personally unrealistic) large text sizes. I confirmed this by using ssh into an ec2 instance and our copy and paste and found that the ssh took ~5mins vs we took ~20mins. 

I tested again with this updated code and we are now much closer to ssh. 

This is a comparison with a 10k chunk size: 
https://drive.google.com/file/d/1phkV3rWQP1EILJcQNXewjoGFsxUuTxu9/view?usp=sharing

I also tried 15k chunk size resulted in 503's: 
https://drive.google.com/file/d/1puT8L0EBfN2LMLQkjBPkyC5steVdN9RQ/view?usp=sharing

Increasing the chunk size to 25k resulted in this as well: 
2021-11-02 at 9 58 AM

**~~~~~~~~~~~~OUTDATED~~~~~~~~~~~~** 

Changes the buffer time to 1ms. Tested and this does decrease the latency: 
2021-11-01 at 1 03 PM

And this is a test where I just removed the buffer time completely: 
2021-11-01 at 1 05 PM

And its very comparable (~2ms diff). 

Also from a user perspective this is what it looks like: 
https://drive.google.com/file/d/1ZCddmxvVfT85WWOoibschmU_BlV6RAwb/view?usp=sharing

We can see from the logs that the program is batching the string, which I believe is okay and is expected.

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1233

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


sidpremkumar:
Also for fun tested copying/pasting this: https://facemessing.blogspot.com/2021/04/super-long-text-to-copy-and-paste.html


And it worked: 
2021-11-01 at 1 13 PM
 (checked the start / end sentence and they matched up)

lipmas (approved):

sidpremkumar:
Another interesting find is copying huge amounts of data results in a buffering, but this is similar to what ssh does (ssh on the right, and zli on the left):
https://drive.google.com/file/d/1c5xWnZlsyMNavbcM4ToDeS6OesA3Su30/view?usp=sharing

sidpremkumar:
I also tested this on a digital ocean droplet: 
2021-11-01 at 5 17 PM
and was able to copy the largest file I was able to create (any larger and my computer would freeze up) 

I used the following command to create that file
```
{repeat 238609294 echo some test} > large-file
```

But ended that command prematurly to create the 12mb text file to test.

src/shell-utils.ts
```diff
@@ -53,27 +52,71 @@ export async function createAndRunShell(
...
+            previousInput = Date.now();
+
+            // If we get a new input, clear the timeout function
+            if (bufferFunction === null) {
+                // send the input to a function after a certain amount of time has passed
+                bufferFunction = setTimeout(() => {
+                    // Loop over the array, and send it as chunks of 10000
+                    // Otherwise we get keysplitting/general errors if we try to send too much data
+                    const chunk = 10000;
+                    for (let i = 0,j = inputBuffer.length; i < j; i += chunk) {
```
tflpd, line 107: small simplification
```suggestion
                    for (let i = 0; i < inputBuffer.length; i += chunk) {
```

src/shell-utils.ts
```diff
@@ -53,27 +52,71 @@ export async function createAndRunShell(
...
+        process.stdin.on('keypress', async (_, key) => {
+            // Implement some custom logic for batching input
+            // Ref: https://stackoverflow.com/questions/66755705/detect-pasted-input-with-readline-nodejs
+
+            // Add our input to our array of input
+            inputBuffer.push(key.sequence);
+
+            // Keep increasing maxInputDelay if the last input was less than 5ms ago
+            // We cap this wait at maxInputDelayLimitms
+            if ((Date.now() - previousInput) < 5) {
```
tflpd, line 87: nit; let's store this `5` in a var instead of hardcoding it

src/shell-utils.ts
```diff
@@ -53,27 +52,71 @@ export async function createAndRunShell(
...
+                // Only increase if our maxInputDelay < maxInputDelayLimit
+                if (maxInputDelay < maxInputDelayLimit) {
+                    maxInputDelay += 5;
```
tflpd, line 90: ^

sidpremkumar: https://github.com/bastionzero/zli/pull/242/commits/e29980c2b38ddd3c49cdbd8639857fa2aed36d55

src/shell-utils.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/242/commits/e29980c2b38ddd3c49cdbd8639857fa2aed36d55

src/shell-utils.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/242/commits/e29980c2b38ddd3c49cdbd8639857fa2aed36d55

lipmas (approved):

PR #251, opened 2021-11-09
Support dynamic buffer for stdin for exec

sidpremkumar:
## Description of the change

We had an issue internally where rsyncing using kubectl exec would take a very long time. This PR adds a dynamic buffer size to our stdin stream, so we will increase the buffer size if we see lots of data coming in. 

I also has a discussion with @ymarcus93 and @lgmugnier regarding the speed and we have come to the conclusion that the speed here might be coming from the keysplitting overhead (specifically that we have to wait for an ACK). We thought that profiling this (i.e. metrics pt2) would not be a good use of time currently, and believe that the parallelization work we have scooped out would also help greatly here. 

I decided to go with a chunk size of 8192 as this is guaranteed to be supported as defined by the SPDY protocol: https://www.chromium.org/spdy/spdy-protocol/spdy-protocol-draft3-1 (under section 2.2.1)

Furthermore both `kubectl` and our `agent` are both sharing the same package/lib to create the SPDY stream: 
* `kubectl`: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/exec/exec.go#L119
* `agent`: https://github.com/bastionzero/zli/blob/fix/support-large-exec-sizes/kube-go-assets/bctl/agent/plugin/kube/actions/exec/exec.go#L194

And since I had to break up the bytes into 8192 in order to get the agents exec to accept the rsync, I believe that its safe to assume kubectl also is sending data in 8192 chunks. I confirmed this by running all the tests beflow with an increased sendspace on my machine (`sudo sysctl -a net.local.stream.sendspace=16384`), and everything worked as expected. This needs to be tested on other people's machines as well, however. 

This is what the normal eks rsync for webapp looks like: 
2021-11-10 at 2 03 PM

You can see about ~1,000,000 bytes/sec

This is what zli `master` looks like right now:
2021-11-10 at 2 22 PM
~150,000 bytes a second

And this is what ours is without a solution for buffered chunk size (using code from this commit: https://github.com/bastionzero/zli/pull/251/commits/feb9ee9833d6665efb8f97c843b8ef8ca53ac404):
(I tested all of this even while increasing my stream buffer size: `sudo sysctl -a net.local.stream.sendspace=16384`)

32mb max buffer: 
2021-11-10 at 2 01 PM
You can still see very slow ~200,000 bytes/sec

64mb max buffer:
2021-11-10 at 2 07 PM
Getting a bit faster ~300,000 bytes/sec

128mb max buffer:
2021-11-10 at 2 09 PM
Actually slowed things down back to ~200,0000

**outdated**

Here is a video of it in action, some context @tflpd timed it this morning without the change and it took him ~5.5min: 
https://drive.google.com/file/d/1PTGMj7dp65FX1PSmRR8VBrcyGEydTDk2/view?usp=sharing

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1174

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


lgmugnier (approved):
great work

PR #252, opened 2021-11-11
Add system test codebuild spec and github action ci

lipmas:
## Description of the change

Run system test wrapper script that downloads zli config from s3 and run system tests with jest multiple times for different idps

Dependent on changes in https://github.com/bastionzero/cwc-infra/pull/297

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1242

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


src/system-tests/system-test-codebuild.yaml
```diff
@@ -0,0 +1,53 @@
...
+          git remote add origin "$remote_origin"                # Add the remote origin so we can fetch
+          git fetch --quiet                                     # Get all the things
+          git config --global advice.detachedHead false         # turn the detached message off
+          git checkout -f "$CODEBUILD_RESOLVED_SOURCE_VERSION"  # Checkout the specific commit we are building
+        fi
+      # Now that setup is complete, get submodules
+      - git submodule init
+      - git submodule update --recursive
+      # Install Bzero-QA
+      - git clone git@github.com:bastionzero/cwc-infra.git -b feat/sys-tests /cwc-infra/
```
lipmas, line 41: remove `-b feat/sys-tests` this is only needed until https://github.com/bastionzero/cwc-infra/pull/297 is merged into master of cwc-infra

ymarcus93: This private clone works because of `build_ssh_key: "/cwc/ssh/rolandfrank"`?

ymarcus93 (commented):
I think we need to change `buildspec.yml`. `npm run test` is run after checking if version is incremented.

https://github.com/bastionzero/zli/blob/master/buildspec.yml

We don't want that CI job to run system tests. Probably need to move all unit tests to separate folder so we can use `jest` filter on folder

src/system-tests/system-test-codebuild.yaml
lipmas: yup this is using the role account ssh key for cloning. Even if we didnt need to clone cwc-infra the ssh key is still needed to update submodules in this repo (webshell-common-ts) because those arent automatically fetched by codepipeline - https://github.com/bastionzero/zli/pull/252/files#diff-4ba236c11d1e24c2b6534985a0f8ca6e9ab7403d010173fe0be3654c3446a9b3R18-R19

ymarcus93:
> I think we need to change `buildspec.yml`. `npm run test` is run after checking if version is incremented.
> 
> https://github.com/bastionzero/zli/blob/master/buildspec.yml
> 
> We don't want that CI job to run system tests. Probably need to move all unit tests to separate folder so we can use `jest` filter on folder

Fixed in c328a2bb81208518fa4cf518b2ceeb40634d8248 which will eventually bubble up to this branch

src/system-tests/system-test-codebuild.yaml
ymarcus93: Done in 864daa6

PR #255, opened 2021-11-12
Fix zli disconnect on m1 cpus

sidpremkumar:
## Description of the change

Zli disconnect was not working on M1 Macs since the `kill -9` command was not working. This PR updates that code to check against the CPUS (since the arch will return the same since we are using node 14 and it is using rosetta), and then uses different kill commands based on the CPU. 

Tested this on my m1 Mac, but need help testing on other Macs. 

@ymarcus93 and @tflpd can you check on your Macs, and @chad-cwc can you please test on your linux system as well. 

Thanks

## Type of change

## Related JIRA tickets

Relates to JIRA: CWC-1265

Release Notes:

## Potential Security Impacts

Does it fix a security issue?
Does it add attack surface?

### Development


src/services/kube/kube.service.ts
```diff
@@ -68,7 +69,15 @@ export async function killDaemon(configService: ConfigService) {
         } else if (process.platform === 'linux') {
             spawn('pkill', ['-s', kubeConfig['localPid'].toString()]);
         } else {
-            spawn('kill', ['-9', kubeConfig['localPid'].toString()]);
+            // Determine if we are on a m1 mac
+            // Ref: https://stackoverflow.com/questions/65146751/detecting-apple-silicon-mac-in-javascript
+            const osCpus = os.cpus();
+            const isM1 = osCpus[0].model.includes('Apple M1');
```
tflpd, line 75: I know, dumb question, but are we 100% this array element is never gonna be null? Otherwise should check for size before accessing its fields (yes, again, I hear you asking but how can we have a pc without a CPU but I gotta bring it up)

sidpremkumar: We can add a check for length if you want, but yes it would imply the device we are running has no cpus, and the documentation does not mention any possibility of returning a null value: https://nodejs.org/api/os.html#oscpus

tflpd:
Also tested the change, it did not kill the process.

src/services/kube/kube.service.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/255/commits/9495ef0b8ab35db6f95f3478dd3d58027c8c5fb7

ymarcus93:
Tested on my Intel Mac. Ran `zli tunnel` then `zli disconnect`. Daemon successfully killed.

ymarcus93 (approved):

chad-cwc:
Tested on Ubuntu 20.04 via WSL. Looks like it worked.

PR #258, opened 2021-11-18
Ensure we await the call to kill the daemon

sidpremkumar:
## Description of the change

Noticed an issue where the daemon was only **sometimes** getting killed. I believe this was due to us not awaiting the exec that we spawned. I've turned the spawn into execs that will run synchronously. Tested on my M1 Mac: 
2021-11-18 at 6 05 PM

## Relevant release note information

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1264

## Have you considered the security impacts?

Does this PR have any security impact?

- [ ] Yes
- [x] No

If yes, please explain:

PR #259, opened 2021-11-19
Catch unknown pid in killDaemon

sidpremkumar:
## Description of the change

@tflpd noticed: 
```
➜  bin git:(master) ./zli-macos tunnel cluster-admin@dev-cluster
Logged in as: thanos@commonwealthcrypto.com, bzero-id:9b2b2284-4b66-42f3-989e-496d522595cd, session-id:032a4450-28a1-4717-8f30-59b356722b49
/bin/sh: line 0: kill: (8134) - No such process
(node:28535) UnhandledPromiseRejectionWarning: Error: Command failed: kill -9 8134
/bin/sh: line 0: kill: (8134) - No such process

    at checkExecSyncError (child_process.js:760:11)
    at Object.execSync (child_process.js:833:15)
    at execSync (pkg/prelude/bootstrap.js:1979:30)
    at /snapshot/zli/dist/src/services/kube/kube.service.js:67:21
    at Generator.next (<anonymous>)
    at /snapshot/zli/dist/src/services/kube/kube.service.js:8:71
    at new Promise (<anonymous>)
    at /snapshot/zli/dist/src/services/kube/kube.service.js:4:12
    at killDaemon (/snapshot/zli/dist/src/services/kube/kube.service.js:44:12)
    at /snapshot/zli/dist/src/handlers/tunnel/tunnel.handler.js:53:43
(Use `zli-macos --trace-warnings ...` to show where the warning was created)
(node:28535) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:28535) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Started kube daemon at localhost:51435 for cluster-admin@dev-cluster
```

Which is caused by this PR: https://github.com/bastionzero/zli/pull/258 where we are actually awaiting the kill call. 

This PR catches any errors in the killDaemon call and continues. 

Tested by throwing a test error inside the kill function:
2021-11-19 at 12 09 PM

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [ ] No

If yes, please explain:


src/services/kube/kube.service.ts
```diff
@@ -65,26 +65,30 @@ export async function killDaemon(configService: ConfigService) {
...
+    
+                const isM1 = osCpus[0].model.includes('Apple M1');
+                if (isM1) {
+                    exec(`pkill -TERM -P ${kubeConfig['localPid'].toString()}`);
+                } else {
+                    exec(`kill -9 ${kubeConfig['localPid'].toString()}`);
+                }
             }
+        } catch {
+            // If the daemon pid was killed, or doesn't exist, just continue
```
sidpremkumar, line 89: @tflpd I wasn't 100% sure what we could do here, we don't have access to a logger, but we could print something to the user if we wanted too (i.e. pass a logger here)

tflpd: For the specific use case we try to solve we do not really need to print something but this may end up catching other exceptions we are not aware of as well. I would add a `Attempt to kill existing daemon failed. This is expected if the daemon has been killed already` (?)

tflpd (approved):

src/services/kube/kube.service.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/259/commits/0f70fcccccb3cc83b8159c4f3b7de96a86ebbfa3

PR #262, opened 2021-11-24
Kube system test part 1: REST API plugin

ymarcus93:
## Description of the change

Adds Kubernetes test suite to system tests in order to test the kube feature of our product end-to-end. System tests now spin up a cluster on-demand, install the agent via `helm` quickstart job, and cleanup the cluster after running the system tests.

Currently, there is only one test in the `kube` suite. It tests the REST API plugin by issuing a `get namespaces` Kube API call using the `bctl-daemon` and a `kubectl` client in Typescript.

Paths tested:
- `helm install` - which inherently tests registration of the kube cluster + installation of agent on cluster
- `generate kubeYaml`
- `zli tunnel`
- REST API `daemon` plugin by issuing a `GET /api/v1/namespaces` (ref: https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/namespace-v1/#list-list-or-watch-objects-of-kind-namespace)

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1321

- [ ] Yes
- [x] No

If yes, please explain:


ymarcus93:
System tests passed in 31 minutes: https://github.com/bastionzero/zli/runs/4359140438?check_suite_focus=true

lipmas (approved):

src/system-tests/tests/system-test.ts
```diff
@@ -120,38 +241,38 @@ beforeAll(async () => {
...
+async function cleanupDOTestTargets() {
+    const allTargetsCleanup = Promise.allSettled(Array.from(testTargets.values()).map((doTarget) => {
+        return doService.deleteDigitalOceanSSMTarget(doTarget);
+    }));
+
+    await checkAllSettledPromise(allTargetsCleanup);
+}
+
+async function setupSystemTestApiKey() {
+    const apiKeyName = 'system-test-cluster-key';
```
sidpremkumar, line 265: Noticing we don't have any Ids for system tests runs. No action on this PR, but something maybe might be helpful is an ID associated with the system test that is used throughout the test, so for example we could add it here in the API key, here: https://github.com/bastionzero/zli/pull/262/files#diff-0ae6903ec28d9ba4741bb0d2bff81832ff5b75b90e75ea6e35e2283fd418243eR94 in the cluster name, etc

src/system-tests/tests/utils/helm/helm-utils.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const exec = promisify(childProcess.exec);
+
+/**
+ * A wrapper around the built-in exec that will throw an exception if anything
+ * is printed to stderr
+ * @param command command to execute
+ */
+export const runCommand = async (command: string) => {
+    const { stdout, stderr } = await exec(command);
```
sidpremkumar, line 14: Does this capture an exit code of != 0 if there is no stderr output, or is that not a thing?

sidpremkumar (approved):
LGTM, just minor comments I left, feel free to resolve and just ignore them if you don't think they are relevant.

src/system-tests/tests/utils/helm/helm-utils.ts
lipmas: Error exit codes are handled by nodejs `promisfy` actually and will result in a rejected promise. From [ref](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)

> If this method is invoked as its util.promisify()ed version, it returns a Promise for an Object with stdout and stderr properties. The returned ChildProcess instance is attached to the Promise as a child property. In case of an error (including any error resulting in an exit code other than 0), a rejected promise is returned, with the same error object given in the callback, but with two additional properties stdout and stderr.

src/system-tests/digital-ocean/digital-ocean-kube-service.ts
```diff
@@ -0,0 +1,224 @@
+import { Retrier } from '@jsier/retrier';
+import { DigitalOcean } from 'digitalocean-js';
+import { ConfigService } from '../../services/config/config.service';
+import { Logger } from '../../services/logger/logger.service';
+import axios from 'axios';
+import { KubeService } from '../../services/kube/kube.service';
+import { KubernetesCluster, KubernetesWorkerNodePool } from 'digitalocean-js/build/main/lib/models/kubernetes-cluster';
```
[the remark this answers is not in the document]
ymarcus93: like we do in python--third-party vs. internal libraries? yeah i don't think we have a consensus on it yet, but it would be nice if our `npm lint` did this for us.

src/system-tests/digital-ocean/digital-ocean-kube-service.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * Polls DigitalOcean's GET Kubernetes cluster API until it says the
+     * provided cluster has status == "running".
+     * @param cluster Cluster to query
+     * @returns Cluster information after its status == "running"
+     */
+    public async pollClusterRunning(cluster: KubernetesCluster): Promise<KubernetesCluster> {
+        // Try 60 times with a delay of 10 seconds between each attempt.
+        // Average ETA: 5-10 minutes to provision
```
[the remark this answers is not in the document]
ymarcus93: I think I wrote down the wrong average ETA (especially since ive never seen 5 minutes). i think i've only ever seen DO take 9 minutes. I can increase from max 10 mins to 12. Hopefully DO does not have bad day.

src/system-tests/tests/system-test.ts
ymarcus93: I think having a system test ID is a good idea. Adding it to cluster name makes sense as well. I think a future refactor related to system-test IDs could be to include the ID as an extra header in all API requests sent by the http client used in `zli`. And then in grafana have a way to parse all logs involving a specific header. It would make it easier to track things in logs related to system tests.

But right now we do have a unique ID that is part of the cluster name. that's what `randomAlphaNumericString(15).toLowerCase()` is. I guess we could also use that in the apiKeyName as well.

src/system-tests/digital-ocean/digital-ocean-kube-service.ts
```diff
@@ -0,0 +1,224 @@
...
+     */
+    public async pollClusterTargetOnline(clusterTargetName: string): Promise<ClusterSummary> {
+        // Try 30 times with a delay of 10 seconds between each attempt.
+        const retrier = new Retrier({
+            limit: 30,
+            delay: 1000 * 10,
+            stopRetryingIf: (reason: any) => reason instanceof ClusterTargetStatusPollError && reason.clusterSummary.status === KubeClusterStatus.Error
+        });
+
+        // We don't know Cluster target ID initially
```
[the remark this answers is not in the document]
ymarcus93: I guess it helps a little bit. But still I think the querying of the cluster to find the target ID would have to be behind a retrier as there is a delay between the cluster being online and accessible through kubectl, and the helm job creating the agent deployment with the env var of the cluster target ID.

And obviously we would have to use the DigitalOcean kubeConfig yaml, but we have access to that before this method is called anyways.

src/system-tests/tests/system-test.ts
lipmas: > But right now we do have a unique ID that is part of the cluster name (could be interpreted as a system test id). that's what randomAlphaNumericString(15).toLowerCase() is. I guess we could also use that in the apiKeyName as well.

Right we could use that in the apiKeyName as well as in the DO ssm targets names. Basically we could just have a global random id that is generated for system tests and reuse that anytime we construct a unique name.

src/system-tests/digital-ocean/digital-ocean-kube-service.ts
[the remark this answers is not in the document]
lipmas: d8bae2f - updated to 15 min timeout to be super safe

src/system-tests/tests/system-test.ts
lipmas: d8bae2f - reuse the same id across cluster/ssm/api key names now

src/system-tests/digital-ocean/digital-ocean-kube-service.ts
```diff
@@ -0,0 +1,224 @@
...
+            await this.policyService.DeletePolicy(kubePolicy.id);
+        } else {
+            throw new Error(`Unexpected error! Expected to find at least one policy with name: ${registeredCluster.bzeroClusterTargetSummary.clusterName}-policy`);
+        }
+    }
+
+    private async deleteClusterEnv(registeredCluster: RegisteredDigitalOceanKubernetesCluster): Promise<void> {
+        // Find the env that Helm creates and delete it
+        const envs = await this.envService.ListEnvironments();
+        const kubeEnv = envs.find(e => e.name === `${registeredCluster.bzeroClusterTargetSummary.clusterName}-Env`);
```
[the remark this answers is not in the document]
lipmas: d8bae2f

PR #263, opened 2021-11-30
Manually create cert/key for Daemon instead of using lib

sidpremkumar:
## Description of the change
We had an error where is a user was running openssl 3.x they would get an error:
2021-11-30 at 3 06 PM

Which is further documented in a PR to the upstream dependency that we use: https://github.com/Dexus/pem/pull/308

Really we can just manually create the openssl cert/key, which is the current state of the PR. Tested this by manually deleting my cert info and nulling the kubeConfig section of my zli config and generating everything again:
2021-11-30 at 3 35 PM

2021-11-30 at 3 35 PM

All of this was done with openssl 3.x
2021-11-30 at 3 39 PM

Alternative to this PR is something like this: 
```js
        // First verify they are not running openssl 3.x
        try {
            const opensslVersion = await execPromise('openssl version')
            if (opensslVersion.stdout.includes('3.')) {
                throw new Error('Openssl version 3.x is not supported currently, please downgrade to 2.x or 1.x. Openssl is needed in order to build a local key/cert for the daemon.');
            }
        } catch (e: any) {
            logger.error(`${e}`);
            await cleanExit(1, logger);
        }
```
Which I originally created, open to doing it that way (so we get the support of the `pem` lib) but also if we want to go down this route we can as well. 

This PR also ensures that we create different key/cert for dev/stage/prod by adding the `-{env}` to every file we create. 

Going to need help testing on linux since I only tested on macOS

Release Notes: Fixes issue creating key/cert for daemon for users with openssl 3.x

## Related JIRA tickets

Relates to JIRA: CWC-1276

- [ ] Yes
- [X] No

If yes, please explain:


src/handlers/generate-kube/generate-kubeconfig.handler.ts
```diff
@@ -26,56 +25,58 @@ export async function generateKubeconfigHandler(
...
-                };
-                configService.setKubeConfig(kubeConfig);
-                resolve();
-            });
+            // Get the path of where we want to save
+            const pathToConfig = path.dirname(configService.configPath());
+            const appName = configService.getConfigName();
+            const pathToKey = `${pathToConfig}/kubeKey-${appName}.pem`;
+            const pathToCsr = `${pathToConfig}/kubeCsr-${appName}.pem`;
+            const pathToCert = `${pathToConfig}/kubeCert-${appName}.pem`;
```
lipmas, line 33: nit: use `path.join` when concating paths

src/handlers/generate-kube/generate-kubeconfig.handler.ts
```diff
@@ -26,56 +25,58 @@ export async function generateKubeconfigHandler(
...
-                    targetGroups: null,
-                    targetCluster: null,
-                    defaultTargetGroups: null
-                };
-                configService.setKubeConfig(kubeConfig);
-                resolve();
-            });
+            // Get the path of where we want to save
+            const pathToConfig = path.dirname(configService.configPath());
+            const appName = configService.getConfigName();
```
lipmas, line 30: nit: configName

src/handlers/generate-kube/generate-kubeconfig.handler.ts
```diff
@@ -26,56 +25,58 @@ export async function generateKubeconfigHandler(
...
+            const localPort = await findPort();
+
+            // Now save the path in the configService
+            kubeConfig = {
+                keyPath: pathToKey,
+                certPath: pathToCert,
+                csrPath: pathToCsr,
+                token: token,
+                localHost: 'localhost',
+                localPort: await localPort,
```
lipmas, line 71: nit: i dont think this needs to be await'd again since you already await on line 62

src/handlers/generate-kube/generate-kubeconfig.handler.ts
```diff
@@ -26,56 +25,58 @@ export async function generateKubeconfigHandler(
...
+
+            // Generate a new key
+            try {
+                await execPromise(`openssl genrsa -out ${pathToKey}`);
+            } catch (e: any) {
+                reject(e);
+            }
+
+            // Generate a new csr
+            const randtoken = require('rand-token');
```
lipmas, line 43: nit: move require top of file

lipmas (approved):

sidpremkumar:
Responded to sebbys comments, tested again by nulling our `key` in the config and letting it regenerate: 
2021-12-02 at 9 31 AM

src/handlers/generate-kube/generate-kubeconfig.handler.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/263/commits/d29fdee9da032611be386acaddb2df8c2f2acd39

src/handlers/generate-kube/generate-kubeconfig.handler.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/263/commits/d29fdee9da032611be386acaddb2df8c2f2acd39

src/handlers/generate-kube/generate-kubeconfig.handler.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/263/commits/d29fdee9da032611be386acaddb2df8c2f2acd39

src/handlers/generate-kube/generate-kubeconfig.handler.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/263/commits/d29fdee9da032611be386acaddb2df8c2f2acd39

lipmas (approved):

PR #265, opened 2021-12-06
Add is-ready endpoint to daemon and poll this in tunnel

ymarcus93:
## Description of the change

Polls for daemon to be ready before exiting `zli tunnel`. This is helpful for customer UX and system-tests.

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [x] No

If yes, please explain:


src/handlers/tunnel/tunnel.handler.ts
```diff
@@ -115,8 +117,11 @@ export async function startKubeDaemonHandler(argv: yargs.Arguments<tunnelArgs>,
             kubeConfig['targetCluster'] = targetCluster;
             configService.setKubeConfig(kubeConfig);
 
+            // Wait for daemon HTTP server to be bound and running
             await waitUntilUsedOnHost(parseInt(daemonPort), 'localhost', 100, 1000 * 5);
```
lipmas, line 121: we probably no longer need to wait for the port to be in use since this will be covered by the pollDaemonReady function

lipmas: Actually maybe we should keep this in case the daemon http server never comes up which should happen much quicker than the daemon being ready. Otherwise we will have a 2 minute timeout on the http server coming up.

src/handlers/tunnel/tunnel.handler.ts
```diff
@@ -115,8 +117,11 @@ export async function startKubeDaemonHandler(argv: yargs.Arguments<tunnelArgs>,
...
 
+            // Poll ready endpoint
+            await pollDaemonReady(kubeConfig['localPort']);
```
lipmas, line 124: and then we can add an info log before we start polling. "Waiting for daemon to become ready"

kube-go-assets/bctl/daemon/httpserver/httpserver.go
```diff
@@ -191,6 +191,17 @@ func (h *HTTPServer) bubbleUpError(w http.ResponseWriter, msg string, statusCode
 func (h *HTTPServer) rootCallback(logger *logger.Logger, w http.ResponseWriter, r *http.Request) {
 	h.logger.Infof("Handling %s - %s\n", r.URL.Path, r.Method)
 
+	// Special case /bzero-is-ready
+	if strings.HasSuffix(r.URL.Path, "/bzero-is-ready") {
+		if h.restApiDatachannel.Ready() {
+			w.Write([]byte("ready"))
```
lgmugnier, line 197: if you're just waiting for a status okay, wouldn't it be more explicit to `w.WriteHeader(http.StatusOK)` to follow the pattern below as well?

ymarcus93: 893de35

src/handlers/tunnel/tunnel.handler.ts
ymarcus93: 7c0d1031e04b4d8e66ed6a97706fc1cbe1040baf

lgmugnier (approved):

lipmas (approved):

PR #266, opened 2021-12-07
Add check to see if port is still in use for kube daemon

sidpremkumar:
## Description of the change

I am still having a touch time killing the kube daemon on my M1 machine. Sometimes it seems to work fine, and other times it does not seem to kill the daemon. This is also a problem while developing, where I have to manually find the PID of the process in that port using `> lsof ...`. 

This PR adds more robustness to our kill and tunnel, by checking if any process is using that port first. 

Tested:
2021-12-07 at 11 06 AM

Could people test on WSL, intel Macs, and Linux to make sure this all still is working for them. 

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1351

- [ ] Yes
- [X] No

If yes, please explain:


tflpd (approved):
Tested in M1, looks good

src/services/kube/kube.service.ts
```diff
@@ -71,41 +72,57 @@ export class KubeService extends HttpService
...
+    } catch {
+        // Don't try to capture any errors incase the process has already been killed
+    }
+}
+
+async function getPidForPort(port: number) {
+    // Helper function to get a pid from a port number
+    const getPidPromise = new Promise<number>(async (resolve, _) => {
+        pids(port).then((pids: any) => {
+            resolve(pids.tcp[0]);
```
sidpremkumar, line 112: Return all ports, and loop over all of them

sidpremkumar: https://github.com/bastionzero/zli/pull/266/commits/5ab913e7b5ec7decba2cdecde79546e58560b650

lipmas (approved):
Tested this out on linux on both debian/rh based distos and zli disconnect is working :+1:

PR #277, opened 2021-12-25
Catch case when kubeconfig does not exist

sidpremkumar:
## Description of the change

Catch case when kubeconfig does not exist. We now just copy the config we generate to the kubeconfig location instead of trying to flatten when we pass `--update`

Release Notes: Catch case when we pass `--update` to generate kubeConfig and kubeconfig does not exist. 

## Related JIRA tickets

Relates to JIRA: CWC-1379

- [ ] Yes
- [X] No

If yes, please explain:


ymarcus93 (approved):
Looks good. I just can't test `--update` right now because `zli` on M1 does not seem to see my ARM installed `kubectl` binary. I think I have to install the x86 one for it to work, or until we get an ARM version of `zli`.

ymarcus93:
Oh nvm. looks like that got fixed in https://github.com/bastionzero/zli/pull/278

PR #278, opened 2021-12-25
Ensure path var is send to exec in generate kubeConfig

sidpremkumar:
## Description of the change

There was a bug that if kubectl was installed via home-brew, our exec command would not be able to find it. It was because we were passing a custom env to our exec session and that was causing the home-brew `kubectl` not to be found. This PR fixes that by ensure the path env var gets passed down to our exec. 

Release Notes: Fixes issue where home-brew installations of kubectl were not being found

## Related JIRA tickets

Relates to JIRA: CWC-1380

- [ ] Yes
- [X] No

If yes, please explain:


ymarcus93 (approved):
Tested. `zli generate kubeConfig --update` works for me on my M1 Mac. Nice!

PR #283, opened 2022-01-11
Capture more fixes on disconnect and status

sidpremkumar:
## Description of the change

This PR ensures that status and disconnect are working as intended. 

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [X] No

If yes, please explain:


EthanHeilman (approved):
These changes look good to merge. I looked up `killPortProcess` and if I read the code correctly it just kills any process which is using that port.  Kill other processes on a system that we didn't start is pretty aggressive. Is there a better way?

src/handlers/status/status.handler.ts
```diff
@@ -49,7 +49,6 @@ async function webStatusHandler(
             await killPortProcess(webConfig['localPort']);
 
             configService.setWebConfig(webConfig);
-            await cleanExit(0, logger);
```
lgmugnier, line 52: why don't we need the cleanExit anymore?

sidpremkumar: incase the user wants to see more than 1 status: 
2022-01-11 at 5 01 PM

sidpremkumar: https://github.com/bastionzero/zli/pull/283/files#diff-10b5461068de4b9dc04c80862ff4b783437bf57e3009ff7b8741519f6856dfadL16-L24

sidpremkumar:
> These changes look good to merge. I looked up `killPortProcess` and if I read the code correctly it just kills any process which is using that port. Kill other processes on a system that we didn't start is pretty aggressive. Is there a better way?

Yeah this is the implementation we've decided to go with, we need to eventually reevaluate if this is the best way to do things, but we save the port info, so we are only killing a process on a port that we know about. 

Tbh I do not like the way we manage daemons that are spawned and would like to re-design this part at a later date.

PR #284, opened 2022-01-12
update kube policy to drop tunnel

sidpremkumar:
## Description of the change

Removes refs to kube tunnel and update to kube

See related common pr: https://github.com/bastionzero/webshell-common-ts/pull/52

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [x] No

If yes, please explain:


src/handlers/group/add-group-policy.handler.v2.ts
```diff
@@ -40,7 +40,7 @@ export async function addGroupToPolicyHandler(groupName: string, policyName: str
     const policy = kubePolicy ? kubePolicy : targetPolicy;
 
     // If this group exists already
-    const group = policy.groups.find(g => g.name == groupSummary.name);
+    const group = policy.groups.find((g: any) => g.name == groupSummary.name);
```
chad-cwc, line 43: Was this giving an error or warning before you changed it to `any`? Can it be of type `Group` from http/v2/policy/types/group.types.ts?

sidpremkumar: it was giving a warning before. I will update to group

src/handlers/policy/list-kubernetes-policies.handler.ts
```diff
@@ -60,14 +60,14 @@ export async function listKubeTunnelPoliciesHandler(
 
     if(!! argv.json) {
         // json output
-        console.log(JSON.stringify(kubeTunnelPolicies));
+        console.log(JSON.stringify(kubernetesPolicies));
     } else {
-        if (kubeTunnelPolicies.length === 0){
+        if (kubernetesPolicies.length === 0){
             logger.info('There are no available Kubernetes Tunnel policies');
```
chad-cwc, line 66: Remove "Tunnel". Probably need to do a project-wide case-insensitive Find on "tunnel".

src/services/v1/policy/policy.types.ts
```diff
@@ -18,7 +18,9 @@ export enum PolicyType {
     TargetConnect = 'TargetConnect',
     OrganizationControls = 'OrganizationControls',
     SessionRecording = 'SessionRecording',
-    KubernetesTunnel = 'KubernetesTunnel'
+    KubernetesTunnel = 'KubernetesTunnel', // Deprecated in favor of Kubernetes
```
chad-cwc, line 21: I'm trying to think of any reason not to just remove this. Truly deprecating something usually means it's still around in the code but shouldn't be used for new code. I think that's probably not the case here, right?

sidpremkumar: Its an enum, and we cannot delete the enum in our backend (SQL limitation I believe), so we keep it around here so the Kubernetes enum maps to the correct int value

sidpremkumar: Its an enum in our backend, https://github.com/bastionzero/webshell-backend/blob/feat/db-support/Webshell.Database/Database/Schema/Types/PolicyType.cs#L7-L17

And we cannot remove it as there is an SQL limitation that blocks us from this (not 100% but I know we have had issues deleting enums before and have used Obsolete). 

So in order to make sure the mapping works, we keep this around

sidpremkumar: https://github.com/bastionzero/webshell-backend/blob/2fe6627f685b5f93fa4ee9955c7ffbce60840a3d/Webshell.Database/Database/Schema/Types/ConnectionType.cs#L11-L13

chad-cwc: But will the back end ever return the obsolete enum value? (I'm assuming a DB migration will change all old values to the new one, but maybe I'm wrong.) If the back end will never return the obsolete value, there is no reason I can think of to keep the client code in sync with the server code.

Of course, it existing policies will use the obsolete PolicyType value, then we must keep it here.

sidpremkumar: From my understanding these PolicyTypes map to this on the backend: https://github.com/bastionzero/webshell-backend/blob/feat/db-support/Webshell.Database/Database/Schema/Types/PolicyType.cs#L2-L17

So we will never return `KubernetesTunnel` but then how can I make sure that `Kubernetes` maps to the correct enum (i.e. `4`)?

Im not sure of a way to delete / omit a value from an enum type in js and preserver int relationship. For example if I updated this to: 
```js
export enum PolicyType {
     TargetConnect = 'TargetConnect',
     OrganizationControls = 'OrganizationControls',
     SessionRecording = 'SessionRecording',
     Kubernetes = 'Kubernetes',
     Proxy = 'Proxy'
 }
```

Now Kuberneters will map to the enum `3` which is incorrect on our backend.

EthanHeilman:
@chad-cwc Does this PR have any security impact?

sidpremkumar:
> @chad-cwc Does this PR have any security impact?

No but its not a PR into master, it's just a PR into a feature branch.

src/handlers/target-group/list-target-group.handler.v2.ts
```diff
@@ -9,7 +9,7 @@ import { PolicyHttpService } from '../../../src/http-services/policy/policy.http
 export async function listTargetGroupHandler(configService: ConfigService, logger: Logger, argv : yargs.Arguments<targetGroupArgs>, policyName: string) {
 
     const policyHttpService = new PolicyHttpService(configService, logger);
-    const kubePolicies = await policyHttpService.ListKubeTunnelPolicies();
+    const kubePolicies = await policyHttpService.ListKubernetesPolicies();
```
lgmugnier, line 12: There's something about the fact that KubeTunnel takes up the same number of pixels as Kubernetes that is *chef's kiss*

lgmugnier (approved):

src/services/v1/policy/policy.types.ts
sidpremkumar: We are in the middle of trying to get everything synced up so I went ahead and merged this into the feature branch. Please feel free to comment on this PR and I will make the changes in a follow up PR.

PR #287, opened 2022-01-18
Update paths for new controllers

sidpremkumar:
## Description of the change

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [ ] No

If yes, please explain:


src/services/db-target/db-target.service.ts
```diff
@@ -0,0 +1,16 @@
+import { ConfigService } from '../config/config.service';
+import { HttpService } from '../http/http.service';
+import { Logger } from '../logger/logger.service';
+import { DbTargetSummary } from './db-target.types';
+
+export class DbTargetService extends HttpService
+{
+    constructor(configService: ConfigService, logger: Logger)
+    {
+        super(configService, 'api/v2/targets/db', logger);
```
chad-cwc, line 10: The path should be: api/v2/targets/database

sidpremkumar: Please see our slack conversation :) , happy to change it but elsewhere we have been using `db`.

chad-cwc: How did this work? api/v2/targets/databases <-- see the `s` at the end? We want it to be no `s`, and I see your back end update has no `s`.

PR #294, opened 2022-01-25
Solve panics

lgmugnier:
## Description of the change

Make it so we don't need panics anymore and we can elegantly bubble up these errors to the user.

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [x] No

If yes, please explain:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bastionzero/zli",
-  "version": "5.4.0",
+  "version": "5.4.1",
```
sidpremkumar, line 3: revert this, since we are merging into the feature branch no need to increment version. We can increment the version when we merge `feat/db-support` -> `master`

src/handlers/connect/web-connect.handler.ts
```diff
@@ -116,18 +115,18 @@ export async function webConnectHandler(argv: yargs.Arguments<connectArgs>, targ
             // Now save the Pid so we can kill the process next time we start it
             webConfig['localPid'] = daemonProcess.pid;
             webConfig['localPort'] = localPort;
-            webConfig['localHost'] = 'localhost'
+            webConfig['localHost'] = 'localhost';
```
sidpremkumar, line 118: this should not be set to `'localhost'` rather it should be set to `localHost` the variable found on line 64

src/utils/daemon-utils.ts
```diff
@@ -3,16 +3,45 @@ import fs from 'fs';
...
 const exec = require('child_process').execSync;
 const pids = require('port-pid');
+const readLastLines = require('read-last-lines');
 
 export const WINDOWS_DAEMON_PATH : string = 'bzero/bctl/daemon/daemon-windows';
 export const LINUX_DAEMON_PATH   : string = 'bzero/bctl/daemon/daemon-linux';
 export const MACOS_DAEMON_PATH   : string = 'bzero/bctl/daemon/daemon-macos';
 
+// Allow errors on early daemon startup to bubble up to the user
+export async function handleServerStart(logPath: string, localPort: any) {
```
sidpremkumar, line 18: Couple of things: 
* `localPort` I believe should be an int so it can read `localPort: number` 
* we need to pass `localHost` as well here since for web and db targets we can have custom host. So instead of hardcoding the string `'localhost'` pass the host from the calling function.

sidpremkumar (changes requested):

package.json
lgmugnier: 2974b54

src/handlers/connect/web-connect.handler.ts
lgmugnier: 2974b54

src/utils/daemon-utils.ts
lgmugnier: 2974b54

src/handlers/connect/db-connect.handler.ts
```diff
@@ -122,8 +122,9 @@ export async function dbConnectHandler(argv: yargs.Arguments<connectArgs>, targe
             dbConfig['name'] = dbTarget.name;
 
             // Wait for daemon HTTP server to be bound and running
-            await waitUntilUsedOnHost(localPort, 'localhost', 100, 1000 * 20);
-            logger.info(`Started db daemon at ${dbConfig['localHost']}:${localPort} for ${targetName}`); 
+            await handleServerStart(loggerConfigService.daemonLogPath(), localPort, localHost);
+
+            logger.info(`Started db daemon at ${dbConfig['localHost']}:${localPort} for ${targetName}`);
```
sidpremkumar, line 127: can you change this to either use `dbConfig` for both host + port or `localHost` and `localPort` (the vars)

src/utils/daemon-utils.ts
```diff
@@ -3,16 +3,43 @@ import fs from 'fs';
...
+    await new Promise<void>(async (resolve, reject) => {
+        await waitUntilUsedOnHost(localPort, localHost, 100, 1000 * 5).then(function() {
+            resolve();
+        }, function() {
+            if (fs.existsSync(logPath)) {
+                readLastLines.read(logPath, 1)
+                    .then((line: string) => {
+                        try {
+                            const lastLog = JSON.parse(line);
+                            reject(`failed daemon startup\nLast log entry: ${lastLog.message}`);
```
sidpremkumar, line 28: Since this will be printed to the user, lets capitalize the first letters of all these reject statements?

sidpremkumar (approved):
minor nits but this is ready to merge overall 🥳

PR #298, opened 2022-01-27
Fix/webserver agent

sidpremkumar:
## Description of the change

This is the zli PR that is related to this: https://github.com/bastionzero/bzero/pull/21

It adds the ability for the web connect to generate its own cert. 

@EthanHeilman could you please take a look at this and can we plan to sync up. I'm not sure if we want to go down the path of the daemon having a https connection, since this results in a insecure issues in the browser (as its a self signed cert). 
Maybe you know of ways to get around that. 

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [ ] No

If yes, please explain:


src/handlers/connect/web-connect.handler.ts
```diff
@@ -73,11 +74,31 @@ export async function webConnectHandler(argv: yargs.Arguments<connectArgs>, targ
...
+
+        const pathToConfig = path.dirname(configService.configPath());
+        const configName = configService.getConfigName();
+
+
+        const [pathToKey, pathToCert, pathToCsr] = await generateNewCert(pathToConfig, 'web', configName);
+
+        webConfig['keyPath'] = pathToKey;
+        webConfig['certPath'] = pathToCert;
+        webConfig['csrPath'] = pathToCsr;
```
lgmugnier, line 89: would you mind leaving a little comment (doesn't have to be here) about why we need to key/cert and what's a csr?

sidpremkumar: https://github.com/bastionzero/zli/pull/298/commits/c5580bacb28a5809469c65c8ba34aa53b526ed57

lgmugnier (approved):

PR #307, opened 2022-02-11
Ensure we are building zli daemons correctly for multiple OS' 

sidpremkumar:
## Description of the change

Had an issue where windows zli was not being built correctly, this has fixed that: 
image

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [ ] No

If yes, please explain:


src/utils/daemon-utils.ts
```diff
@@ -199,16 +197,12 @@ export async function copyExecutableToLocalDir(logger: Logger, configPath: strin
     let daemonExecPath = undefined;
     let finalDaemonPath = undefined;
     if (process.platform === 'win32') {
-        daemonExecPath = path.join(prefix, WINDOWS_DAEMON_PATH);
+        daemonExecPath = path.join(prefix, DAEMON_PATH);
 
         finalDaemonPath = path.join(configFileDir, 'daemon-windows.exe');
```
lipmas, line 202: This line can be removed now i think.

package.json
```diff
@@ -15,15 +15,18 @@
   },
   "scripts": {
     "start": "ts-node src/index.ts",
-    "build": "npm run build-binaries && npm run build-clean",
-    "build-binaries": "tsc && cd bzero/bctl/daemon && GOOS=darwin GOARCH=amd64 go build -o daemon-macos && GOOS=windows GOARCH=amd64 go build -o daemon-windows && GOOS=linux GOARCH=amd64 go build -o daemon-linux",
-    "build-clean": "ts-node clean-build.ts",
+    "build-binaries-macos": "tsc && cd bzero/bctl/daemon && GOOS=darwin GOARCH=amd64 go build -o daemon",
+    "build-binaries-windows": "tsc && cd bzero/bctl/daemon && GOOS=windows GOARCH=amd64 go build -o daemon",
+    "build-binaries-linux": "tsc && cd bzero/bctl/daemon && GOOS=linux GOARCH=amd64 go build -o daemon-linux",
```
lipmas, line 20: should this be `-o daemon` as well?

package.json
```diff
@@ -15,15 +15,18 @@
...
     "update-daemon-version": "replace '\\$DAEMON_VERSION' \"$(cat bzero/VERSION)\" bzero/bctl/daemon/daemon.go",
-    "package": "npm run copy-files && pkg .",
-    "release": "npm run build && npm run package",
-    "release-prod": "npm run update-daemon-version && npm run release",
     "release-dev": "tsc && npm run package",
     "copy-files": "copyfiles src/**/*.html src/**/*.css bzero/bctl/daemon/* dist/",
+    "package": "npm run copy-files && pkg .",
+    "release": "npm run clean && npm run release-macos && npm run release-linux && npm run release-windows",
```
lipmas, line 25: How does release work now? Wont this just overwrite the daemon executable each time it builds for a different OS? Shouldnt release now detect which OS its being run on and just call the appropriate release-XXX script?

sidpremkumar: Under `release-macos`: 
```
"release-macos": "npm run build-binaries-macos && npm run copy-files && pkg . -t node14-macos-x64 -o bin/zli-macos",
```
We go through the whole release process, from building the daemon binaries, to packaging. 

So for each operating system, we build the daemon binaries -> zli binaries. 

So its okay if it overwrites it since when it does that we have moved on to the next OS.

src/utils/daemon-utils.ts
sidpremkumar: We actually still use that, but I've updated it to just use `daemon.exec` https://github.com/bastionzero/zli/blob/fix/release-process/src/utils/daemon-utils.ts#L219

package.json
sidpremkumar: https://github.com/bastionzero/zli/pull/307/commits/cc8f902bde5920dd9d4f6ac5ac037914192d9a83

src/utils/daemon-utils.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/307/commits/cc8f902bde5920dd9d4f6ac5ac037914192d9a83

sidpremkumar:
See that system tests are failing w a new error. 
2022-02-11 at 1 21 PM

Seems to be working locally. Investigating...

sidpremkumar:
Tested on al2 and it seems to be working: 
2022-02-11 at 1 47 PM

Super strange sys test are not passing

package.json
lipmas: Oh i see yeah that makes sense. I was that the daemon just gets bundled up by pkg to produce the single executable before the next os release code happens.

sidpremkumar:
System tests are passing now (for db, web, kube) but we still have that issue with connections and I think that's causing shell connections not to be created.

lipmas (approved):
I think this is a fine change to make but might be worth mentioning in documentation somewhere that the correct way to build the zli locally our your machine is now to use the `npm run release-xxx`. If developers use `npm run release` its going to build the windows one last and leave the windows daemon executable in their config directory.

lipmas:
Ok actually i have another proposal what about using this [npm package](https://www.npmjs.com/package/run-script-os) to detect which os developer is on and automatically call the right one. Then having a separate script `release-all` which builds all of them.

sidpremkumar:
Tested and it auto detected my os: 
2022-02-11 at 2 44 PM

lipmas:
Nice also worked for me on linux:
image

lipmas (approved):

PR #311, opened 2022-02-12
Add more system tests

sidpremkumar:
## Description of the change

See related webshell-common-ts PR here: https://github.com/bastionzero/webshell-common-ts/pull/58

This PR fixes our VT tests (by switching to `pg` from `pgtools` and also adds support for even more tests:
* Create/Delete a unique environment for the specific system tests
* Create/Delete a unique targetConnect and proxy policy for the specific system tests
* Checks for Kube, ssm and vt's that there are expected connection events in Bastion (opened -> closed)
* Checks to ensure that bzero-ssm-targets and bzero targets are correctly spun up in the custom environment we create for the `list-targets` test
* Checks to ensure that there is a kube log event for our call to list namespaces

Release Notes: Adds more system tests for Kube, Ssm and Vt tests

## Related JIRA tickets

Relates to JIRA: CWC-1373

- [ ] Yes
- [ ] No

If yes, please explain:


src/system-tests/tests/suites/kube.ts
```diff
@@ -15,39 +25,137 @@ export const kubeSuite = () => {
...
+
+            // Ensure that the connect events were created for this connection
+            expect(await testUtils.EnsureConnectionEventCreated(doCluster.bzeroClusterTargetSummary.id, 'CLUSTER'));
+
+            // Ensure that we see a log of this under the kube logs
+            expect(await testUtils.EnsureKubeEvent(doCluster.bzeroClusterTargetSummary.name, KubeTestUserName, KubeTestTargetGroup, 'N/A', ['/api/v1/namespaces'], []));
+            testPassed = true;
+        }, 30 * 1000);
+
+        test.each(clusterVersionsToRun)('zli targetUser - add target user to policy %p', async (clusterVersion) => {
```
lipmas, line 85: So these are in the kube test suite rather than a new policy suite because they are interacting with a kube specific policy  right? Trying to think if that makes sense or not. Adding a user to  a policy should be same functionality across all policies type.

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (!fs.existsSync(daemonLogPath)) {
+            this.logger.warn(`No daemon logs found under ${daemonLogPath}`);
+            return;
+        };
+
+        if (!testPassed) {
+            // Print the logs from the daemon
+            try {
+                const daemonLogs = fs.readFileSync(daemonLogPath, 'utf8');
+                this.logger.error(`Test failed! Daemon logs:\n${daemonLogs}`);
```
lipmas, line 146: Might be nice to also include the test name here? Or is it obvious from how it shows up in the logs?

src/system-tests/tests/system-test.ts
```diff
@@ -135,10 +157,19 @@ afterAll(async () => {
...
     await checkAllSettledPromise(Promise.allSettled([
         cleanupDOTestTargets(),
         cleanupDOTestClusters()
     ]));
+
+    // Delete the environment for this system test
+    // Note this must be called after our cleanup, so we do not have any targets in the environment
+    const listOfEnvs = await environmentService.ListEnvironments();
+    const systemTestEnv = await getEnvironmentFromName(systemTestEnvName, listOfEnvs, logger);
+    await environmentService.DeleteEnvironment(systemTestEnv.id);
```
lipmas, line 172: why do we need to list environments here cant we just check if the global `systemTestEnvId` was set by the setup code:

```
if( systemTestEnvId) {
   await environmentService.DeleteEnvironment(systemTestEnv.id);
}
```

src/system-tests/tests/system-test.ts
```diff
@@ -510,6 +535,63 @@ async function cleanupSystemTestApiKeys() {
     await apiKeyService.DeleteApiKey(systemTestRegistrationApiKey.apiKeyDetails.id);
 }
 
+/**
+ * Helper function to create all needed policies for our system test
+ * Note: We do not create the kube policy since that will be created on cluster registration
+ * @param {string} environmentId Environment Id for the unique env created for this system test
+ */
+async function createAllNeededPolicies(environmentId: string) {
```
lipmas, line 543: The only thing im wondering about is from code organization point of view do we want to move the policy creation/deletion to a `beforeAll` and `afterAll` function in each specific test suite. That way the logic for creating the target specific policy is contained in the test suite for that type of target.

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -0,0 +1,175 @@
...
+        this.logger = logger;
+    };
+
+    /**
+     * Helper function to ensure that a connection event was created
+     * (i.e. client connect -> closed exists in our db events logs)
+     * @param {string} targetId Target id we are looking for
+     * @param {string} targetType Target type we are looking for (i.e. CLUSTER)
+     */
+    public async EnsureConnectionEventCreated(targetId: string, targetType: string) {
```
lipmas, line 33: Why not pass the specific `ConnectionEventType` you are expecting to find as a parameter to this function. That way tests can be more explicit about which types of events should be created and when. They can also call this multiple times to assert both a create/closed event were created for example

src/system-tests/tests/suites/kube.ts
sidpremkumar: I figured they were specific to kube, and we could add specific policy tests for that type under that test (bc then it can assume that the policy will exist). 

And I will move the policy creation to `beforeAll` (and removal `afterAll`).

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Helper function to ensure we see a given kube event
+     * @param {string} targetName Target name we are hitting
+     * @param {string} role Target role we want to connect as
+     * @param {string} targetGroup Target group we want to connect as
+     * @param {string} kubeEnglishCommand Kube english command we are expecting
+     * @param {string[]} expectedEndpoints List of expected endpoints to have been hit
+     */
+    public async EnsureKubeEvent(targetName: string, role: string, targetGroup: string, kubeEnglishCommand: string, expectedEndpoints: string[], expectedExecs: string[]) {
```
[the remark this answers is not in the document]
sidpremkumar: https://github.com/bastionzero/zli/pull/311/commits/1a53dcf2aaf65a9095a5505075dc1e91bec6b2d4

src/system-tests/tests/utils/test-utils.ts
sidpremkumar: Its usually pretty obvious from the logs, but could you explain to me what you meant by passing test name? 
How could we get test name from inside the test (I don't think I know jest well enough :( )

src/system-tests/tests/system-test.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/311/commits/23dbddfe4744f342a6d622caed6bda865811f695

src/system-tests/tests/system-test.ts
sidpremkumar: I think it makes sense to put in beforeAll after allhttps://github.com/bastionzero/zli/pull/311/commits/fcb93206dc482a034b718d2e58da5a67e75c921f

src/system-tests/tests/utils/test-utils.ts
sidpremkumar: Hmm so I guess if I did that I would be able to test if the `Created` event and `Connected` even were created after I run `zli connect`, and then after I could check `Disconnected` and `Closed`. 

My only question would be is there any advantage to that? Since we would be checking the same set of events (Created -> Closed) either way? Err maybe I'm missing something. It might be useful if we check the negation (i..e `Created` and `Connected` event exist, but no `Disconnected` or `Closed` but I'm not sure if we want to add that.

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -0,0 +1,175 @@
...
+
+const fs = require('fs');
+const pids = require('port-pid');
+
+const EVENT_QUERY_TIME = 2;
+
+/**
+ * Class that contains our common testing functions that can be used across tests
+ */
+export class TestUtils {
```
ymarcus93, line 16: In general, I would prefer if we could use an equivalent `jest.expect` _matcher_ (https://jestjs.io/docs/expect) to assert things in the tests. That way we get a better error message when one of the expectations is false / incorrect. Right now, it looks like the `Ensure` functions of these tests check a lot of things in one function and then assert that the everything evaluates to `true`.

So if one of the nested expectations fails, `jest` won't be able to tell us what specifically failed--it will just say that the `Ensure` method returned `false`.

src/system-tests/tests/utils/test-utils.ts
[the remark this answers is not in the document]
sidpremkumar: So it looks like impersonating a non-existent group will result in a rejection: 
```
➜  zli git:(feat/more-system-tests) ✗ kubectl auth can-i  get namespace --as=foo --as-group=system:masters,foo
Warning: resource 'namespaces' is not namespace scoped
no
```
So I split up the test, 1 connect to use `system:masters` where we check that the request was able to be executed, and then another one where we just impersonate multiple groups and ensure that it exists.

https://github.com/bastionzero/zli/pull/311/commits/45b8cd97622630f951fc2f3d1ed8042a4005272b

src/system-tests/tests/utils/test-utils.ts
lipmas: I think the advantage is the flexibility/extensibility it will give us for augmenting connection event checks in the future. Right now this helper function assumes that exactly one `createdEvent`, `connectEvent`, `disconnectEvent`, and `closeEvent` will exist in the connection events and will fail otherwise. What if we wanted to write a test that should generate a different set of events (like multiple `connectEvent`s or something). Also if this were per event type we could extend the tests to check more specific things for example checking what the `closeReason` was for a `closeEvent`.

src/system-tests/tests/utils/test-utils.ts
lipmas: Havent tested it but maybe like this:  https://stackoverflow.com/a/62781554

src/system-tests/tests/suites/kube.ts
```diff
@@ -1,53 +1,213 @@
...
+import { configService, logger } from '../system-test';
+import { TestUtils } from '../utils/test-utils';
+
+const fs = require('fs');
 
 export const KubeTestUserName = 'foo';
+export const KubeTestTargetGroups = ['system:masters', 'foo'];
+export const KubeBctlNamespace = 'bastionzero';
 
 export const kubeSuite = () => {
```
lipmas, line 14: @sidpremkumar Looks like we are missing a `beforeAll` function in the kube suite that will create the `kubernetes` policy. Running into this issue when trying to run https://github.com/bastionzero/zli/pull/315 against these system tests

sidpremkumar: That policy should be created with the helm chart that is why I left I did not add it here.

src/system-tests/tests/utils/test-utils.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/311/commits/9b067805df3b9ad961b17f06b818945734aa21e7

src/system-tests/tests/utils/test-utils.ts
sidpremkumar: Gotcha, tried to handle that and @ymarcus93 comment regarding the expects in this commit: https://github.com/bastionzero/zli/pull/311/commits/d873f4cdb1358bcd6ab2bcd68f9ab4847a9bd2f1

Now we will do what you mentioned where we check if each event exists individually, and we expect the result so if one doesn't exist we will be able to see that in the logs of the test.

src/system-tests/tests/utils/test-utils.ts
sidpremkumar: https://github.com/bastionzero/zli/pull/311/commits/d873f4cdb1358bcd6ab2bcd68f9ab4847a9bd2f1

I figured it might be okay to leave the kube log event test the way it is, since we are expecting a single kube logs, and if it fails we will know what log we failed on. Lmk if that was an ok assumption.

lipmas (approved):

ymarcus93 (approved):

PR #313, opened 2022-02-17
BZero Target Shell Support 

lgmugnier:
## Description of the change

Related Feature Branch PRs:

https://github.com/bastionzero/webshell-backend/pull/1012
https://github.com/bastionzero/webshell-common-ts/pull/60
https://github.com/bastionzero/bzero/pull/55

### Dependencies

backend: In order to test the shell plugin you need to be on the latest `feat/shell` branch of webshell-backend https://github.com/bastionzero/webshell-backend/pull/1012 (which includes running a db migration).

bzero: Make sure you are running an agent rsync'd from the `feat/shell` branch:

### Connecting to a bzero target

You should be able to connect directly to a bzero target provided you have a target access policy that allows access to that target or environment. If you are testing against a development agent spawned via `bzero-dev` you will need to make sure to connect as target user: `ec2-user`.  

Then Simply run the `zli connect` command. e.g

```
./bin/zli-linux connect ec2-user@sebby-bzero-agent
```

Anything is fair game to test in a shell connection but some things to try out specifically are:

1. Using an editor (vim, emacs, nano).
2. Resizing the terminal dimensions.
3. Copy-pasting output into the terminal. We debounce StdIn  so that copy+paste should be fast and not send each keypress individually.
4. Making sure command extraction is working properly (from the webapp navigate to logs->commands)
5. Making sure session recording is working properly. Add a session recording policy then open a shell connection, do some things and make sure the session recording looks accurate.
6. Closing connection. Typing `exit` or `CTRL+d` should kill the shell connection and exit normally without any error output (also process should exit with error code 0.

### Attaching to a bzero target

1. First connect to a bzero target in a terminal.
2. Then  in a new terminal list your open zli connections via `zli list-connections`. Note the `connectionId` of the connection.
3. Attach to the connection in the second terminal using `zli attach <connection-id>
4. Once attached the shell terminal output should be replayed to restore the state of terminal output from the other terminal window. Then you should be able to use the attached terminal window normally (but dont use the original one- see note). You can close the original terminal window and the attached terminal should continue working though.

**Note**: There is a know issue right now with the shell attach flow where as soon as you try typing any input in the original terminal window the shell connection will get closed and both terminals will exit. This is due to the fact that the first terminal connecting causes a keysplitting validation error when it tries to send a new input and currently both clients process that error. @lgmugnier pipelining changes will address this so that clients will start ignoring keysplitting validation errors that do not originate from them (have a different SYN hpointer) to prevent this bug.   

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1417

- [ ] Yes
- [x] No

If yes, please explain:


johncmerfeld:
Confirmed items 1-6 above work!

Attaching didn't work for me but I think that's a separate issue on my end:
```
zli attach  aed56117-eec8-4539-ac61-89a4f230612b

Logged in as: john@commonwealthcrypto.com, bzero-id:1203284b-9ce5-4929-98a7-61053c3ba6fa, session-id:2309962d-d4db-436d-a3e2-8d49cd32805d
Failed to fetch dynamic access configs: Error: Authentication Error:
Response code 401 (Unauthorized)
Failed to fetch ssm targets: Error: Authentication Error:
Response code 401 (Unauthorized)
Authentication Error:
Response code 401 (Unauthorized)
```

src/handlers/connect/shell-connect.handler.ts
```diff
@@ -75,20 +86,7 @@ export async function connectHandler(
...
-
-    if(! connectionId)
-    {
-        logger.error('Connection creation failed');
-
-        logger.error(`You may not have a policy for targetUser ${parsedTarget.user} in environment ${parsedTarget.envName}`);
-        logger.info('You can find SSM user policies in the web app');
-
-        await cleanExit(1, logger);
-    }
```
tflpd, line 91: Why isn't this needed anymore?

src/handlers/middleware.handler.ts
```diff
@@ -67,6 +70,16 @@ export function fetchDataMiddleware(configService: ConfigService, logger: Logger
...
 
+    const bzeroTargets = new Promise<BzeroAgentSummary[]>( async (res) => {
+        try {
+            const response = await bzeroHttpService.ListBzeroTargets();
+            res(response);
+        } catch (e: any) {
+            logger.error(`Failed to fetch bzero targets: ${e}`);
+            res([]);
+        }
+    });
```
tflpd, line 81: Didn't we say that we will stop prefetching? 😅

src/terminal/bzero-terminal.ts
```diff
@@ -0,0 +1,159 @@
...
+import { ConnectionHttpService } from '../http-services/connection/connection.http-services';
+import { BzeroTargetHttpService } from '../http-services/targets/bzero/bzero.http-services';
+
+export class ShellTerminal implements IDisposable
+{
+    private shellWebsocketService : IShellWebsocketService;
+    private shellEventDataSubscription: Subscription;
+    private currentTerminalSize: TerminalSize;
+
+    // private refreshTargetInfoOnReady: boolean = false;
```
tflpd, line 20: Remove unused code?

tflpd (approved):

src/utils/utils.ts
```diff
@@ -984,7 +1004,20 @@ export function dynamicConfigToTargetSummary(config: DynamicAccessConfigSummary)
         agentPublicKey: 'N/A'
     };
 }
-/**
+
+export function bzeroTargetToTargetSummary(bzeroTarget: BzeroAgentSummary) {
```
johncmerfeld, line 1008: Might be worth having a docstring explaining how / why this is used

src/handlers/middleware.handler.ts
lipmas: Yes we did...but currently we still use this fetch middleware pattern for connect and in particular for the `disambiguateTarget` util function. So i agree we should remove this but id rather fix it by refactoring the `disambiugateTarget` function for all of the targets instead of having it be hybrid solution. Could make the change here but i think its going to get a little messy and id rather not couple those refactors with the shell feature.

src/terminal/bzero-terminal.ts
lipmas: This entire file actually wasnt needed anymore now that we moved the terminal wrapping code to daemon and out  of typescript. [06611ba](https://github.com/bastionzero/zli/pull/313/commits/06611baf7b77bfbaa2697080bac65e87f1127eff)

src/utils/utils.ts
lipmas: Ok added some docstrings in [30f82d7](https://github.com/bastionzero/zli/pull/313/commits/30f82d7e7f5bc9acdf1cd2ee7c77d329b185a11d). These are just helper functions we also have for ssm/dynamic configs to convert all the targets into a common `TargetSummary` interface.

src/handlers/connect/shell-connect.handler.ts
lipmas: This code was previously catching an error that occurs in `CreateConnection` and returning undefined:

```
// 'TypeError: Cannot read property 'connectionId' of undefined'
// so we need to catch and return undefined
const connectionId = await connectionHttpService.CreateConnection(parsedTarget.type, parsedTarget.id, cliSpaceId, targetUser).catch(() => undefined);    
```

I think this code was from a very long time ago and was just suppressing the actual error that occurred. Now if an error occurs in the `connectionHttpService` it should be treated as a fatal error for that `zli connect` command and the error should just be caught at the top level by yargs and printed to the user. The code was also assuming an error could happen because of policy but that also doesnt really apply anymore since most policy errors should get caught [here](https://github.com/bastionzero/zli/blob/755af9ef2f093401a1ca351cbbea7c4ed24cf1ba/src/handlers/connect/shell-connect.handler.ts#L49-L74) before the `CreateConnection` call even happens.

src/http-services/targets/bzero/bzero.http-services.ts
```diff
@@ -0,0 +1,22 @@
+import { BzeroAgentSummary } from '../../../../webshell-common-ts/http/v2/target/bzero/types/bzero-agent-summary.types';
+import { ConfigService } from '../../../services/config/config.service';
+import { HttpService } from '../../../services/http/http.service';
+import { Logger } from '../../../services/logger/logger.service';
+
+export class BzeroTargetHttpService extends HttpService {
```
sidpremkumar, line 6: Isn't this a duplcate of : https://github.com/bastionzero/zli/blob/feat/shell/src/http-services/bzero-agent/bzero-agent.http-service.ts#L1-L24

sidpremkumar (approved):
Testing was done in backend PR

src/http-services/targets/bzero/bzero.http-services.ts
lipmas: Yup it was nice catch. Got rid of that one and converted everywhere to consistently use `BzeroTargetHttpService` instead. [d26d39d](https://github.com/bastionzero/zli/pull/313/commits/d26d39d4546e1a9dc890ed116863672aafdb555a)

src/handlers/list-connections/list-connections.handler.ts
```diff
@@ -26,8 +26,14 @@ export async function listConnectionsHandler(
...
 
-    // await and concatenate
-    const allTargets = [...await ssmTargets];
+    const ssmTargetHttpService = new SsmTargetHttpService(configService, logger);
+    const ssmTargets = await ssmTargetHttpService.ListSsmTargets(true);
+
+    const bzeroAgentService = new BzeroAgentService(configService, logger);
+    const bzeroTargets = await bzeroAgentService.ListBzeroAgents();
+
+    const allTargets = [...ssmTargets.map(ssmTargetToTargetSummary), ...bzeroTargets.map(bzeroTargetToTargetSummary)];
```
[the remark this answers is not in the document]
lipmas: I think it still does cause you can still close those connections with `zli close <connection-id>`. With [0a3369f](https://github.com/bastionzero/zli/pull/313/commits/0a3369fd8b6b652869d9d683519548fad09688e4) if they do try and attach to a bzero connection they will get an error message that says its currently unsupported.

sidpremkumar (approved):

PR #319, opened 2022-02-23
Fix CWC-1559: Check id_token's exp time instead of access_token's exp

ymarcus93:
## Description of the change

Fixes CWC-1559.

Check id_token's expiration time instead of access_token's expiration time in `oauth.service`'s `getIdToken()`.

In addition, I've added a 5 minute buffer time, so we refresh a little bit before the token actually expires.

Release Notes: Fixes CWC-1559

## Related JIRA tickets

Relates to JIRA: CWC-1559

- [ ] Yes
- [x] No

If yes, please explain:


src/services/oauth/oauth.service.ts
```diff
@@ -290,7 +290,9 @@ export class OAuthService implements IDisposable {
 
         // Refresh if the token exists and has expired
         if (tokenSet) {
-            if (this.configService.tokenSet().expired()) {
+            const nowUnixEpochTime = Math.floor(Date.now() / 1000);
+            const bufferMinutes = 5;
+            if ((nowUnixEpochTime + 60 * bufferMinutes) >= tokenSet.claims().exp) {
```
lipmas, line 295: We might also want to change this in `isAuthenticated` [ref] (https://github.com/bastionzero/zli/blob/master/src/services/oauth/oauth.service.ts#L221) because we are also using the `tokenSet.expired()` method there. 

We use it in the login handler to determine if we need to generate new nonce. https://github.com/bastionzero/zli/blob/master/src/handlers/login/login.handler.ts#L54

ymarcus93: I don't think https://github.com/bastionzero/zli/blob/e40f374994edcee016dea0401a7189acfc0f3fcb/src/services/oauth/oauth.service.ts#L221 can ever be hit

because https://github.com/bastionzero/zli/blob/312287546b1210a0662b81ad4fca4d41a4f05ace/src/services/oauth/oauth.service.ts#L217-L219

will be hit first, and `tokenSet` is cleared here:

https://github.com/bastionzero/zli/blob/67bfb90535c20698a881c3dc26511c445f206324/src/handlers/login/login.handler.ts#L49

ymarcus93: Maybe I should just remove `return !tokenSet.expired();`?

lipmas: Doesnt it make sense to keep that check in `Authenticated`  though in case we call it from somewhere else besides the login handler

ymarcus93: True. That makes sense. `isAuthenticated` is `public`, so others beside `login` might use it one day

lipmas (approved):

src/services/oauth/oauth.service.ts
ymarcus93: I put it in a private function here: [ad64ffd](https://github.com/bastionzero/zli/pull/319/commits/ad64ffd1aedbaa12ad62e9b4d962f68199cba1ce)

PR #323, opened 2022-02-24
Fix listing policies for TargetConnect and Kubernetes

sidpremkumar:
## Description of the change

We were comparing to check to see if the value was defined, but we return a empty list either way. So this PR ensures we display the correct value (env or target) depending on the length of the list. 

2022-02-24 at 3 14 PM

Release Notes: Fixes issues listing Kube and Target policies that have targets

## Related JIRA tickets

Relates to JIRA: CWC-1563

- [ ] Yes
- [X] No

If yes, please explain:


src/cli-driver.ts
```diff
@@ -205,7 +205,8 @@ export class CliDriver
     private targetStatusChoices: string[] = Object.keys(TargetStatus).map(s => s.toLowerCase());
 
     // available options for PolicyType autogenerated from enum
-    private policyTypeChoices: string[] = Object.keys(PolicyType).map(s => s.toLowerCase());
+    // We filter out the deprecated Kube tunnel
+    private policyTypeChoices: string[] = Object.keys(PolicyType).filter(s => s != PolicyType.KubernetesTunnel).map(s => s.toLowerCase());
```
lipmas, line 209: can we just delete `PolicyType.KubernetesTunnel` from the enum  now it should have 0 refs right and all policies were migrated in the db

sidpremkumar: :blob-salute:

lipmas (approved):
Looks good. We could also add environment/targets logic to the `Proxy` policy table too - im not sure why but this was never added to proxy policy even though it should also have targets/environments in the context.
https://github.com/bastionzero/zli/blob/b58118ddef0404f118d4622172ad57ca8b24d879/src/utils/utils.ts#L607-L655

sidpremkumar:
> 

Yeah not sure why I didnt add it. Fixed it though here: 
2022-02-24 at 3 35 PM

lipmas (approved):
LGTM

PR #327, opened 2022-03-04
Microsoft prompt option during OIDC login [CWC-1579]

tflpd:
## Description of the change

Adjusted Microsoft prompt option during OIDC login to match the web app

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1579

- [ ] Yes
- [X] No

If yes, please explain:


src/services/oauth/oauth.service.ts
```diff
@@ -195,14 +195,34 @@ export class OAuthService implements IDisposable {
             throw new Error('Unable to get authUrl from with undefined nonce');
         }
 
+        const idp = this.configService.idp();
+        if(idp === undefined){
+            throw new Error('Unable to get authUrl from with undefined idp');
```
chad-cwc, line 200: I can't tell if there's a missing word between "from" and "with" or if one of them should be removed. Either way, this (and line 195) need a minor adjustment.

tflpd: Haha, yup fixing it!

tflpd: [1e9b014](https://github.com/bastionzero/zli/pull/327/commits/1e9b014ad863a0e3e4adf08ac33bcae4419758f9)

src/services/oauth/oauth.service.ts
```diff
@@ -192,7 +192,27 @@ export class OAuthService implements IDisposable {
...
+            throw new Error('Unable to get authUrl with undefined nonce');
+        }
+
+        const idp = this.configService.idp();
+        if(idp === undefined){
+            throw new Error('Unable to get authUrl from undefined idp');
+        }
+
+        let prompt = '';
+        switch (idp) {
```
sidpremkumar, line 204: This could be simplified:
```js
switch (idp) {
    case IdentityProvider.Google:
    case IdentityProvider.Okta:
        prompt = 'consent';
        break;
....
````

File: src/services/oauth/oauth.service.ts [hunk 3]
```diff
@@ -192,7 +192,27 @@ export class OAuthService implements IDisposable {
...
+                prompt = 'consent';
+                break;
+            case IdentityProvider.Okta:
+                prompt = 'consent';
+                break;
+            case IdentityProvider.Microsoft:
+                prompt = 'login';
+                break;
+            default:
+                throw new Error('Unsupported IdP ' + idp);
```
sidpremkumar, line 215: We usually use the notation 
```
throw new Error(`Unsupported IdP: ${idp}`);
```

sidpremkumar (approved):
nits but LGTM

src/services/oauth/oauth.service.ts
lipmas, line 204: I could suggest moving this to a function inside `config.service` like we do for the oauth scopes which also differ per idp. But not sure its actually the best place to keep putting things like this. Feels like it could be useful at some point to make a base class with abstract methods that different ouath providers can implement. Clearly this PR isnt the place for this refactor though.

lipmas (approved):

src/services/oauth/oauth.service.ts
tflpd: [717950f](https://github.com/bastionzero/zli/pull/327/commits/717950fbfd505094d7fb16bc5bc00b27194b7f94)

src/services/oauth/oauth.service.ts
tflpd: [717950f](https://github.com/bastionzero/zli/pull/327/commits/717950fbfd505094d7fb16bc5bc00b27194b7f94)

src/services/oauth/oauth.service.ts
tflpd: Yeah, I have been thinking about this as well, especially the webapp file is getting chaotic.. Will try to squeeze that in as soon as I have some cycles

PR #328, opened 2022-03-04
Simplify agent public key pass + System test fixes

sidpremkumar:
## Description of the change

This PR: 
* Update the bzero submodule version 
* Simplifies some of the building args to now take a agentPublicKey since we share that across all our agents. 
* Update the DO Ubuntu snapshot image
    * I noticed when I ssh'd into the boxes as system tests was running that we would see a huge spike in CPU on the ubuntu box: 
        * image
     * This spike could take up 100% of the single CPU that the tests ran on. I followed these instructions: https://linuxhint.com/enable-disable-unattended-upgrades-ubuntu/ in order to disable the auto upgrade and recreated a ubuntu snapshot + uploaded that to DO.
     * Snap daemon refresh was also causing similar spikes so I disabled that as well: https://dannyda.com/2021/04/13/how-to-disable-snap-snapd-services-in-ubuntu-20-04-1-lts-how-to-uninstall-remove-snap/

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

- [ ] Yes
- [X] No

If yes, please explain:


src/system-tests/tests/suites/connect.ts
```diff
@@ -51,7 +51,7 @@ export const connectSuite = () => {
                 policy.name == systemTestPolicyTemplate.replace('$POLICY_TYPE', 'target-connect')
             );
             policyService.DeleteTargetConnectPolicy(targetConnectPolicy.id);
-        });
+        }, 15 * 1000);
```
sidpremkumar, line 54: Had a test pass all test suites, but then fail on this `afterAll` statement with a timeout exceeded 5000 ms. Updated these `afterAll`/`beforeAll` to have a timeout of 15 seconds as it is important that we clean up / create any policies even if it takes a bit of time for the http request to be completed

lipmas (approved):

PR #330, opened 2022-03-04
Made commands case insensitive [CWC-1460]

danimelchor:
## Description of the change
Converted camelCase commands to lowercase commands. Process args are split and the command is lower cased to allow for case-insensitive commands.

Example:
Previously, `zli targetuser [...]` would fail. Now, all variations of `targetuser` are valid (`targetUser`,`TargetUser`,`TARGETUSER`...)

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1460

- [ ] Yes
- [X] No

If yes, please explain: N/A


sidpremkumar:
Looking good 🥳 , one thing you'll need to do is update the kube system tests here: https://github.com/bastionzero/zli/blob/fix/condence-agent-public-key/src/system-tests/tests/suites/kube.ts#L152-L169

Since now you'll be using a different command `targetUser` -> `targetuser`

src/cli-driver.ts
```diff
@@ -208,9 +208,19 @@ export class CliDriver
...
     }
 
+    private parseArgvCaseInsensitive() {
+        // Converting commands to lowercase
+        const argv = process.argv.slice(2);
+        if(!argv[0]) return argv;
+
+        argv[0] = argv[0].toLowerCase();
+        return argv;
+    }
```
tflpd, line 218: Let's put this in utils

src/system-tests/tests/utils/zli-utils.ts
```diff
@@ -1,6 +1,11 @@
 import { CliDriver } from '../../../cli-driver';
 import * as CleanExitHandler from '../../../handlers/clean-exit.handler';
 
+const parseWithCaseInsensitive = (zliArgs: string[]) => {
+    zliArgs[0] = zliArgs[0].toLowerCase();
+    return zliArgs;
+};
+
```
tflpd, line 8: Let's make this a `function` to keep the existing pattern

tflpd (changes requested):

src/system-tests/tests/utils/zli-utils.ts
danimelchor: [b3571c6](https://github.com/bastionzero/zli/pull/330/commits/b3571c6383b496b52aebe7057981b64c4272ba08)

src/cli-driver.ts
danimelchor: [b3571c6](https://github.com/bastionzero/zli/pull/330/commits/b3571c6383b496b52aebe7057981b64c4272ba08)

sidpremkumar:
> Looking good 🥳 , one thing you'll need to do is update the kube system tests here: https://github.com/bastionzero/zli/blob/fix/condence-agent-public-key/src/system-tests/tests/suites/kube.ts#L152-L169
> 
> Since now you'll be using a different command `targetUser` -> `targetuser`

Hey @danimelchor @tflpd pointed out to me that this PR should have ensured that this change should have been needed (even though we still should update sys tests to reflect the lowercase name).

But yeah this change should have been unneeded for sys tests to pass. My b!

danimelchor:
> > Looking good 🥳 , one thing you'll need to do is update the kube system tests here: https://github.com/bastionzero/zli/blob/fix/condence-agent-public-key/src/system-tests/tests/suites/kube.ts#L152-L169
> > Since now you'll be using a different command `targetUser` -> `targetuser`
> 
> Hey @danimelchor @tflpd pointed out to me that this PR should have ensured that this change should have been needed (even though we still should update sys tests to reflect the lowercase name).
> 
> But yeah this change should have been unneeded for sys tests to pass. My b!

Yep! I realized as well. Just had to add transformation to lower case in the tests instead of changing the commands executed. However, I think it's good to have the tests use the same command that is specified in the documentation.

PR #331, opened 2022-03-04
Command lists changed from arrays to sets [CWC-1581]

danimelchor:
## Description of the change
Converted the command arrays `oauthCommands`, `mixpanelCommands`, `fetchCommands`, `adminOnlyCommands` to sets for slightly improved performance.

Previously, command lookup in the middlewares was done using `lodash` `includes` O(n) method. Now, command lookup is done in O(1) using JavaScript's built in set method `has`. This should slightly reduce the response time of the zli when running a command.

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1581

- [ ] Yes
- [X] No

If yes, please explain: N/A


src/cli-driver.ts
```diff
@@ -229,31 +228,31 @@ export class CliDriver
                 this.keySplittingService = initResponse.keySplittingService;
             })
             .middleware(async (argv) => {
-                if(!includes(this.oauthCommands, argv._[0]))
+                if(!this.oauthCommands.has(argv._[0] as string))
```
sidpremkumar, line 231: Instead of replicating this logic for `argv._[0] as string` lets just define this at the top of `getCliDriver`

So something like
```js
public getCliDriver(isSystemTest: boolean = false) {
         const argvPassed = process.argv.slice(2);
         const baseCommand: string = argvPassed[0];
...
```

And then we can change all refs to `args._[0]` to `baseCommand`

Believe that should work I just tested it locally with lt

danimelchor: I believe this would work normally, but not during testing. Since the tests are executing `.parseAsync(...)` on the return value of `getCliDriver(...)`, they would lose the access to the local variable `baseCommand`, right? The `argv` you mention is passed along the middlewares when yargs parses the command you want to run, but in the tests there is no real `process.argv` but instead we use a function call to yargs which processes the input into the `argv` variable that is passed along.

Maybe I can define a custom type for argv instead?

tflpd: Have not worked with the tests so I may be wrong - however if I am not mistaken every test calls the `callZli` for every command it wants to test. So every zli invocation is for a single command. We do not expect the lifecycle of the `getCliDriver` to extend over multiple runs so we shouldn't care if its access gets lost (?)
I may be wrong though

tflpd (approved):

src/cli-driver.ts
danimelchor: No, no. I didn't mean that `baseCommand` should extend during multiple runs. What I mean is that it isn't accessible even in one run.

There are two ways to run the cli using yargs. Either you initialize yargs with `yargs(process.argv.slice(2))` or you call the methods `.parse/.parseAsync/.parseSync` on the already initialized yargs.

However, these methods are executed differently in our code. Since the first method is done inside `getCliDriver` but the second one is done outside of `getCliDriver`, only the first method actively changes the value of `baseCmd` to the `process.argv` cmd value. The second method would keep using whatever the value `baseCmd` was when the cli was initialized to (`undefined` for tests).

I just made this little example on a different test project to try out my theory. If you want to, just init an npm project and install yargs and run it.

The results are the following:
```bash
baseCmd has value 'undefined' - $0 has value 'refresh'   # When ran with .parse(cmd)
baseCmd has value 'refresh' - $0 has value 'refresh'     # When ran using process.argv
```

Here is the code if you want to trace it.
```typescript
// What we do in cli-driver.ts
const yargs = require("yargs");

class CliDriver {
    private fakeCommands = new Set<string>(["refresh", "test", "hello"]);

    public start() {
        this.getCliDriver().argv;
    }

    public getCliDriver() {
        const argvParsed = process.argv.slice(2);
        const baseCmd: string = argvParsed[0];

        return yargs(argvParsed)
            .scriptName("zli")
            .usage("$0 <cmd> [args]")
            .middleware((argv: any) => {
                console.log(`baseCmd has value '${baseCmd}' - $0 has value '${argv._[0]}'`);
            })
            .help();
    }
}

// Config
const RUN_AS_TEST = true

// What we do in index.ts
// (reads process.argv, so run `ts-node index.js refresh`)
if (!RUN_AS_TEST){
    const driver = new CliDriver();
    driver.start()
}

// What we do in zli-utils.ts
// (simultes cmd using .parse, so run `ts-node index.js`)
if (RUN_AS_TEST) {
    const driver = new CliDriver();
    const cli = driver.getCliDriver()
    cli.parse("refresh")
}
```

tflpd: Huh.. Nice find, good job!

lipmas: > However, these methods are executed differently in our code. Since the first method is done inside getCliDriver but the second one is done outside of getCliDriver, only the first method actively changes the value of baseCmd to the process.argv cmd value. The second method would keep using whatever the value baseCmd was when the cli was initialized to (undefined for tests).

Just throwing this idea out...maybe having `getCliDriver` take an `argv: string[]` argument  that it can perform any pre-processing logic on before using it to initialize yargs would help here. That way the pre-processing logic could also be shared with system-tests (seems like this would be useful for case insensitive change you are making in #330).

```ts
public start() {
    // @ts-ignore TS2589
    this.getCliDriver(process.argv.slice(2))
        .argv; // returns argv of yargs
}
```
And then in system tests I was thinking like this:

```ts
const cliDriver = new CliDriver();
const callbackComplete = new Promise<void>(async (res, rej) => {
    try {
        await cliDriver.getCliDriver(zliArgs, true).parseAsync(zliArgs, {}, async (err, argv, output) => {
    ...
```

I took a quick look at `parseAsync` and it doesnt have an overload that allows you to pass a custom parseCallback function without also  including the args being passed (which would overwrite what yargs was previously initialized with in `getCliDriver`)... so maybe this doesnt really help.

danimelchor: Okay, I changed a few things on [89fb1f0](https://github.com/bastionzero/zli/pull/331/commits/89fb1f05f49de914dfcb0c1621fc4b52b6f06452) that should make this approach possible and cleaner. Now there is a single method that runs commands (not two different ways anymore). I will add the changes from my different open branch here:

```typescript
public run(argv?: string[], isSystemTest?: boolean, callback?: (err: Error, argv: any, output: string) => void) {
        // ----- CONVERT TO LOWER CASE HERE -----
        // @ts-ignore TS2589
        return this.getCliDriver(isSystemTest).parseAsync(argv || process.argv.slice(2), {}, callback);
    }
```

I also changed the `as string` thing that Sid didn't like to be in the cmd list (a bit cleaner and similar to Yargs documentation), but we can probably switch back to the `baseCmd` approach if we want to since its a single method for both ways now.

lipmas: Nice! Yeah this is so much cleaner now that both index.ts and system tests call the same yargs `parseAsync` function. 

> I also changed the as string thing that Sid didn't like to be in the cmd list (a bit cleaner and similar to Yargs documentation), but we can probably switch back to the baseCmd approach if we want to since its a single method for both ways now.

Not a huge deal but i guess my preference would be to switch back to the `baseCmd` approach sid suggested now that we can just so we dont have to  make our cmd lists `string | number` type when they will should always just be a `string` type.

danimelchor: Done: [7094e62](https://github.com/bastionzero/zli/pull/331/commits/7094e622e288b96c9e0915e6429966e217687181)

lipmas (approved):

PR #337, opened 2022-03-16
Fix/list policies

sidpremkumar:
## Description of the change

Catches the edge case where policies are created via the api / autogenerated and we can get null target / envs.

Release Notes: Catches null edge case for policy targets and envs

## Related JIRA tickets

Relates to JIRA: CWC-1615

- [ ] Yes
- [X] No

If yes, please explain:


src/utils/utils.ts
```diff
@@ -426,18 +426,20 @@ export function getTableOfKubernetesPolicies(
         let formattedTargetUsers = '';
         let formattedTargetGroup = '';
 
-        if (p.environments.length != 0) {
+        if (p.environments != undefined && p.environments.length != 0) {
```
lipmas, line 429: I think just as good practice we should change this check to `p.environments &&  p.environments.length != 0` so we handle the case where p.environments is null in addition to undefined too.

lipmas (approved):

PR #340, opened 2022-03-17
Fixed list-targets bug [CWC-1580]

danimelchor:
## Description of the change

Commands such as `list-targets` errored out when the user had no existing targets/policies/...

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1580

- [ ] Yes
- [X] No

If yes, please explain:


lipmas (approved):
No longer a bug because backend returns `allowedTargetsUsers` now. Still not a bad idea to keep this check in the code just in case this changes in the future.

PR #366, opened 2022-04-08
Feat/json silent or verbose

danimelchor:
## Description of the change

When using `--json` flag now makes the output silent by default. If the user passes in a `--verbose/-v` flag, the output is no longer silent.

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1508

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


tflpd (approved):
LGTM! Should we add `example`s that show the usage of verbose?
Not sure if super useful though

src/handlers/user/user.command-builder.ts
```diff
@@ -50,6 +50,15 @@ export function userCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<userArgs> {
                 alias: 'j',
             }
         )
+        .option(
+            'verbose',
```
sidpremkumar, line 54: I'm a bit confused on what the difference between `-v` and `--debug`? Could we reuse `--debug` in this case instead of creating another param?

danimelchor: Honestly, no clue. I didn't know about the debug flag. The only thing --verbose does is show the "Logged in as 123412341234' message that is shown after every command.

sidpremkumar: I will leave this decision up to product, but personally it feels like tech debt if we add a new flag that basically does the same thing right? Maybe there is some other difference that I am not aware of though.

tflpd (approved):

src/handlers/user/user.command-builder.ts
tflpd: Discussed this with product, `verbose` is to be used to provide more information to customers, `--debug` is to provide an extremely detailed trace of actions to a developer or a customer who wants to figure out what is going on. Their functionality is different, the one is product oriented the other engineering oriented @sidpremkumar

PR #370, opened 2022-04-11
Ensure that list-targets table is correctly formatted

sidpremkumar:
## Description of the change

There is an issue currently for web and db listing where the name of the type of target is less blocks then the header:
image

This PR fixes that: 
2022-04-12 at 12 43 AM

Release Notes: Ensure that filtering based on web+db targets allows the header for lt to remain correctly formatted

## Related JIRA tickets

Relates to JIRA: CWC-1712

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


tflpd (dismissed):

johncmerfeld (approved):

src/utils/utils.ts
```diff
@@ -177,7 +177,7 @@ export function getTableOfTargets(targets: TargetSummary[], envs: EnvironmentSum
     // The following constant numbers are set specifically to conform with the specified 80/132 cols term size - do not change
     const targetNameLength = max(targets.map(t => t.name.length)) + 2 || 16; // || 16 here means that when there are no targets default the length to 16
     const envNameLength = max(envs.map(e => e.name.length)) + 2 < 16 ? 16 : max(envs.map(e => e.name.length));
-    const targetTypeLength = max(targets.map(t => t.type.length)) + 2;
+    const targetTypeLength = max(targets.map(t => t.type.length)) + 2 < 6 ? 6 : max(targets.map(t => t.type.length));
```
lipmas, line 180: Not that familiar but why are we not doing something like:

```
max(max(targets.map(t => t.type.length)), 6)
```

This would guarantee its always at least 6 characters wide.

sidpremkumar: Totally right, I was just reusing the patten we defined for env, but I updated both now to use the syntax you mentioned: https://github.com/bastionzero/zli/pull/370/commits/a2845855aaec0cef1d860b23569565ec77683647

2022-04-12 at 12 58 AM

lipmas (approved):

PR #371, opened 2022-04-11
Add policy unit tests

sidpremkumar:
## Description of the change

Adds list policy unit test to test all types of filters + general list policies

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1691

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:


chad-cwc:
There's a lot going on here, and I'm sure we'll need to discuss further. My main thoughts right now are:
- Should unit tests live as close to the unit/module being tested as possible (e.g. in the same directory)? I think it's a good practice, but I haven't explored other options.
- Are these really unit tests? I see a lot of mocking, but they don't really look like unit tests. Looking at the code, I can see how we landed here. Have you discussed this with @ymarcus93, since he had the idea of "unit testing" list-targets?

sidpremkumar:
> Should unit tests live as close to the unit/module being tested as possible (e.g. in the same directory)? I think it's a good practice, but I haven't explored other options.

We can do that if you want. I wanted to keep the same structure of system-tests and I personally like this structure but not enough to fight for it. So if you want to move this logic to be closer to the handlers themselves im okay with that. 

> Are these really unit tests? I see a lot of mocking, but they don't really look like unit tests. Looking at the code, I can see how we landed here. Have you discussed this with @ymarcus93, since he had the idea of "unit testing" list-targets?

Yes, and after discussion and trial and error this was the easiest solution. The problem is we pass the `yargs.Argument` object into the handler itself, and that was very hard to mock. Really we should have an interface between the yargs parsing and the handler itself but we did not write the code that way. And short of a major refactor of the codebase that did not seem viable. 

The conclusion was this is fine for now, but not what we should be looking to do moving forward, instead of refactoring the entire codebase right now to work with what 'real unit tests' should be, we will go down the route of this janky mocking, and then in the future as we add new handlers + unit tests, we will add code in a way that makes it easier to mock out.

chad-cwc (approved):

PR #373, opened 2022-04-12
Add test for connect with only ssh access

sidpremkumar:
## Description of the change

If we have a policy with only ssh access, we should not be able to connect. This adds that test. 

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1685

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:


src/system-tests/tests/suites/ssh.ts
```diff
@@ -106,6 +108,31 @@ export const sshSuite = () => {
             }, 60 * 1000);
         });
 
+        ssmTestTargetsToRun.forEach(async (testTarget: TestTarget) => {
+            it(`${testTarget.sshCaseId}: connect test with only tunnel policy - ${testTarget.awsRegion} - ${testTarget.installType} - ${testTarget.dropletImage}`, async () => {
```
chad-cwc, line 112: Would it be helpful if this description says something like 'connect should fail with tunnel-only policy'?

sidpremkumar: https://github.com/bastionzero/zli/pull/373/commits/8d9f758207dce1751e217351936c9a70d244395c

chad-cwc: Sorry to be nitpicky here, but your new text "connect test should fail with only tunnel policy" reads like the test should fail. We know it's a test so the word test doesn't need to be in there anyway.

sidpremkumar: https://github.com/bastionzero/zli/pull/373/commits/c7dbdecc70ab89b32e48356a894f3c868a63f50b

chad-cwc (approved):

PR #381, opened 2022-04-15
Add system test for ssh-proxy-config

johncmerfeld:
## Description of the change

Doing right by the bugfix for CWC-1726 by adding a new system test to cover the ssh-proxy-config output and make sure it can be used to connect to a target

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-1726

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


src/system-tests/tests/suites/ssh.ts
```diff
@@ -50,6 +56,40 @@ export const sshSuite = () => {
...
+                id: configService.me().id,
+                type: SubjectType.User
+            };
+            const environment: Environment = {
+                id: systemTestEnvId
+            };
+
+            // create our policy
+            await policyService.AddTargetConnectPolicy({
+                name: systemTestPolicyTemplate.replace('$POLICY_TYPE', 'target-connect'),
```
sidpremkumar, line 70: name could be `tunnel-connect` so we don't conflict with regular connect: https://github.com/bastionzero/zli/blob/ssh-proxy-host/src/system-tests/tests/suites/connect.ts#L43

src/system-tests/tests/suites/ssh.ts
```diff
@@ -50,6 +56,40 @@ export const sshSuite = () => {
...
+            await callZli(['ssh-proxy-config']);
+
+            const sshProxyInstructions = loggerSpy.mock.calls[2][0];
+
+            // ensure the file was written correctly given the test context
+            expect(sshProxyInstructions).toContain('npm run start ssh-proxy');
+
+            fs.writeFileSync(staticConfigFile, sshProxyInstructions);
+
+            await cleanupTargetConnectPolicies(systemTestPolicyTemplate.replace('$POLICY_TYPE', 'target-connect'));
```
sidpremkumar, line 90: Do we have to delete it, if we are just going to create it in the next test?

src/system-tests/tests/suites/ssh.ts
```diff
@@ -89,13 +129,25 @@ export const sshSuite = () => {
...
-
             const pexec = promisify(exec);
-            const { stdout } = await pexec(command);
+
+            // first, test the auto-generated file (no user or prefix needed)
+            let command = `ssh -F ${userConfigFile} -o CheckHostIP=no -o StrictHostKeyChecking=no ${doTarget.ssmTarget.name} echo success`;
+            let { stdout } = await pexec(command);
+            expect(stdout.trim()).toEqual('success');
+
+            // next, test the ssh-proxy-config file
```
sidpremkumar, line 143: These should be two different tests.

johncmerfeld: The crazy thing is if I put them in separate tests, whichever one is second will fail -- but I can keep hacking on that

src/system-tests/tests/system-test.ts
```diff
@@ -38,7 +38,7 @@ const configName = envMap.configName;
```
sidpremkumar, line 38: Also could we revert this file? I made a bunch of changes to this file in the system-test branch, so rebasing will be a bit tricky. Not sure if I could find any changes 🤔

src/system-tests/tests/suites/ssh.ts
sidpremkumar: That's strange, what kind of failure do you get?

src/system-tests/tests/system-test.ts
johncmerfeld: Ah, whoops, definitely can do

src/system-tests/tests/suites/ssh.ts
sidpremkumar: Since we are getting permission denied, it sounds like it's getting to the agent and then the ssh protocol is rejecting us. 

So seems like not a policy issue / parsing issue. Might make sense to pull agent / zli logs. If you go down that path, just don't clean up (i.e. comment out) the cleanup for the digital ocean targets, but not sure how helpful that will be, hasn't been terribly helpful in the past.  

If you leave the targets up though you could also try connecting after the fact.

PR #386, opened 2022-04-21
Session recording test suite

chad-cwc:
## Description of the change

Added new HTTP service for Session Recording CRUD operations and added test suite to cover basic session recording functionality.

Release Notes: New HTTP service for Session Recording CRUD operations and new test suite to cover basic session recording functionality

## Related JIRA tickets

Relates to JIRA: CWC-1677

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


src/system-tests/tests/suites/session-recording.ts
```diff
@@ -0,0 +1,201 @@
...
+        });
+
+        test('3042: Connect to target and verify session is recorded', async () => {
+            const testTarget = ssmTestTargetsToRun[0];
+            const doTarget = testTargets.get(testTarget) as DigitalOceanSSMTarget;
+
+            // Spy on output pushed to stdout
+            const outputSpy = jest.spyOn(ShellUtils, 'pushToStdOut');
+            const connectPromise = callZli(['connect', `${targetUser}@${doTarget.ssmTarget.name}`]);
+            const message = 'testing 123';
```
sidpremkumar, line 116: So if we ever decide to test connect and session recording in parallel, this test might fail on: 
```
expect(testRecording.length).toEqual(1);
```

In order to fix that, lets just make this message unique, and if we find multiple session recordings, loop till we find our unique message. 

this way we are consistent with our assumption that each `describe` is isolated and will not conflict with another describe.

src/system-tests/tests/suites/session-recording.ts
```diff
@@ -0,0 +1,201 @@
...
+            const allRecordings = await sessionRecordingService.ListSessionRecordings();
+            const testRecording = allRecordings.filter(r => r.targetId === doTarget.ssmTarget.id);
+            // There should only be one recording that matches the doTarget's ID; however, if
+            // there is more than one recording returned by the filter, bail out so we don't
+            // end up deleting a recording that may be needed for something else.
+            expect(testRecording.length).toEqual(1);
+            testConnectionId = testRecording[0].connectionId;
+
+            // Get session recording and verify the echo'd message is in the asciicast data.
+            let messageFound = false;
```
sidpremkumar, line 157: This is unneeded, `includes` returns false: https://www.w3schools.com/jsref/jsref_includes.asp, we can just 
```
const messageFound = downloadedSessionRecording.includes(message);
```

src/system-tests/tests/suites/session-recording.ts
sidpremkumar: When we find the sessionRecording that matches this unique value, we can then save that connectionId and reuse it in the `Delete a session recording - should not delete because connection is open` test. 

And then the `Delete a session recording - should succeed because connection is closed` can just loop over all session recordings that match that targetId so we can cleanup any potential session recording that might have been created. 

I think there is an open question if we want to make these tests run in series though: 
```
if(SSM_ENABLED) {
    connectSuite();
    sshSuite();
}
```

But either way, even if its possible to leave dangling session recordings, I think the pros of ensuring that each `describe` in parallel is better than the off chance of leaving a dangling session recording.

src/system-tests/tests/suites/session-recording.ts
```diff
@@ -0,0 +1,201 @@
...
+            expect(allRecordings.find(r => r.connectionId === testConnectionId)).toBeDefined();
+        }, 15 * 1000);
+
+        test('3044: Delete a session recording - should not delete because connection is open', async () => {
+            let expectedError = undefined;
+            try {
+                await sessionRecordingService.DeleteSessionRecording(testConnectionId);
+            } catch (error) {
+                // The error for trying to delete an active session recording should be a SessionRecordingException.
+                if (error.errorType === 'SessionRecordingException') {
```
chad-cwc, line 174: This actually fails right now. The zli `HttpService.getHttpErrorMessage` returns only the error message string for this error. I could match on that string, but I'd prefer not to do that. Ideally, our back-end error handling would be improved and a code would be bubbled up so that could be checked here. For now, I'm going to treat any error as the expected error, which isn't great but will suffice for now.

src/system-tests/tests/suites/session-recording.ts
```diff
@@ -0,0 +1,205 @@
...
+            allTestConnections.forEach(connectionId => allDeleteSessionRecordingPromises.push(sessionRecordingService.DeleteSessionRecording(connectionId)));
+            try {
+                // Using allSettled so that each of these clean-up requests is attempted even if one fails.
+                await Promise.allSettled([
+                    policyService.DeleteTargetConnectPolicy(targetConnectPolicy.id),
+                    policyService.DeleteSessionRecordingPolicy(sessionRecordingPolicy.id),
+                    allDeleteSessionRecordingPromises
+                ]);
+            } catch (error) {
+                // catching and ignoring errors here so that test running can continue
```
sidpremkumar, line 107: Just curious, If we throw an error here, will the test fail?

src/system-tests/tests/suites/session-recording.ts
```diff
@@ -0,0 +1,205 @@
...
+        });
+
+        afterEach(() => {
+            if (mockStdin) {
+                mockStdin.restore();
+            }
+        });
+
+        ssmTestTargetsToRun.forEach(async (testTarget: TestTarget) => {
+            it(`3042: Connect to target ${testTarget.dropletImage} and verify session is recorded`, async () => {
```
sidpremkumar, line 124: This id should probably be unique per ssmTestTarget that is run. So one for AL2 with autodiscovery, with PM, etc.

sidpremkumar (dismissed):
Just minor nit regarding the testId, since we should be creating unique ones per ssm target we are testing. But LGTM

src/system-tests/tests/suites/session-recording.ts
chad-cwc: I believe so. I was convinced that any time an exception occurs and isn't handled, the test fails and no other tests run.

sidpremkumar: Ah okay so the test fails, I was wondering if the whole suite would exit prematurely. Just curious bc we thought saw an issue like that before and was wondering if you might have found something

src/system-tests/tests/suites/session-recording.ts
chad-cwc: 6ed7d86e182c398e6f4a2c1b4cf985e0da45b110

sidpremkumar (approved):

PR #407, opened 2022-04-27
Bzero target system tests

lipmas:
## Description of the change

Abstracts the connect system test suite to be able to test both bzero and ssm targets. Moved a lot of the code into a new `ConnectTestUtils` class that can be reused across different test suites that want to test basic connect functionality. For instance  i converted the group suite to reuse this same code as it was basically just duplicating the same connect test logic. Same is now true for the session recording tests and the close tests (part of connect test suite). I also now added zli attach tests to the connect suite. All of the test suites mentioned here now test bzero and ssm targets.

This also renames variables from `VT` => `Bzero` in a few places where it now makes sense to consider these targets as bzero targets instead of specifically vt targets. Added `BZERO_TARGETS_ENABLED` environment variable to enable/disable creating bzero target DO droplets.

Another change worth mentioning is that i added [extra provisioning logic for bzero DO droplets](https://github.com/bastionzero/zli/blob/58ebaee1416dce57888d334cce31a9bec064d7c4/src/system-tests/tests/system-test-setup.ts#L455) to create a new `bzuser` on the target so that we can use this user in our connect tests. This creates the user and sets their default shell to bash to ensure that command extraction works properly now that this has been added to the connect tests.

Finally I also fixed the `process.exit()` bugs that we would see from time to time in system tests which would cause the tests to abruptly exit without cleaning up. The issue was that we were mocking the `cleanExit` function inside of the `callZli` function and also clearing all mocks after each test. This would lead to a race condition where one test that called `callZli` but then completed before the zli exited (usually because of the jest test timeout) was still running at the same time as a new test was being started and cleared the cleanExit mock. Thus if the old test's zli exited at this time the cleanExit function was not mocked and would call the real process.exit(). Fixed this issue by setting up the cleanExit mock in the global jest  `beforeEach` setup method right after clearing all existing mocks. Since it clears and then sets up the cleanExit mock in the same function there is never a chance that the cleanExit function is unmocked during system tests. See changes in [1801514](https://github.com/bastionzero/zli/pull/407/commits/1801514bcebe7c81f2380753b0e1d5ead74dbda8) and [9411250](https://github.com/bastionzero/zli/pull/407/commits/9411250c83b4a3fcede774a74119f60ae6c223b8)

## TODO:

- [x] Add test rail IDs for the new bzero target  tests (also attach tests)

## Testing

**backend branch:** 
**bzero branch:** 

- [x] Yes

Release Notes: Added Bzero Target shell system tests

## Related JIRA tickets

Relates to JIRA: CWC-1650

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


src/system-tests/tests/system-test.ts
```diff
@@ -131,10 +132,10 @@ if(SSM_ENABLED) {
     logger.info(`Skipping adding ssm targets because SSM_ENABLED is false`);
 }
 
-if(VT_ENABLED) {
-    allTargets = allTargets.concat(vtTestTargetsToRun);
+if(BZERO_ENABLED) {
```
ymarcus93, line 135: Env var changes, like these ones, require a change in the python wrapper. This is because we expose test suites as options in the python command.

Code to update:
1) https://github.com/bastionzero/cwc-infra/blob/3fbbbe37c62e41e616012f4610080d3148921f26/Bzero-QA/bzero_qa/main.py#L71-L83
2) https://github.com/bastionzero/cwc-infra/blob/3fbbbe37c62e41e616012f4610080d3148921f26/Bzero-QA/bzero_qa/main.py#L57

src/system-tests/tests/suites/connect.ts
```diff
@@ -77,104 +77,26 @@ export const connectSuite = () => {
...
                 const expectedErrorMessage = 'Expected error';
                 jest.spyOn(CleanExitHandler, 'cleanExit').mockImplementationOnce(() => {
                     throw new Error(expectedErrorMessage);
                 });
-                const connectPromise = callZli(['connect', `baduser@${doTarget.ssmTarget.name}`]);
 
                 await expect(connectPromise).rejects.toThrow(expectedErrorMessage);
-
-                // Assert shell connection auth details has not been called
-                expect(shellConnectionAuthDetailsSpy).not.toHaveBeenCalled();
```
ymarcus93, line 177: Was removing this intended?

src/system-tests/tests/system-test.ts
lipmas: Hmm im wondering if it still makes sense to have `BZERO_ENABLED` and `SSM_ENABLED` if they dont correspond directly 1-1 with suites which seems to be the intended usage if the infra test command.

The semantics of `BZERO_ENABLED` and `SSM_ENABLED` was not tied to a specific suite but instead whether or not to create DO droplets of that type. Because the connect tests (and presumably other test suites as well) now test every target in [allTargets](https://github.com/bastionzero/zli/blob/4cdceea086497b825d0178ad0184b8b85c6ebf30/src/system-tests/tests/system-test.ts#L127)  which concatenates bzero/ssm targets.

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -273,5 +274,44 @@ export class TestUtils {
...
+     * @param expectationFunc The function to retry until it doesnt throw an
+     * error
+     * @param timeout A global timeout that will reject this promise with the
+     * last known error from the expectation function as soon as the timeout is
+     * reached. Doesnt wait for the  final expectation function to finish before
+     * rejecting.
+     * @param retryInterval Time to wait in-between invocations of
+     * expectationFunc
+     */
+    public async waitForExpect(expectationFunc: () => Promise<void>, timeout: number = 30 * 1000, retryInterval: number = 1 * 1000) {
```
ymarcus93, line 290: We could use generics in case we want to return a value from the `expectationFunc`, so `Promise<T>`. But I guess in general we probably won't be returning anything, just doing `expect` instead.

ymarcus93, line 290: Nice! 🎉🎉🎉🎉

src/system-tests/tests/system-test.ts
lipmas: https://github.com/bastionzero/cwc-infra/pull/434

src/system-tests/tests/utils/test-utils.ts
lipmas: [d1194d3](https://github.com/bastionzero/zli/pull/407/commits/d1194d312a570be4d5e3a1236cc1e32672a727b7)

src/system-tests/tests/system-test.ts
ymarcus93: Looks like I missed a reference. I commented on the infra PR

src/system-tests/tests/suites/connect.ts
ymarcus93: I think this is fine. I'm not sure if there is any other way. I guess the best scenario would be to capture the WSS url returned and assert the region is right, perhaps by exposing a `WhereAmI()` SignalR method similar to the HTTP REST one. But not sure how this would work with the new daemon.

src/system-tests/tests/suites/kube.ts
```diff
@@ -36,8 +36,6 @@ export const kubeSuite = () => {
         });
 
         beforeEach(() => {
```
sidpremkumar, line 38: Can we just remove `beforeEach` in that case?

src/system-tests/tests/suites/list-targets.ts
```diff
@@ -8,8 +8,6 @@ import { TargetType } from '../../../../webshell-common-ts/http/v2/target/types/
 export const listTargetsSuite = () => {
     describe('list targets suite', () => {
         beforeEach(() => {
```
sidpremkumar, line 10: Similarly here, we can just remove `beforeEach` right?

src/system-tests/tests/suites/ssh.ts
```diff
@@ -48,12 +48,6 @@ export const sshSuite = () => {
 
         // Called before each case
         beforeEach(() => {
```
sidpremkumar, line 50: similar, i think we can remove this now

src/utils/shell-utils.ts
```diff
@@ -207,21 +204,9 @@ export async function startShellDaemon(
...
-            };
-
-            const daemonProcess = await spawn(finalDaemonPath, args, options);
-
-            daemonProcess.on('close', (exitCode) => {
-                logger.debug(`Shell Daemon close event with exit code ${exitCode}`);
-                resolve(exitCode);
-            });
-
+            const daemonProcessExitCode = await spawnDaemon(finalDaemonPath, args, cwd);
```
sidpremkumar, line 207: Its expected now `detached` will be set to `true` since that is was is defined in `spawnDaemon`?

src/system-tests/tests/utils/connect-utils.ts
```diff
@@ -0,0 +1,313 @@
...
+        if(doTarget.type === 'bzero') {
+
+            let daemonPty: pty.IPty;
+            const capturedOutput: string[] = [];
+
+            jest.spyOn(ShellUtilWrappers, 'spawnDaemon').mockImplementation((finalDaemonPath, args, cwd) => {
+                return new Promise((resolve, reject) => {
+                    try {
+                        daemonPty = this.spawnDaemonPty(finalDaemonPath, args, cwd);
+                        daemonPty.onData((data) => capturedOutput.push(data));
```
sidpremkumar, line 175:
```
daemonPty.onData((data: string) => capturedOutput.push(data));
daemonPty.onExit((e: { exitCode: number | PromiseLike<number>; }) => resolve(e.exitCode));
```
maybe

sidpremkumar (approved):
Minor nits, but LGTM overall. Hopefully this will also fix our pipeline issue, thanks for catching that.

sidpremkumar:
We should add this: 
https://github.com/bastionzero/zli/blob/shell/system-tests/src/system-tests/tests/suites/kube.ts#L42-L43

To any test that uses the daemon , in order to clean the logs for each test. If all the shell tests pass, then fails on a kube test. The logs will include everything that went into that shell tests.

src/utils/shell-utils.ts
lipmas: yes this is expected its fine to run the shell daemon in detached mode which will make it run under a separate process group. The zli process still waits for the daemon process to exit before exiting itself though.

lipmas:
> We should add this: https://github.com/bastionzero/zli/blob/shell/system-tests/src/system-tests/tests/suites/kube.ts#L42-L43
> 
> To any test that uses the daemon , in order to clean the logs for each test. If all the shell tests pass, then fails on a kube test. The logs will include everything that went into that shell tests.

Added in [03186f2](https://github.com/bastionzero/zli/pull/407/commits/03186f28cc99e1ba9108ce4613300f73252f3a62)

src/system-tests/tests/suites/kube.ts
lipmas: [23f4a4a](https://github.com/bastionzero/zli/pull/407/commits/23f4a4ad2f1dc91fe44e35e0da98c3dccbbe155b)

lipmas:
Also added test rail case IDs in [af9e1a9](https://github.com/bastionzero/zli/pull/407/commits/af9e1a9a9e5f3c961e123dd8f1a164981e6a6d08)

sidpremkumar (dismissed):

sidpremkumar:
Just need to lint 🚢

ymarcus93 (approved):

PR #411, opened 2022-04-28
MFA REST API test suite

chad-cwc:
## Description of the change

Add MFA REST API test suite.

## Testing

**backend branch:** 
**bzero branch:** 

- [x] Yes

Release Notes: MFA REST API test suite

## Related JIRA tickets

Relates to JIRA: CWC-1674

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


ymarcus93 (commented):
This looks good. However, I think we will run into issues with this suite if we run all suites in parallel.

Imagine the mfa suite is running and hits the enable mfa test or reset mfa test. Then, if the next test that is run is not part of the MFA suite, it will error saying "MFA is required, please login in".

So let's keep this in mind and possibly just disable running MFA suite in parallel with other suites.

Also, it would be nice to assert in the enable MFA suite and reset MFA suite that after doing enable/reset, the next command issued on zli should fail with an MFA error--because they haven't verified their token.

I also think a test that tries connect or something after enabling MFA _and_ providing the OTP token would be good to test that MFA is truly working e2e. To write this latter test, we can call the verify OTP token endpoint manually with a token. You'll have to use an OTP library, something like https://github.com/yeojz/otplib, so you can programmatically get a valid OTP token using the MFA secret returned from the API.

One could argue that providing the OTP token manually via hitting the API, and not via the respective zli command, breaks the e2e aspect of the test--however, I still think the test is better than not having it. We could do it via zli, but only if we add a way to pass the OTP token without having to go through entire login process interactively. Currently, token logic is bundled with login command.

Also, it's still technically e2e imho--as it does call the same API zli calls--the test just won't be testing exactly the codepath the zli makes (I.e API call originating from a cli handler). But if we want to test the command handler more closely for correctness, we can write a unit test for it.

chad-cwc:
> This looks good. However, I think we will run into issues with this suite if we run all suites in parallel.
> 
> Imagine the mfa suite is running and hits the enable mfa test or reset mfa test. Then, if the next test that is run is not part of the MFA suite, it will error saying "MFA is required, please login in".
> 
> So let's keep this in mind and possibly just disable running MFA suite in parallel with other suites.
> 
> Also, it would be nice to assert in the enable MFA suite and reset MFA suite that after doing enable/reset, the next command issued on zli should fail with an MFA error--because they haven't verified their token.
> 
> I also think a test that tries connect or something after enabling MFA _and_ providing the OTP token would be good to test that MFA is truly working e2e. To write this latter test, we can call the verify OTP token endpoint manually with a token. You'll have to use an OTP library, something like https://github.com/yeojz/otplib, so you can programmatically get a valid OTP token using the MFA secret returned from the API.
> 
> One could argue that providing the OTP token manually via hitting the API, and not via the respective zli command, breaks the e2e aspect of the test--however, I still think the test is better than not having it. We could do it via zli, but only if we add a way to pass the OTP token without having to go through entire login process interactively. Currently, token logic is bundled with login command.
> 
> Also, it's still technically e2e imho--as it does call the same API zli calls--the test just won't be testing exactly the codepath the zli makes (I.e API call originating from a cli handler). But if we want to test the command handler more closely for correctness, we can write a unit test for it.

Some very valid ideas here. However, we should discuss offline why I'm not seeing a way to make these changes. (Since it's related to authentication, I don't want to get into the details here.)

PR #419, opened 2022-05-04
Bzero target REST API test suite

chad-cwc:
## Description of the change

Bzero target REST API test suite

## Testing

**backend branch:** 
**bzero branch:** 

- [ ] Yes

Release Notes: Bzero target REST API test suite

## Related JIRA tickets

Relates to JIRA: CWC-1704

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:


src/system-tests/tests/suites/rest-api/bzero-targets.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        test('6431: Get a Bzero target by ID', async () => {
+            const doTarget = testTargets.get(vtTestTargetsToRun[0]) as DigitalOceanBZeroTarget;
+            const bzeroTarget = await bzeroTargetService.GetBzeroTarget(doTarget.bzeroTarget.id);
+            expect(bzeroTarget.id).toEqual(doTarget.bzeroTarget.id);
+        }, 15 * 1000);
+
+        test('6432: Get all Bzero targets', async () => {
+            const bzeroTargets = await bzeroTargetService.ListBzeroTargets();
+            expect(bzeroTargets.length).toBe(vtTestTargetsToRun.length);
```
chad-cwc, line 22: This isn't going to work during CI jobs because the tests run against a BastionZero instance that anyone could be adding targets to. Could change to either:
```javascript
expect(bzeroTargets.length).toBeGreaerThanOrEqual(vtTestTargetsToRun.length);
```
or something like:
```javascript
const doTarget = testTargets.get(vtTestTargetsToRun[0]) as DigitalOceanBZeroTarget;
expect(bzeroTargets.find(t => t.id === doTarget.bzeroTarget.id).toBeDefined();
```

sidpremkumar: I like `expect(bzeroTargets.length).toBeGreaerThanOrEqual(vtTestTargetsToRun.length);` personally, since we are already testing the 2nd option in `Get a Bzero target by Id`

src/system-tests/tests/suites/rest-api/bzero-targets.ts
```diff
@@ -0,0 +1,39 @@
...
+        test('6433: Edit a Bzero target', async () => {
+            const doTarget = testTargets.get(vtTestTargetsToRun[0]) as DigitalOceanBZeroTarget;
+            const bzeroTarget = await bzeroTargetService.GetBzeroTarget(doTarget.bzeroTarget.id);
+            const updatedTarget = await bzeroTargetService.EditBzeroTarget(bzeroTarget.id, {
+                targetName: 'testName1234'
+            });
+            expect(updatedTarget.name).toEqual('testName1234');
+
+            // set the name back to original name in case another test relies on the name for some reason
+            await bzeroTargetService.EditBzeroTarget(bzeroTarget.id, {
```
sidpremkumar, line 34: Not blocking, but something interesting to note is if we run the tests in parallel now, its possible that a test could fail because the name has been changed. 

I don't think we should spend effort trying to correct this, but for this tests and the ssm test, maybe a comment under `system-tests.ts` where the suite it called alerting regarding that is sufficient for now. 

The simple solution is, if we decide to run them in parallel, the ssm + bzero test has to be done in the same thread as the connect + vt tests, thus I think a comment is good enough for now.

ymarcus93: Oh yeah that's a good point. When we make the suites parallel, this will break. If there was something we could edit that doesn't affect connecting, e.g. description field of targets (which I don't think we have), then we could just edit one of those unimportant fields.

I guess the cleanest way (but would add a lot of time to system tests due to droplet creation) would be to create a target that is solely used for this ssm suite and separately the bzero REST suite. Then you could edit what ever you want because the target is reserved for the test.

ymarcus93: Or we could change the connect tests to connect by ID and then changing name won't matter.

One could argue that we now have lost coverage on connecting via name, but the same argument applies to us not having a test for connecting via unique target ID in the first place.

Both aspects can be resolved by picking one to use in the system tests, and then just making a simple unit test in zli that checks when you pass the name or the id in zli connect, with a mocked Bastion API that returns a list of targets matching the name+id, zli connect correctly continues to the start shell method and doesn't exit.

Small details/edge cases like these feel best to do in unit tests.

chad-cwc: I'm not sure what to consider in scope for this PR. The only other property that can be edited on Bzero and SSM targets is `envronmentId`, but I assume that has the same possibility of corrupting other tests since a policy could be based on environment and changing the environment of a target could cause a connect to fail.

I could add the comment as @sidpremkumar suggested and add to that comment that using IDs in connect tests might allow parallelization.

ymarcus93: Sounds good. Can you also make the suffix change for the name done in the SSM suite?

chad-cwc: Yes. I was planning on that as well as the change on the list endpoint to check what the array contains.

chad-cwc: 1618f54ddfdbd386ae4fdc64483a6a1127ad9ed8

ymarcus93 (approved):

sidpremkumar (approved):

PR #435, opened 2022-05-12
WIP: bzero ssh

johncmerfeld:
## Description of the change

## Testing

**backend branch:** 
**bzero branch:** 

- [ ] Yes

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:


src/handlers/ssh-proxy/ssh-proxy.handler.ts
```diff
@@ -61,8 +66,116 @@ export async function sshProxyHandler(configService: ConfigService, logger: Logg
...
+            detached: false,
+            shell: true,
+            stdio: 'pipe'
+        };
+
+        const daemonProcess = spawn(finalDaemonPath, args, options);
+
+        configService.logoutDetected.subscribe(() => {
+            logger.error(`\nLogged out by another zli instance. Terminating ssh tunnel\n`);
+            daemonProcess.stdin.end();
```
lipmas, line 135: This triggers the daemon process to exit? Why not just send sigterm with `.kill()` instead? I guess this must be working otherwise we wouldnt ever hit the `daemonProcess.on('close'` but still seems safer to just kill the process.

src/system-tests/tests/utils/ssh-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (doTarget.type === 'ssm') {
+        user = ssmUser;
+        target = doTarget as DigitalOceanSSMTarget;
+        targetName = target.ssmTarget.name;
+    } else {
+        user = bzeroTargetCustomUser;
+        target = doTarget as DigitalOceanBZeroTarget;
+        targetName = target.bzeroTarget.name;
+    }
+    return { user, targetName };
```
lipmas, line 24: nit: can we define this return type as a new typescript interface in this file just so its a bit easier to extend in the future.

johncmerfeld: 3d2f868

src/handlers/ssh-proxy/ssh-proxy.handler.ts
johncmerfeld: Yeah, I guess it's not a very explicit thing on this side but the daemon is watching for EOF from stdin, and sends an explicit close message to the agent when it receives it. It seemed a little more safe / graceful than killing it outright, but maybe that is the safer move, since the agent will figure out that the daemon died anyway

lipmas: Ok I see yeah then this is probably ok for now since it does seem better to gracefully kill the agent connection as well. We could catch the sigterm signal from the daemon and still send the close message but not sure we have any code to do that right now and might be more work so we can leave that as a future enhancement.

lipmas (approved):
LGTM. Although i would still update the submodule once more to point at develop once the bzero PR is merged

PR #441, opened 2022-05-17
Add Support for Container Agents

sidpremkumar:
## Description of the change

This adds system test support for our container agents. Along with example docker files (that are used to build the images)

## Testing

There is automatic testing in this PR, but to manually test you can build the docker image: 
```
docker buildx build --platform linux/amd64 -t test-ecs:latest -f {dockerfile} . 
```

to one of the docker files, then run with: 
```
docker run  -e REGISTRATION_KEY="ABC" -e TARGET_NAME="test-target" -e SERVICE_URL="https://sid.bastionzero.com/" test-ecs:latest -e SERVICE_URL=https://sid.bastionzero.com/
```

And you should be able to register and connect to that target. 

**backend branch:** 
**bzero branch:** 

- [x] Yes

Release Notes: Add tests + examples for agents inside a container

## Related JIRA tickets

Relates to JIRA: CWC-1883

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:


src/system-tests/tests/utils/AgentContainerDockerfiles/al2.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Before -> 
+# ENTRYPOINT [ "python3", "-m", "http.server" ]
+
+# After -> 
+# Params:
+ARG REGISTRATION_KEY
+# Optional
+ARG TARGET_NAME
+# ARG SERVICE_URL=https://cloud.bastionzero.com/
```
sidpremkumar, line 21: TODO: Switch this before merging

src/system-tests/tests/suites/agent-container.ts
```diff
@@ -0,0 +1,101 @@
...
+                const sessionRecordingTestMessage = `session recording test - ${systemTestUniqueId}`;
+                const connectionId = await connectTestUtils.runShellConnectTest(testTarget, sessionRecordingTestMessage, true);
+
+                // Get session recording and verify the echo'd message is in the asciicast data.
+                const downloadedSessionRecording = await sessionRecordingService.GetSessionRecording(connectionId);
+                const messageFound = downloadedSessionRecording.includes(sessionRecordingTestMessage);
+                expect(messageFound).toEqual(true);
+
+                testPassed = true;
+            }, 2 * 60 * 1000);
```
lipmas, line 98: Any reason to do a session recording test as opposed to just a normal connect test?

src/system-tests/tests/system-test-setup.ts
```diff
@@ -44,11 +44,133 @@ export async function setupSystemTestApiKeys() {
...
+        // Get the config file
+        const cluster = await doKubeService.getDigitalOceanClusterById(systemTestDigitalOceanClusterId);
+        const kubeConfigFileContents = await doKubeService.getClusterKubeConfig(cluster);
+
+        // Init Kube client
+        const kc = new k8s.KubeConfig();
+        kc.loadFromString(kubeConfigFileContents);
+        const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+
+        const targetName = `agent-container-${target.type}-${resourceNamePrefix}`;
```
lipmas, line 77: nit: prefix should come first

src/system-tests/tests/utils/connect-utils.ts
```diff
@@ -157,8 +157,8 @@ export class ConnectTestUtils {
      * ssm target into a common interface ConnectTarget that can be used in
      * system-tests
      */
-    public getConnectTarget(doTarget: DigitalOceanSSMTarget | DigitalOceanBZeroTarget) : ConnectTarget {
-        if(doTarget.type === 'bzero') {
+    public getConnectTarget(doTarget: DigitalOceanSSMTarget | DigitalOceanBZeroTarget | DigitalOceanBZeroContainerTarget ) : ConnectTarget {
```
lipmas, line 160: Just noting i did some similar refactoring in #442 to allow for different target types but split the common parts into a new function `getBzeroConnectTarget`

src/system-tests/tests/system-test-setup.ts
```diff
@@ -44,11 +44,133 @@ export async function setupSystemTestApiKeys() {
...
+            bzeroTarget: undefined,
+            podInfo: pod,
+        };
+
+        try {
+            const containerAgentTarget = await doService.pollBZeroTargetOnline(targetName);
+            target.awsRegion = containerAgentTarget.region;
+            bzeroContainerTestTarget.bzeroTarget = containerAgentTarget;
+
+            testTargets.set(target, bzeroContainerTestTarget);
```
lipmas, line 138: Maybe we shouldnt be putting these container targets in the `testTargets` map. They are different enough because they use the do kube cluster instead of creating new droplets that i feel like we shouldnt be coupling them with the other DO target setup code. The main advantage to doing this is if we wanted to include this in `allTargets` and have the connect/session-recording/groups etc test suites also test against them. But we seem to just be skipping them in all of those test suites anyway and have a dedicated test suite for container targets.

So why not just create/destroy the container test target specifically within the agent-container suite setup/teardown? Then you dont have to worry about it conforming to `TestTarget` interface. Seems like this would simplify a lot of code and reduce the scope creep where we now have to handle the `pm-container` special case everywhere.

src/system-tests/tests/utils/AgentContainerDockerfiles/ubuntu.Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ARG REGISTRATION_KEY
+# Optional
+ARG TARGET_NAME
+ARG SERVICE_URL=https://cloud.bastionzero.com/
+ARG ENVIRONMENT_NAME
+
+# New entrypoint:
+RUN curl "${SERVICE_URL}api/v2/autodiscovery-scripts/container" -o entrypoint.sh 
+RUN chmod +x ./entrypoint.sh
+ENTRYPOINT [ "./entrypoint.sh", "python3", "-m", "http.server" ]
```
lipmas, line 29: Probably missing something...whats the point of running the python webserver instead of just running the agent?

lipmas (commented):
Tested the container autodiscovery and its working:

image

image

src/system-tests/tests/suites/connect.ts
```diff
@@ -72,6 +72,11 @@ export const connectSuite = () => {
         });
 
         allTargets.forEach(async (testTarget: TestTarget) => {
+            // We want to skip any pm-pod test target
+            if (testTarget.installType === 'pm-pod') {
+                return;
+            }
```
lipmas, line 78: If we have to handle this in every test suite i think it defeats the purpose putting container targets in `allTargets`

src/system-tests/tests/utils/AgentContainerDockerfiles/al2.Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM amazonlinux:2
+
+# Install normal dependencies 
+RUN yum install python3 wget bash -y 
+
+# BastionZero Deps
+RUN yum install curl procps -y
```
lipmas, line 7: should these not be contained in the autodiscovery entrypoint if they are our dependencies?

src/system-tests/tests/system-test.ts
```diff
@@ -209,6 +219,14 @@ beforeAll(async () => {
                 return;
             }
             testCluster = await setupDOTestCluster();
+        }(),
+        async function() {
+            // Skip container setup unless we are running in pipeline, TODO
+            if (false) {
+                logger.info(`Skipping agent-container tests because IN_CLUSTER is false`);
```
lipmas, line 226: Need to check IN_PIPELINE from TODO. But side question- why not run this in CI too? We already have DO cluster created so it seems like this test wouldnt add too much extra overhead?

src/system-tests/tests/utils/AgentContainerDockerfiles/ubuntu.Dockerfile
sidpremkumar: Because this is not a container that is a target, this is a container that is just normal (i.e. our bastion) and we are side loading our agent onto that container. So that way in ECS a customer could start a server normal as they would in their dockerfile (i.e. `python3 -m http.server`) but still load our agent into that container. 

This way the container logs will still show the output of the server, and the lifecycle of the container still follows the server. But our agent is also on that same box.

src/system-tests/tests/utils/AgentContainerDockerfiles/al2.Dockerfile
sidpremkumar: This are the dependencies to run the script, i.e. we can't curl the script if we don't have curl. And `prcps` is used for the spinners.

src/system-tests/tests/suites/agent-container.ts
sidpremkumar: Just resolving this bc it's making my editer act really weird for some reason

sidpremkumar:
test

lipmas (approved):

PR #458, opened 2022-06-01
Universal Connect

lipmas:
## Description of the change

This moves most the zli connect logic to the backend (see https://github.com/bastionzero/webshell-backend/pull/1140) into a single API call (universal connect) in order to speedup shell (and other types of targets) connect time in the zli and also does the same for zli ssh-proxy command.

Additionally the universal connect API creates and returns the connection resource and connection service auth details so that these values can be passed directly to the daemon and the daemon can connect to the connection-node websocket right away.

Also removed `connect` and `ssh-proxy` commands from the list of fetch commands (where we make API calls to all target APIs) and removed `disambiguateTarget` util as this functionality has been moved to the backend where its much faster! 

Related PRs:

backend: https://github.com/bastionzero/webshell-backend/pull/1140
bzero: https://github.com/bastionzero/bzero/pull/95
common-ts: https://github.com/bastionzero/webshell-common-ts/pull/76

## Testing

Make sure you are on the related feature branch in backend (also requires db migration). Connect to bzero (shell), db, web, kube targets or ssh to bzero targets and everything should be working the same as before only faster. Additionally exiting on shell and ssh should also be working much faster now.

**backend branch:** feat/universal-connect
**bzero branch:** feat/universal-connect

- [x] Yes

Release Notes: Universal connect API updates + Shell/SSH connection optimizations

## Related JIRA tickets

Relates to JIRA: CWC-1889

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain:

See https://github.com/bastionzero/bzero/pull/95 PR


src/handlers/ssh-proxy/ssh-proxy.handler.ts
```diff
@@ -1,51 +1,94 @@
...
+    const configName = configService.getConfigName();
+    if(configName != 'prod') {
+        prefix = `${configName}-${prefix}`;
+    }
 
-    if (!ssmTarget.allowedVerbs.map(v => v.type).includes(VerbType.Tunnel)) {
-        logger.error('You do not have sufficient permission to open a ssh tunnel to the target');
-        await cleanExit(1, logger);
+    if(! argv.host.startsWith(prefix)) {
+        this.logger.error(`Invalid host provided must have form ${prefix}<target>. Target must be either target id or name`);
```
johncmerfeld, line 35: nit but there should probably be a colon after `Invalid host provided`

src/handlers/attach/attach.handler.ts
```diff
@@ -49,9 +49,10 @@ export async function attachHandler(
 
         const bzeroTargetService = new BzeroTargetHttpService(configService, logger);
         const bzeroTargetRequest = bzeroTargetService.GetBzeroTarget(connectionSummary.targetId);
+        const connectionAuthRequest = connectionHttpService.GetShellConnectionAuthDetails(connectionSummary.id);
 
         // Make requests in parallel
-        const [bzeroTarget, attachInfo] = await Promise.all([bzeroTargetRequest, attachInfoRequest]);
+        const [bzeroTarget, attachInfo, connectionAuthDetails] = await Promise.all([bzeroTargetRequest, attachInfoRequest, connectionAuthRequest]);
 
         // TODO: Adjust this version check once pipelining changes are in and we support attaching
```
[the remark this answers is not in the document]
lipmas: Alas no, the comment was a bit optimistic in thinking attaching would just work after pipelining changes got in but in reality there are still a few more things we need to fix before attach is working for bzero targets. So we have a ticket to add attach support + add detach functionality at the same time- https://commonwealthcrypto.atlassian.net/browse/CWC-1826.

I should update this comment now and link the ticket though.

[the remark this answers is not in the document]
lipmas: [f137c5c](https://github.com/bastionzero/zli/pull/458/commits/f137c5c47a1ab45400dc210dd680111bf257f3df)

src/handlers/ssh-proxy/ssh-proxy.handler.ts
lipmas: [7abf426](https://github.com/bastionzero/zli/pull/458/commits/7abf426ddde087f132f2c4494327a49c19ae1145)

johncmerfeld (approved):

src/handlers/connect/shell-connect.handler.ts
```diff
@@ -107,26 +40,32 @@ export async function shellConnectHandler(
...
             // Finally once the dat is created and the connection is open get
             // the updated bzero target details so we can connect
-            bzeroTarget = await bzeroTargetService.GetBzeroTarget(connectionSummary.targetId);
+            const bzeroTargetService = new BzeroTargetHttpService(configService, logger);
+            const bzeroTarget = await bzeroTargetService.GetBzeroTarget(connectionSummary.targetId);
+            agentPublicKey = bzeroTarget.agentPublicKey;
+            agentVersionString = bzeroTarget.agentVersion;
+
+            // Get auth details for DATs to pass daemon
+            authDetails = await connectionHttpService.GetShellConnectionAuthDetails(connectionSummary.id);
```
ymarcus93, line 57: What is `authDetails` set to when receiving the universal connect API response for a DAT?

https://github.com/bastionzero/zli/blob/f722d94b06369f8aeb6384d5eb7ebfd88e218516/src/handlers/connect/shell-connect.handler.ts#L26

I'm assuming it's unknown until DAT is created and that's why we have to make antother API call to Bastion to get the shell auth details for DAT?

ymarcus93: I found this comment in the tests that I think explains it: https://github.com/bastionzero/zli/blob/f722d94b06369f8aeb6384d5eb7ebfd88e218516/src/system-tests/tests/utils/connect-utils.ts#L83-L89

Maybe we should also write a similar comment in `shell-connect.handler.ts` as well to avoid confusion

ymarcus93 (approved):

src/handlers/connect/shell-connect.handler.ts
lipmas: > What is authDetails set to when receiving the universal connect API response for a DAT? I'm assuming it's unknown until DAT is created and that's why we have to make antother API call to Bastion to get the shell auth details for DAT?

Correct yeah its just set to null in the API. We can only get auth details once the target is registered and we therefore know which AWS region to use for the connection service. Sure i can also add a comment to  `shell-connect.handler.ts`

lipmas: [575ba02](https://github.com/bastionzero/zli/pull/458/commits/575ba02476479d8b0ce3952633d59c9530ff7a70)

PR #476, opened 2022-07-07
Validate BZCert in Daemon and handle errors in zli

lipmas:
## Description of the change

This PR handles errors bubbled up from the daemon in https://github.com/bastionzero/bzero/pull/109 when checking the BZCert during a `zli connect`. If the daemon exits with a custom error code we now handle that exit by logging a user facing message. This adds a single custom exit code for errors with bzcert validation so it can bubble up the error to the user and prompt them to login again. This works in any `zli connect` command across all target types.

It also adds methods in `OrganizationService` to call the new `bzcert-validation-info` from  https://github.com/bastionzero/webshell-backend/pull/1173 and adds a middleware to ensure this data is always populated in the zli config file so it can be loaded and used by the daemon.

TODO:

- [x] Add API test for new `bzcert-validation-info` endpoint 

## Testing

1. Modify your zli config file json (the one returned by `zli configure`) to have an error in the `keySplitting.initialIdToken` field. Simply changing the first character of this token should be enough to trigger the error since the initialIdToken will no longer validate when building the `BZCert` on the daemon side.

2. The following error message should appear whenever you try and run a `zli connect` or ssh to any target (includes web, db, kube as well as bzero and ssm targets) 
```
Error constructing BastionZero certificate: IdP tokens are invalid/expired. Please try logging in again with 'zli login' to resolve this issue.
```

**backend branch:** develop
**bzero branch:** develop

- [x] Yes

Release Notes: Populates bzcert validation info in config file for daemon. Handles specific exits codes when there is an error with bzcert validation and bubbles up the error to the user and prompts them to login again.

## Related JIRA tickets

Relates to JIRA: CWC-1861

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


johncmerfeld:
When I check out these branches and replace a character in `keySplitting.initialIdToken`, my connections don't work but I don't see the above error message. They mostly fail silently (or with `Exit code 2` in the case of shell). I am able to fix my config with a `zli login`

src/utils/daemon-utils.ts
```diff
@@ -20,6 +23,39 @@ export const DAEMON_PATH : string = 'bzero/bctl/daemon/daemon';
...
+ * Starts the daemon as a background process with stdio ignored. If the daemon
+ * process exits with an error the zli process will log any custom error
+ * messages and then exit.
+ * @param logger Logger service to use for logging custom errors and clean exit
+ * @param cwd current working directory to start daemon process in
+ * @param daemonPath path to daemon executable
+ * @param args daemon command line args
+ * @returns The spawned child process
+ */
+export async function spawnDaemonInBackground(logger: Logger, cwd: string, daemonPath: string, args: string[]): Promise<cp.ChildProcess> {
```
johncmerfeld, line 37: Love this

src/services/web/web.service.ts
```diff
@@ -2,7 +2,7 @@ export interface WebConfig {
     name: string,
     localHost: string,
     localPort: number,
-    localPid: null,
+    localPid: number,
```
johncmerfeld, line 5: Weird, why was this `null` before?

src/utils/daemon-exit-codes.ts
```diff
@@ -0,0 +1,4 @@
+export const enum DAEMON_EXIT_CODES {
+    UNSPECIFIED_ERROR     = 1,
```
johncmerfeld, line 2: I know this is a bit of a bikeshed, but does it make sense for these to go here? I worry about a proliferation of files with constants, where nobody quite knows where to put new constants.

I don't know if the answer is to have a more centralized constants file, but I might at least put them inside daemon-utils?

src/services/web/web.service.ts
lipmas: Yeah i think it was just a typo/bug in our typescript interface.

src/utils/shell-util-wrappers.ts
```diff
@@ -21,7 +24,7 @@ export function pushToStdErr(output: Uint8Array) {
  * @param cwd current working directory to use for the spawned subprocess
```
johncmerfeld, line 24: nit but the logger should be in the @param list

src/utils/daemon-exit-codes.ts
lipmas: I thought maybe it was better to add exit codes to a separate file so it would be very easy to navigate to that file when looking up an error code. Totally could just go in `daemon-utils` though. I was more thinking of a future where we had a lot more error codes and not just 2 but maybe we should cross that bridge later.

lipmas:
> When I check out these branches and replace a character in `keySplitting.initialIdToken`, my connections don't work but I don't see the above error message. They mostly fail silently (or with `Exit code 2` in the case of shell). I am able to fix my config with a `zli login`

Hmm im not able to reproduce this. Its especially weird because its actually exiting with code 2 for shell then it should be hitting this error log: https://github.com/bastionzero/zli/blob/d52480a69bd3575d81552e5d172398dc5d39c867/src/utils/daemon-utils.ts#L402-L404. You dont have the silent (-s) flag on for some reason do you?

src/utils/shell-util-wrappers.ts
lipmas: [645a60b](https://github.com/bastionzero/zli/pull/476/commits/645a60bbf0b70098e819a4ecc8ebb0d1a17a85d9)

src/utils/daemon-utils.ts
lipmas: [645a60b](https://github.com/bastionzero/zli/pull/476/commits/645a60bbf0b70098e819a4ecc8ebb0d1a17a85d9)

src/utils/daemon-exit-codes.ts
johncmerfeld: Sounds like either option has merits; I say follow your heart

johncmerfeld:
As I mentioned over Slack, I think [this behavior](https://stackoverflow.com/questions/55731760/go-os-exit2-show-a-bash-value-of-1) was the cause of my problem. If I unset `ZLI_CUSTOM_DAEMON_PATH`, the daemon was run as a pre-built executable instead of by `go run`, so we're all good.

However it's probably worth noting this behavior somewhere, perhaps in daemon-exit-codes.ts or `waitForDaemonProcessExit`, because other developers will probably encounter this at some point

johncmerfeld (approved):

src/handlers/middleware.handler.ts
```diff
@@ -162,3 +163,13 @@ export async function initMiddleware(argv: any, logger : Logger, isSystemTest :
         keySplittingService: keySplittingService
     };
 }
+
+export async function bzCertValidationInfoMiddleware(keySplittingService: KeySplittingService, configService: ConfigService, logger: Logger) {
+    const ksConfig = configService.loadKeySplitting();
+    if( ! ksConfig.orgProvider) {
+        // Update the Org BZCert Validation parameters
+        const orgHttpService = new OrganizationHttpService(configService, logger);
+        const orgBZCertValidationInfo = await orgHttpService.GetUserOrganizationBZCertValidationInfo();
```
chad-cwc, line 172: I realize the probability is low, but what happens if this HTTP request gets an error response? Is that handled gracefully up the stack somewhere? I poked around a bit but didn't find the answer yet.

src/handlers/ssh-proxy/ssh-proxy.handler.ts
```diff
@@ -185,13 +185,8 @@ async function bzeroOpaueSshProxyHandler(configService: ConfigService, logger: L
             daemonProcess.stdin.end();
         });
 
-        daemonProcess.on('close', async (exitCode) => {
-            if (exitCode !== 0) {
-                logger.error(`ssh daemon close event with nonzero exit code ${exitCode} -- for more details, see ${loggerConfigService.daemonLogPath()}`);
```
chad-cwc, line 190: I like the `waitForDaemonProcessExit` change, but we lose the indication that this is the "ssh daemon" since the error message is now generic for any daemon. Could that make debug/support more difficult or is there other context that can be used to know we're dealing with the ssh daemon?

src/handlers/middleware.handler.ts
lipmas: It is caught and handled in the http service code [here](https://github.com/bastionzero/zli/blob/dbe19cc9b3317ff2a0ce6fe619a79b36b409bbc3/src/services/http/http.service.ts#L137-L140) but then the error is re-thrown and will get caught at the top level of the cli driver [here](https://github.com/bastionzero/zli/blob/92b9ad55c1c47e1b291978ea80cfe08d346982db/src/cli-driver.ts#L729-L754). So in this case if the request fails it will just cause the zli command to fail.

src/handlers/ssh-proxy/ssh-proxy.handler.ts
lipmas: I dont think its an issue that we lose the information about it being an ssh daemon since i think it should be pretty obvious from the context. But now that you mention it i do think the log message here with the exit code/daemon log path is pretty nice for debug:

```
logger.error(`ssh daemon close event with nonzero exit code ${exitCode} -- for more details, see ${loggerConfigService.daemonLogPath()}`);
```
so i can try and add that in the waitForDaemonProcessExit

chad-cwc (approved):

lgmugnier (approved):
tested

PR #477, opened 2022-07-08
Distinguish targets by environment

SaiDadireddy:
## Description of the change

This feature allows users to specify the environment, of the target they are connecting, to in the target string. This will allow users to distinguish between targets with the same name but in different environments. 

Examples of the zli connect command including the following:
`zli connect target-user@target-name.environment-name`
`zli connect target-user@target-name.environment-id`
`zli connect target-name.environment-id`

This PR covers the following:
1. Updating `parseTargetString` to extract environment id/name from the target string
2. Adds examples of target strings with the environment to the command builder. The examples are visible when the user runs the command `zli connect --help`
4. Adding unit tests to cover the environment variable extraction in `parseTargetString`
5. Adding end to end tests for `connect.handler`
6. Adds the .environment feature to connect system tests
7. Adds environment Id and environment Name checks in the logs for system tests

## Testing

End to end tests for the connect handler:
 `npm test connect.handler`

Unit tests for `parseTargetString`:
`npm test utils`

**backend branch:** distinguish-targets-by-env 
**bzero branch:** develop

- [X] Yes

Release Notes: 

## Related JIRA tickets

Relates to JIRA: CWC-1943

Relevant HotFix PR Number: 

- [ ] Yes
- [X] No

If yes, please explain:


src/utils/utils.ts
```diff
@@ -166,13 +167,30 @@ export function parseTargetString(targetString: string) : ParsedTargetString
 
     // extract targetId and maybe targetPath
     const colonSplit = atSignSplit[0].split(':', 2);
-    const targetSomething = colonSplit[0];
+    const targetAndEnvironment = colonSplit[0];
```
lipmas, line 170: We can remove the colonSplit logic know because its just making this function more confusing. This is legacy from when we use to have a copy command where you could specify paths. But `parseTarget.path` has 0 references so we can just remove that from the interface.

src/utils/utils.ts
```diff
@@ -166,13 +167,30 @@ export function parseTargetString(targetString: string) : ParsedTargetString
...
+    // extract the environment id or name
+    const firstSeparatorIndex = targetAndEnvironment.indexOf('.');
+    let targetIdOrName = '';
+    let environmentIdOrName = '';
+    if(firstSeparatorIndex >= 0) {
+        targetIdOrName = targetAndEnvironment.slice(0, firstSeparatorIndex);
+        environmentIdOrName = targetAndEnvironment.slice(firstSeparatorIndex + 1);
+    } else {
+        targetIdOrName = targetAndEnvironment;
+    }
```
lipmas, line 181: why not use split instead to be consistent with how we parse the target user

```
const dotSplit = targetAndEnvironment.split('.')
const targetIdOrName = dotSplit[0];
const environmentIdOrNAme = dotSplit[1];
```

src/utils/utils.ts
```diff
@@ -139,6 +139,7 @@ export function parseTargetStatus(targetStatus: string) : TargetStatus {
 export function parseTargetString(targetString: string) : ParsedTargetString
 {
     // case sensitive check for [targetUser@]<targetId | targetName>[:targetPath]
+    // not sure if i need to modify this
```
lipmas, line 142: Yeah so i think the regex still matches because it treats the environment as part of the target name capture group. But tbh i dont think this regex is doing much for us anyway so id be in favor of just removing this code. cc @tflpd wdyt?

src/utils/utils.ts
SaiDadireddy: Checked the docs and there's no examples with the :path, so removing this logic, removing the `path` field from the `ParsedTargetString` and updating the unit tests.

src/utils/utils.ts
lipmas: Ah i see your point. That made me thing why isnt there a bug with how we extract the user then because we use `split` there and there could be an `@` in the target name as well. But i think the current code actually works because of the limit [option on split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split):

```
let atSignSplit = targetString.split('@', 2);
``` 

So  that code guarantees only 2 substrings will be returned and any additional `@` will be ignored.

src/handlers/connect/connect.handler.test.ts
```diff
@@ -0,0 +1,151 @@
...
+            // Assert that GetShellConnectionAuthDetails was not called
+            // This should only be called for DAT targets
+            expect(getShellAuthDetails).not.toHaveBeenCalled();
+
+            // Assert that startShellDaemon was called
+            // We are simulating that an error occured and we hit reject(1)
+            expect(startShellDaemonSpy).toHaveBeenCalled();
+        });
+
+        // skipping this test becuase it is passing locally but not in the PR
```
lipmas, line 87: Do you have an idea why this test is failing in the CI? Id rather we either just delete the test if its not needed or fix it so it works in CI as well instead of leaving the code unused.

SaiDadireddy: I wasn't able to figure out why since it's just a unit test and isolated to that suite. I'll delete it - the functionality being tested is covered by system tests anyways.

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -53,6 +53,8 @@ export class TestUtils {
             targetName: targetName,
             targetUser: targetUser,
             timestamp: expect.anything(),
+            environmentId: targetEnvId,
+            environmentName: expect.anything(),
```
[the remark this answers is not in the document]
SaiDadireddy: Thanks! Good idea, I'll add this in before merging.

lipmas (approved):
LGTM

src/utils/utils.ts
tflpd: Sorry just saw this :(
Uhm, I mean, there was probably a reason it was there in the first place..?
Not sure, I just don't like touching things that work, especially when there is no negative impact of them existing (?)

PR #481, opened 2022-07-11
Fix base command detection [CWC-1983]

tflpd:
## Description of the change

This fixes a bug in the way we detect the base command during a zli call. Functionality in our middleware is based on detecting this correctly thus this could lead to unexpected behaviors.

## Testing

Execute zli commands placing flags and arguments first and the command in the end or in the middle of the arguments. Try that using different combinations of capitalization in the command string.

**backend branch:** 
**bzero branch:** 

- [x] Yes

Release Notes: Fix base command detection

## Related JIRA tickets

Relates to JIRA: CWC-1983

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


src/utils/utils.ts
```diff
@@ -29,14 +29,34 @@ import { Group } from '../../webshell-common-ts/http/v2/policy/types/group.types
...
+        baseCmd: '',
+        parsedArgv: []
+    };
+    caseInsensitiveArgv.parsedArgv = [];
+    argv.forEach(arg => {
+        if(zliCommands.has(arg.toLowerCase())) {
+            const baseCmd = arg.toLowerCase();
+            caseInsensitiveArgv.baseCmd = baseCmd;
+            caseInsensitiveArgv.parsedArgv.push(baseCmd);
+        }
```
lipmas, line 53: Probably should find the first arg that maches a command and make that the `baseCmd` instead of doing this which will find the last arg that matches. Assuming some degenerate edge cases where target names are the same as a zli command.

tflpd: Yes and no. I see your point and will make the adjustment but keep in mind that the issue would still apply if the target name argument comes first and the command second. Also if for some reason an argument matches exactly an existing zli command yargs will fail anyway

tflpd: [d6736a1](https://github.com/bastionzero/zli/pull/481/commits/d6736a181203aa4a0ddda1d513133159ab045440)

src/utils/utils.ts
```diff
@@ -910,17 +931,17 @@ export function removeIfExists(file: string): void {
...
-    return {
-        baseCmd: argv[0],
-        parsedArgv: argv
-    };
-}
+//     return {
+//         baseCmd: argv[0],
+//         parsedArgv: argv
+//     };
+// }
```
lipmas, line 944: remove commented out code

src/cli-driver.ts
```diff
@@ -215,6 +246,12 @@ export class CliDriver
                 this.configService = initResponse.configService;
                 this.keySplittingService = initResponse.keySplittingService;
             })
+            .middleware(async (argv) => {
```
lipmas, line 249: Love this middleware. Maybe this is causing the `zli --version` sys test failure though because in that case maybe one of `baseCmd` or `argv._` is undefined?

tflpd: I tried logging argv entering that middleware and the first in order middleware, nothing gets printed, I believe `--version` is treated in some other way, it is not even getting parsed

lipmas (approved):
LGTM

PR #494, opened 2022-07-21
Pass runtime config to the daemon via environment variables instead of flags

johncmerfeld:
## Description of the change

ZLI changes corresponding to [this daemon change](https://github.com/bastionzero/bzero/pull/119). Instead of passing runtime config as a set of flags, we create key/value pairs and pass them via `spawn`'s [env option](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options), along with the current `process.env`

## Testing

No functionality change, so the key is that it behaves normally

**backend branch:** develop
**bzero branch:**develop

- [x] Yes

Release Notes:  Remove sensitive information from the command that spawns the daemon

## Related JIRA tickets

Relates to JIRA: CWC-1992

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

It has the positive impact of making it harder to retrieve any secret values passed by the ZLI to the daemon!


lgmugnier (approved):
tested with shell

PR #500, opened 2022-08-05
Add `create` functionality to `zli policy`

SaiDadireddy:
## Description of the change

This PR adds `create` functionality to the `zli policy` command. 
`zli policy create-recording` -> creates a new session recording policy
`zli policy create-proxy` -> creates a new proxy policy
`zli policy create-tconnect` -> creates a new target connect policy
`zli policy create-cluster` -> creates a new kubernetes cluster policy

## Testing

**Manual tests:**
You can download two bash scripts I wrote in our `Engineering` shared drive:
[Create Policy Test Scripts](https://drive.google.com/drive/folders/1tyLKJrAr9tavKn2Yns9XXNwlZbDCATj7?usp=sharing)

These scripts should be run against **staging** where everything is already set up for them. You should be using your local instance of zli for this. In both scripts, `success-test-cases.sh` and  `error-test-cases.sh`, I made clear comments about what each test case is aiming for. If there's any cases I'm missing that you think would be beneficial to see please let me know.

**System tests:**
These commands are covered by the `policySuite()` test suite in the system tests.

**backend branch:** 
**bzero branch:** develop

- [x] Yes

Release Notes: Users can now create a policy from the zli

## Related JIRA tickets

Relates to JIRA: CWC-1541

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/handlers/policy/policy-create/create-cluster-policy.command-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+            }
+        )
+        .option('users',
+            {
+                type: 'string',
+                array: true,
+                demandOption: true,
+                requiresArg: true,
+                alias: 'u',
+                description: 'BastionZero the this policy applies to (idpEmails)'
```
asamborski, line 31: typo here I think - did you mean to say BastionZero users that this policy applies to?

src/handlers/policy/policy-create/create-cluster-policy.command-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function createClusterPolicyCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<createClusterPolicyArgs> {
+    return yargs
+        .option('name',
+            {
+                type: 'string',
+                demandOption: true,
+                requiresArg: true,
+                alias: 'n',
+                description: 'Name of the new policy. If spaces included, wrap in double quotes.'
```
asamborski, line 21: Do we want to put the word 'new' or would name of the policy be enough?

src/handlers/policy/policy-create/create-cluster-policy.command-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+            {
+                type: 'string',
+                default: null,
+                demandOption: false,
+                requiresArg: true,
+                alias: 'd',
+                description: 'Policy description. Wrap this sentence in double quotes.'
+            }
+        )
+        .example('$0 policy create-cluster -n policy_name -u user@random.com -c test_cluster --targetUsers ec2-user', 'Create a new kube cluster policy with the specified args');
```
asamborski, line 94: I'd modify to say 'create a new cluster policy' instead of kube cluster to match the verbiage of the policy type

src/handlers/policy/policy-create/create-proxy-policy.command-builder.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function createProxyPolicyCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<createProxyPolicyArgs> {
+    return yargs
+        .option('name',
+            {
+                type: 'string',
+                demandOption: true,
+                requiresArg: true,
+                alias: 'n',
+                description: 'Name of the new policy. If spaces included, wrap in double quotes.'
```
asamborski, line 19: Same feedback on the word 'new' - up to you to decide. Let's just stay consistent

src/handlers/policy/policy-create/create-proxy-policy.command-builder.ts
```diff
@@ -0,0 +1,75 @@
...
+            }
+        )
+        .option('users',
+            {
+                type: 'string',
+                array: true,
+                demandOption: true,
+                requiresArg: true,
+                alias: 'u',
+                description: 'BastionZero users this policy applies to (idpEmails)'
```
asamborski, line 29: I think this is what you meant to have up above. Can we also amend idpEmails to SSO emails ?

src/handlers/policy/policy-create/create-proxy-policy.command-builder.ts
```diff
@@ -0,0 +1,75 @@
...
+            }
+        )
+        .option('targets',
+            {
+                type: 'string',
+                array: true,
+                demandOption: false,
+                requiresArg: true,
+                alias: 't',
+                description: 'Targets that this policy applies to (db or web targets)'
```
asamborski, line 50: Where are we surfacing error messages? If a user puts a target that isn't a db or web target, for example, what error will the user receive back?

src/handlers/policy/policy-create/create-recording-policy.command-builder.ts
```diff
@@ -0,0 +1,63 @@
...
+            }
+        )
+        .option('users',
+            {
+                type: 'string',
+                array: true,
+                demandOption: true,
+                requiresArg: true,
+                alias: 'u',
+                description: 'BastionZero users the policy applies to (idpEmails)'
```
asamborski, line 28: Same feedback - change idpEmails to SSO emails

src/handlers/policy/policy-create/create-recording-policy.command-builder.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function createRecordingPolicyCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<createRecordingPolicyArgs> {
+    return yargs
+        .option('name',
+            {
+                type: 'string',
+                demandOption: true,
+                requiresArg: true,
+                alias: 'n',
+                description: 'Name of the new policy. If including spaces, wrap in double quotes.'
```
asamborski, line 18: same feedback here about the word "new"

src/handlers/policy/policy-create/create-recording-policy.command-builder.ts
```diff
@@ -0,0 +1,63 @@
...
+            }
+        )
+        .option('recordInput',
+            {
+                type: 'boolean',
+                default: false,
+                demandOption: false,
+                requiresArg: true,
+                alias: 'r',
+                description: 'Indicates whether input should be recorded. Anything but "true" will be false.'
```
asamborski, line 49: Does this message mean you can leave it blank and it'll default to false?

src/handlers/policy/policy-create/create-tconnect-policy.command-builder.ts
```diff
@@ -0,0 +1,108 @@
...
+            }
+        )
+        .option('users',
+            {
+                type: 'string',
+                array: true,
+                demandOption: true,
+                requiresArg: true,
+                alias: 'u',
+                description: 'BastionZero users the policy applies to (idpEmails)'
```
asamborski, line 32: idpEmails -> SSO emails

src/handlers/policy/policy-create/create-tconnect-policy.command-builder.ts
```diff
@@ -0,0 +1,108 @@
...
+            }
+        )
+        .option('apikeys',
+            {
+                type: 'string',
+                array: true,
+                demandOption: false,
+                requiresArg: true,
+                alias: 'a',
+                description: 'SSO API Keys the policy applies to'
```
asamborski, line 42: This field actually doesn't do anything - I think it's worth considering whether or not we should include it in this new command. I *believe* it's in the backlog to remove that field all together from the policy. Maybe @lipmas could confirm this? (Not sure who would know definitively)

src/handlers/policy/policy-create/create-tconnect-policy.command-builder.ts
```diff
@@ -0,0 +1,108 @@
...
+            }
+        )
+        .option('targets',
+            {
+                type: 'string',
+                array: true,
+                demandOption: false,
+                requiresArg: true,
+                alias: 't',
+                description: 'Targets the policy applies to (ssm, bzero, and dac targets)'
```
asamborski, line 63: Do you mean dynamic instead of dac?

src/handlers/policy/policy-create/create-cluster-policy.command-builder.ts
SaiDadireddy: I think name of the policy would be enough actually, using `create` command should be enough to imply it's a new policy.

src/handlers/policy/policy-create/create-proxy-policy.command-builder.ts
SaiDadireddy: Yup! Changing `idpEmails` to `SSO emails`

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,233 @@
...
+        }
+        // If the user exists
+        const user: Subject = {
+            id: userSummary.id,
+            type: SubjectType.User
+        };
+        users.push(user);
+    }
+    return users;
+}
```
lipmas, line 42: I think its going to just be more efficient in cases where there are multiple users passed to just get all users using the `userHttpService.ListUsers();` like we do in the list-users command https://github.com/bastionzero/zli/blob/a4d710613d761497e19928a8f51eb383514a494d/src/handlers/policy/policy-user/list-users.handler.ts#L15 and then filter the users by email using this local set (you could even put the users into a map/object indexed by the email so its very efficient to check if a user email exists or not). So i think the `getUsersByEmail` should just be refactored to work this way even in the case of a single user.

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,233 @@
...
+            // If the environment exists
+            const environment: Environment = {
+                id: resourceSummary.id,
+            };
+            resource = environment;
+        }
+        resources.push(resource);
+    }
+    return resources;
+}
```
lipmas, line 86: ah great i see you have already done this for searching for environments/groups. But yeah from my other comment i would suggest just building a map(plain js object also should do) of the resources that is indexed by the name/email (whatever is being searched on).

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,233 @@
...
+
+export async function getBaseDacTargetsByNameOrId(targetNames: string[], baseTargetHttpService: BzeroTargetHttpService, dynamicAccessHttpService: DynamicAccessConfigHttpService, logger: Logger) {
+    // For each target, if it exists, grab the BzeroAgentSummary or DynamicAccessConfigSummary and create a new Target object
+    // id = BzeroAgentSummary.id | DynamicAccessConfigSummary.id, type = TargetType.Bzero | TargetType.DynamicAccessConfig
+    const allBaseTargets = await baseTargetHttpService.ListBzeroTargets();
+    const allDacTargets = await dynamicAccessHttpService.ListDynamicAccessConfigs();
+    return parseTwoTargetsTypesByNameOrId(targetNames, allBaseTargets, allDacTargets, false, logger);
+}
+
+async function parseTwoTargetsTypesByNameOrId(targetNames: string[], allTargetsType1: any[], allTargetsType2: any[], isVirtual: boolean, logger: Logger) {
```
lipmas, line 164: For this i would suggest reusing the same function `parseSingleResourceTypeByNameOrId` but instead reduce all of the targets to a common interface (can use the existing `Target` interface). Then we just need to handle the case where a single target name maps to multiple targets. I think in this case we should just throw an error that says "more than 1 target exists with name <name> please specify the target using its uuid instead".

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,233 @@
...
+
+export async function getVirtualTargetsByNameOrId(targetNames: string[], dbHttpService: DbTargetService, webHttpService: WebTargetService, logger: Logger) {
+    // For each target, if it exists, grab the DbTargetSummary or WebTargetSummary and create a new Target object
+    // id = DbTargetSummary.id | WebTargetSummary.id, type = TargetType.Db | TargetType.Web
+    const allDbTargets = await dbHttpService.ListDbTargets();
+    const allWebTargets = await webHttpService.ListWebTargets();
+    return parseTwoTargetsTypesByNameOrId(targetNames, allDbTargets, allWebTargets, true, logger);;
+}
+
+export async function getBaseDacTargetsByNameOrId(targetNames: string[], baseTargetHttpService: BzeroTargetHttpService, dynamicAccessHttpService: DynamicAccessConfigHttpService, logger: Logger) {
```
lipmas, line 156: This could just build a dictionary of name -> `Target[]` that we then pass to the helper function.

```ts
async function getAllTargetConnectTargets(targetNames: string[], baseTargetHttpService: BzeroTargetHttpService, dynamicAccessHttpService: DynamicAccessConfigHttpService, logger: Logger) {
    const allBaseTargets = await baseTargetHttpService.ListBzeroTargets();
    const allDacTargets = await dynamicAccessHttpService.ListDynamicAccessConfigs();

    let allTargetsByName: {[name: string] : Target[]} = {};

    allBaseTargets.forEach(bt => {
        // Create a common Target interface for this Bzero target
        let target: Target = {
            id: bt.id,
            type: TargetType.Bzero
        };

        if(! allTargetsByName[bt.name]) {
            allTargetsByName[bt.name] = [target];
        } else {
            allTargetsByName[bt.name].push(target);
        }
    });

    // Do same for DacTargets
    ...
}
```

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,233 @@
...
+    return parseTwoTargetsTypesByNameOrId(targetNames, allBaseTargets, allDacTargets, false, logger);
+}
+
+async function parseTwoTargetsTypesByNameOrId(targetNames: string[], allTargetsType1: any[], allTargetsType2: any[], isVirtual: boolean, logger: Logger) {
+    // Type 1 and 2 refer to the types of the targets this function is parsing
+    // Type 1 and 2 will always be Db and Web, or Bzero and DynamicAccessConfig, respectively
+    let errorMessageTarget = 'base target or dynamic access config';
+    if(isVirtual) {
+        errorMessageTarget = 'virtual target';
+    }
```
lipmas, line 170: Instead of handling this error message at this level i think it would be cleaner if you instead just had this function return an empty array of targets and then check at a higher layer if the targets is empty and if so create a custom error message there.

src/handlers/policy/policy-create/create-proxy-policy.command-builder.ts
lipmas: I think right now we would fail to find the web/db target and then an error log would be surfaced to the user that says:

"Unable to find vritual target with name: <name>"

https://github.com/bastionzero/zli/blob/79c967655bf50903adb4fd4b1a4bc3fc21c036b2/src/utils/policy-utils.ts#L230

src/handlers/policy/policy-create/create-tconnect-policy.command-builder.ts
lipmas: :+1:  Yup totally agreed i think we should remove the ability entirely to add ApiKeys as subjects in `TargetConnect` policies. ApiKeys would not actually be able to connect to targets given how the Mr.Zap protocol requires a valid SSO id token. Once we have service accounts those types of users will be able to but that we can address with the service account PRs.  Right now its still possible to add api keys to policies  from the webapp but should be removed at some point (hopefully there is a ticket for that somewhere but if not we should create one).

src/handlers/policy/policy-create/create-tconnect-policy.handler.ts
```diff
@@ -0,0 +1,102 @@
...
+        case 'shell':
+            verbType = VerbType.Shell;
+            break;
+        case 'filetransfer':
+            verbType = VerbType.FileTransfer;
+            break;
+        case 'tunnel':
+            verbType = VerbType.Tunnel;
+            break;
+        }
```
lipmas, line 80: Lets at least move this to a reusable helper function `parseVerbType` that parses a a verb type from string. We could also do something similar that i added for the `targetType` which allows us to map the enum to display string values so that we have more control over how these enum values are actually displayed to the user.

https://github.com/bastionzero/zli/blob/952b89f5a6b34a50aaf7beb362ff067560d2970d/src/utils/utils.ts#L63-L83

https://github.com/bastionzero/zli/blob/952b89f5a6b34a50aaf7beb362ff067560d2970d/src/utils/utils.ts#L201-L219

https://github.com/bastionzero/zli/blob/6529e675355f42a42d4c38c886db74331183db0a/src/cli-driver.ts#L190

src/cli-driver.ts
```diff
@@ -172,26 +179,10 @@ export class CliDriver
...
-
-    private fetchCommands: Set<string> = new Set([
-        'user',
-        'targetuser',
-        'targetgroup',
-        'describe-cluster-policy',
-        'generate',
-        'policy',
-        'group',
     ]);
```
lipmas, line 182: All of this https://github.com/bastionzero/zli/blob/f436832df270bb519109ef1bca9ed888dc482ad2/src/handlers/middleware.handler.ts#L23-L101 code can now also be removed since its no longer referenced.

src/handlers/policy/policy-create/create-cluster-policy.command-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+            }
+        )
+        .option('users',
+            {
+                type: 'string',
+                array: true,
+                demandOption: true,
+                requiresArg: true,
+                alias: 'u',
+                description: 'BastionZero users the this policy applies to (SSO emails)'
```
asamborski, line 31: @SaiDadireddy small nit: extra "the" in here

src/handlers/policy/policy-create/create-proxy-policy.command-builder.ts
asamborski: Got it. Is there a way to easily incorporate a more precise error message here? If not, I do not think this is a blocking issue

lipmas: I think saying "no virtual target with name found" is as precise an error message we can give without also querying for other types of targets which i dont think we want to do for performance reasons.

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+export async function getEnvironmentByName(environmentNames: string[], envHttpService: EnvironmentHttpService, logger: Logger) {
+    // For each environment, if it exists, grab the EnvironmentSummary and create a new Environment object
+    // id = EnvironmentSummary.id
+    const allEnvironments = await envHttpService.ListEnvironments();
+    return await parseUniqueResource(environmentNames, allEnvironments, 'environment', logger);
+}
+
+async function parseUniqueResource(resourceNames: string[], allResources: any, type: string, logger: Logger) {
```
lipmas, line 41: Instead of making the resource type `any` you can use generics so that you have better type safety within this function and as the return type of this function.

```ts
async function parseUniqueResource<T>(resourceNames: string[], allResources: T[], type: string, logger: Logger): Promise<T[]> { ... }
```

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,257 @@
...
+                id: summary.id,
+            };
+            resource = environment;
+        }
+        if(type === 'user') {
+            allResourcesMap[summary.email] = resource;
+        } else {
+            allResourcesMap[summary.name] = resource;
+        }
+    });
```
lipmas, line 70: One trick you could use here to simplify this logic and make it more extendable in the future is to have this function take a lambda function that maps from the resource type `T` to a string key value:

 ```ts
 async function parseUniqueResource<T>(resourceNames: string[], allResources: T[], type: string, toResourceKey: (res: T) => string, logger: Logger): Promise<T[]> {
    allResources.forEach((summary: T) => {
        allResourcesMap[toResourceKey(summary)] = summary;
    });
    ...
}
 ```
 
 Then example usage for users:
 
 ```ts
 return await parseUniqueResource(emails, allUsers, 'user', (u: UserSummary) => u.email, logger);
 ```

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+async function parseMultipleTargetTypes(
+    targetNames: string[],
+    allTargetNamesType1: any,
+    allTargetIdsType1: any,
+    allTargetNamesType2: any,
+    allTargetIdsType2: any,
+    isVirtual: boolean,
+    logger: Logger) {
```
lipmas, line 212: again id use generics here `parseMultipleTargetTypes<T1,T2>`

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,257 @@
...
+    // Type 1 and 2 refer to the types of the targets this function is parsing
+    // Type 1 and 2 will always be Db and Web, or Bzero and DynamicAccessConfig, respectively
+    let errorMessageTarget = 'base target or dynamic access config';
+    if(isVirtual) {
+        errorMessageTarget = 'virtual target';
+    }
```
lipmas, line 218: how about just pass this error message target in as a [parameter instead of the `isVirtual` boolean.

src/utils/policy-utils.ts
```diff
@@ -0,0 +1,257 @@
...
+            const totalTargetCount = targetsType1Count + targetsType2Count;
+            // Using arrays because multiple targets with same name can exist
+            if(totalTargetCount > 1) {
+                logger.error(`More than 1 ${errorMessageTarget} with name: ${targetNameOrId}. Must use target id instead.`);
+                await cleanExit(1, logger);
+            } else if(totalTargetCount === 1) {
+                // Each pair's value is an array that can have multiple values
+                allTargetNamesType1[targetNameOrId].forEach((t:any) => {
+                    targets.push(t);
+                });
```
lipmas, line 247: Shouldnt this be:

```
else if(targetsType1Count === 1) {
...
```

lipmas (approved):
Code changes look good...retested LGTM!

PR #504, opened 2022-08-16
Multi-Db

ymarcus93:
Related PRs: https://github.com/bastionzero/webshell-backend/pull/1208, https://github.com/bastionzero/webshell-common-ts/pull/88

## Description of the change

Adds support for multiple database connections simultaneously. Listed below are some highlights:
- Add ability to list database connections with list-connections command
- Add ability to close database connections with the `zli close` command
- Updates `zli disconnect`, `zli status` (now named to `zli list-daemons`), and `zli logout` to work with multiple DB daemons
- Centralizes all daemon management logic to new service: `DaemonManagementService`. It is currently only being used by db codepath, but it can be used by other daemons (kube, web) in a future refactor PR.
- Check if local port is available for daemon by attempting to bind:
   - Change behavior of `checkIfPortAvailable()` to follow same logic we use
when doing interactive login (`zli login`). We use the `tcp-port-used`
library to check if a port is available by attempting to create and bind
to the socket for the provided port. If bound successfully, then we can
assume the port is available.
   - This change was necessary to fulfill a product requirement for multi-db. See commit details for more info.
- Refactor some system tests
   - Split vt test suite into `db` and `web`
   - New API for `EnsureConnectionEventCreated()`: Use Partial type and defaults using the spread operator and converts to using `waitForExpect()`.
- Add `jest-mock-extended` for easier type-safe mocks when writing unit tests

### Limitations

Since we do not lock the `zli`'s config file (managed by `config-service.ts`), multiple `zli` processes can compete when writing the file and they clobber each other's work. The implication of this is that a user must wait for `zli connect` to a db target to finish executing before issuing another `zli connect` command. Otherwise, the list of `dbDaemons` will not be consistent and the user will have to rely on `zli close` to close connections (`disconnect` and `logout` will not be reliable because they're based on local config file).

### Testing multi-db connections

There are system tests for this whole flow: `connect`, `list-connections`, `list-daemons` (previously known as `status`), `disconnect`, `close`:

```
multi-db virtual target connect - us-east-1 - bz-ubuntu (10510 ms)
list db connections - us-east-1 - bz-ubuntu - zli ld (2103 ms)
list db connections - us-east-1 - bz-ubuntu - zli lc (2377 ms)
deleted db target should close db connection - us-east-1 - bz-ubuntu (6450 ms)
close single db connection - us-east-1 - bz-ubuntu - zli close (12807 ms)
close multiple db connections - us-east-1 - bz-ubuntu - zli disconnect db (12590 ms)
close multiple db connections - us-east-1 - bz-ubuntu - zli close -t db --all (12725 ms)
db virtual target connect - us-east-1 - bz-ubuntu - reconnect via db pool (13802 ms)
```

You can test this manually by attempting to run `zli connect` to multiple DB targets. You can also connect to the same target twice assuming you haven't set local port field in db target config on backend (you can circument this by using `--customPort`).

### Testing migration

There are unit tests for the handling of legacy DB connections in the `daemon-management-service.test.ts` file. However, the migration flow driven by `conf`:

https://github.com/bastionzero/zli/blob/c2bc16b21b8e5065f737287e13ead8c6d120c0fc/src/services/config/config.service.ts#L93-L135

is not automatically tested. You can test this manually by opening a db connection before checking out this PR, or faking an old config by setting `dbConfig` in the JSON and changing the migration version field that `conf` uses. Like so:

```
"dbConfig": {
  "localHost": "localhost",
  "localPort": 3001,
  "localPid": 2300,
  "name": "test"
},
<...>
"__internal__": {
  "migrations": {
     "version": "6.7.0"
   }
}
```

Then, run any `zli` command to trigger the migration. Assert that it's working by running `zli ld` and seeing some output related to your old db connection.

**backend branch:** feat/multi-db
**bzero branch:**

- [X] Yes

Release Notes: Adds support for multiple database connections simultaneously. Fixes typo in zli close --help example. Deprecate zli status; replaced with zli list-daemons (alias: zli ld).

## Related JIRA tickets

Relates to JIRA: CWC-1739, CWC-1797

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/handlers/close-connection/close-connection.command-builder.ts
```diff
@@ -11,11 +16,22 @@ export function closeConnectionCmdBuilder(yargs : yargs.Argv<{}>) : yargs.Argv<c
...
+                alias: 't',
+                describe: 'Filter for specific connection type when using the --all flag',
+                implies: 'all'
+            }
+        )
         .example('$0 close d5b264c7-534c-4184-a4e4-3703489cb917', 'close example, unique connection id')
-        .example('$0 close --all', 'close all shell connections');
+        .example('$0 close --all', 'close all connections')
+        .example('$0 close --all -t shell', 'close all shell connections in cli-space')
+        .example('$0 close --all -t db', 'close all db connections');
```
SaiDadireddy, line 36: If we're able to close db connections using the `close` command with your changes, does the `disconnect` command no longer support disconnecting from db daemons?

ymarcus93: You can still use `zli disconnect` or `zli disconnect db` to kill db daemons. The difference between `disconnect` and `close` is that `disconnect` only kills daemons running locally on your machine while `zli close` lets you close all (or a specific) DB connection running on _any_ of your machines (e.g. `zli` logged in as same user at work vs. home).

src/services/daemon-management/daemon-management.service.ts
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * Some type safety is achieved by mapping the JSON config type T (DaemonConfig)
+ * to T2 (ManagedDaemonConfig) which is a union type that can be discriminated
+ * based on the ManagedDaemonConfigType field.
+ *
+ * This class handles the legacy configuration (non-KV dictionary) and the new
+ * configuration where daemon configs are stored in a dictionary keyed by their
+ * connectionId. The class can be simplified once we can move away from the
+ * legacy configuration.
+ */
```
lipmas, line 38: Im wondering if we could simplify this class now by removing the legacy configuration `getLegacyConfiguration` and `setLegacyConfiguration` from the `DaemonStore` if we instead just used a [conf migration ](https://github.com/sindresorhus/conf#migrations)to migrate the legacy config. 

Something like:

```ts
 // migrate old dbConfig to new dbDaemons format 
  '>=6.8.0': (config: Conf<BastionZeroConfigSchema>) => {
      const legacyConfig = config.get('dbConfig');
      config.set('dbDaemons', { 'n/a': legacyConfig })
  }
```

src/services/daemon-management/daemon-management.service.ts
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /**
+     * Add a daemon config to the KV store
+     * @param connectionId Connection ID bound to this daemon. Must be defined
+     * otherwise this function throws an error.
+     * @param config Daemon config object
+     */
+    public addDaemon(connectionId: string, config: T) {
+        if (connectionId === undefined) {
```
lipmas, line 95: Maybe we should use `!connectionId` instead which would also catch null or empty string?

src/services/daemon-management/daemon-management.service.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+export function newDbDaemonManagementService(
+    dbDaemonStore: DbDaemonStore,
+    processManager: ProcessManager
+): DaemonManagementService<DbConfig, ManagedDbConfig> {
+    return new DaemonManagementService(
+        getDefaultDbConfig,
+        newManagedDbConfig,
+        processManager,
```
lipmas, line 248: Seems like everywhere this is instantiated we always do:

```
const processManager = new ProcessManagerService();
```

So maybe we can just move that to the constructor for now until there is a use case for having a separate process manager implementation.

src/services/daemon-management/types/managed-daemon-config.types.ts
```diff
@@ -0,0 +1,36 @@
...
+    Web: 'web'
+} as const;
+export type ManagedDaemonConfigType = typeof ManagedDaemonConfigType[keyof typeof ManagedDaemonConfigType];
+
+export type ManagedDaemonConfig =
+    | ManagedDbConfig
+    | ManagedKubeConfig
+    | ManagedWebConfig;
+
+interface ManagedConfig<T extends DaemonConfig> {
```
lipmas, line 15: So the `ManagedConfig` type is just adding a `type` property to the config type. Im sure there is a reason im not seeing but couldnt we just have added this type field directly to the `DbConfig`/`WebConfig`/`KubeConfig` instead of creating a separate type?

src/services/daemon-management/daemon-management.service.ts
ymarcus93: Oh wow I totally forgot that `conf` can do migrations. Yeah this seems way simpler than having to deal with legacy configs in the `DaemonManagementService`.

And then I can remove `dbConfig` key after doing `config.set('dbDaemons')` line?

src/services/daemon-management/daemon-management.service.ts
ymarcus93: `DaemonManagementService` takes it in as an interface, so that we can mock it in unit tests. I can make it so the parameter is optional and default it to `new ProcessManagerService()` in the constructor. That way we reduce code duplication, like you mentioned above, and we still get the flexibility of injecting mocks (or other implementations of that service one day).

src/services/daemon-management/types/managed-daemon-config.types.ts
ymarcus93: Yes. That's correct. It adds the `type` property, so we can do discriminated unions in the code. I wanted to add the field directly to `DbConfig`, `WebConfig`, and `KubeConfig`, but I was worried about existing configs. I guess now that you mentioned `conf` can do migrations, I can make it so each of them get this type field added. That was really my only main concern in not adding it there.

One potential downside, probably not important though, is that if I add it to the `DaemonConfig` type (the one stored in the JSON configuration), the conf schema now contains a key: `<type>config` with a value that also includes the discriminant property of the same `<type>`. Not sure if that's awkward or not.

src/services/list-connections/list-connections.service.ts
```diff
@@ -0,0 +1,68 @@
...
+): Promise<DbConnectionInfo[]> {
+    const dbTargetService = new DbTargetService(configService, logger);
+    const connectionHttpService = new ConnectionHttpService(configService, logger);
+
+    const openDbConnections = await connectionHttpService.ListDbConnections(true, ConnectionState.Open);
+    if (openDbConnections.length === 0) {
+        return [];
+    }
+
+    const dbTargetSummaries = await dbTargetService.ListDbTargets();
```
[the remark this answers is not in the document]
ymarcus93: I like this idea and it stops the issue we mentioned offline about there being a race between connection still existing but target being gone

src/system-tests/tests/suites/web.ts
```diff
@@ -13,23 +12,26 @@ import { SubjectType } from '../../../../webshell-common-ts/http/v2/common.types
...
+export const webSuite = () => {
+    describe('web suite', () => {
         let policyService: PolicyHttpService;
         let testUtils: TestUtils;
 
         let testPassed = false;
 
-        const localDbPort = 6100;
-        const localWebPort = 6200;
+        // Proxy policy ID created for this entire suite in order to make DB
```
lipmas, line 30: nit: make web connections

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -115,14 +125,16 @@ export class TestUtils {
             const startTimestamp = new Date();
             startTimestamp.setHours(startTimestamp.getHours() - EVENT_QUERY_TIME);
             const events = await this.eventsService.GetConnectionEvents(startTimestamp, [configService.me().id]);
+            // this.logger.warn(JSON.stringify(events, null, 4))
```
lipmas, line 128: remove

src/system-tests/tests/utils/test-utils.ts
```diff
@@ -96,15 +96,25 @@ export class TestUtils {
...
+    * (i.e. client connect -> closed exists in our db events logs)
+    * @param {string} targetId Target id we are looking for
+    * @param {string} targetName Target name we are looking for
+    * @param {string} targetUsers Target user we are connected as
+    * @param {string} targetType Target type we are looking for (i.e. CLUSTER)
+    * @param {string} targetEnvId Target environment ID we are looking for
+    * @param {ConnectionEventType} eventType Event we are checking for
+    * @param extraEventFilter Extra predicate to use when filtering for specifc event
+    */
+    public async EnsureConnectionEventCreatedWithEventFilter(
```
lipmas, line 109: It feels strange to allow passing in extra event filters this way that dont actually get checked in the final `toMatchObject`. We could refactor this function to take in the expected ConnectionEventDataMessage and instead have the caller construct it by exposing the `BuildConnectionEvent` function. This function could take a bunch of optional values and do something like this instead:

```ts
const toReturn: ConnectionEventDataMessage = {
  ...
  connectionId: connectionId ?? expect.anything(),
  ...
```

src/services/daemon-management/daemon-management.service.ts
ymarcus93: Hmm. Not sure how easy it will be to unit test this migration. I'll have to think about it for a bit.

src/services/daemon-management/daemon-management.service.ts
ymarcus93: 3f14a68

src/services/daemon-management/daemon-management.service.ts
ymarcus93: a0ab73a

src/system-tests/tests/suites/web.ts
ymarcus93: b81032f

src/system-tests/tests/utils/test-utils.ts
ymarcus93: b81032f

src/system-tests/tests/utils/test-utils.ts
ymarcus93: I've refactored the function entirely.

5bcabdf

- Use `expect.arrayContaining` so that if the match fails `jest` tells us what events it did see in the test error message.
- Use `waitForExpect()` as you mentioned to use offline.
- Use `Partial<ConnectionEventDataMessage>` (https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) so caller can choose which properties of the type they want to check. It's type-safe and gives us the option to check more parameters as the underlying type or testing requirements evolve, without having to change the method parameters for `EnsureConnectionEventCreated()`.
- Defaults to `expect.anything()` as you mentioned above or some other sane default depending on the property.
- Made querying of environment name optional if already provided.

ymarcus93: I also added a `testStartTime` global (global bad--I know, but we've been doing that for system tests for some time now and also these are just tests), so we don't have to do the -2 hour thing we were doing before. And I've changed the query for events to include `targetId` filter to help shorten the list of events queried in case the match fails and the developer has to debug from the list received.

lipmas: Nice looks good and love the use of `Partial<type>` here. RaR

src/services/daemon-management/types/managed-daemon-config.types.ts
ymarcus93: Removed `ManagedConfig` and did migration with `conf` here: c2bc16b. Still running some manual tests.

src/services/daemon-management/daemon-management.service.ts
ymarcus93: c2bc16b. Still running some manual tests.

I removed `dbConfig` as well after doing the migration

ymarcus93: I've done some manual tests. Seems to be working well

lipmas (approved):
LGTM

src/services/daemon-management/daemon-management.service.ts
lipmas: tested the conf migration process and kube/web/db connections before/after the migration worked well

PR #505, opened 2022-08-17
Distinguish targets by environment for `SSH`

SaiDadireddy:
## Description of the change

The purpose of this PR is to update `zli generate sshConfig` to now generate host names that specify the environment of the target to be able to distinguish between bzero targets that have the same name. 

For example:
If we have:
- An environment with `Name: sai-test-env`
- A target with `Name: some-bzero-target`

The config file, `dev-bzero-bz-config`, will include the following Host names:
`Host some-bzero-target some-bzero-target.sai-test-env`

We are excluding a host name that includes the environment ID and only including a host name option with environment name because environments are unique. This should declutter the config file a little as well. 

This will allow users to be able to connect via `ssh some-bzero-target.sai-test-env` and distinguish between targets with the same name using the environment name when connecting via ssh. 

This feature does not support SSM targets and will only be able to distinguish between Bzero targets that have the same name. So for any SSM target, the only host name will be the target name. 

## Testing

To test this PR:
1. Create two bzero targets with the same name in different environments
2. Create a target connect policy for the two bzero targets that gives you, the user, `SSH Tunnel` acess
3. Run `bin/zli-macos generate sshConfig`
4. Open the `dev-bzero-bz-config` file to see the host names listed as described above. There should be one listing for each bzero target. The names of both the targets will be the same, but the environment names will be different. 
5. Connect to the two targets using a host name that includes the environment to distinguish between the targets. 

I have this set up in [my dev](https://saic.bastionzero.com/) so feel free to login, and I'll make you an admin. You'll just need to add yourself as a user to the two target connect policies I created.

**backend branch:** distinguish-ssh-by-env
**bzero branch:** 

- [x] Yes

Release Notes: 

## Related JIRA tickets

Relates to JIRA: CWC-2027

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/handlers/ssh-proxy/ssh-proxy.handler.ts
```diff
@@ -57,6 +57,8 @@ export async function sshProxyHandler(
         targetId: parsedTarget.id,
         targetName: parsedTarget.name,
         targetUser: targetUser,
+        envId: parsedTarget.envId,
+        envName: parsedTarget.envName,
```
johncmerfeld, line 61: If you take my feedback in webshell-common, these will change

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -66,8 +66,16 @@ function formatBzConfigContents(sshTargets: SshTargetsResponse[], identityFile:
     for (const target of sshTargets) {
         // only add username if there is exactly one -- otherwise, user must specify user@host
         const user = target.targetUsers.length === 1 ? `User ${target.targetUsers[0].userName}` : ``;
+        // only need to check envId because either both id and name exist or neither do
+        if(target.environmentId) {
+            contents += `
+Host ${target.targetName} ${target.targetName + '.' + target.environmentName} ${target.targetName + '.' + target.environmentId}`;
```
johncmerfeld, line 72: Very much a nit but let's not mix formatted strings and string concatenation in the same line, i.e. let's have this be
```ts
contents += `
Host ${target.targetName} ${target.targetName}.${target.environmentName} ${target.targetName}.${target.environmentId}`
```

johncmerfeld:
Config file looks good to me. I see my SSM target has the host name only, and the bzero target has the environment name and id:
```
Host ip-172-31-84-126
    IdentityFile /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-temp-key
    UserKnownHostsFile /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-known_hosts
    ProxyCommand /Users/johncmerfeld/work/code/zli/bin/zli-macos ssh-proxy --configName=dev -s dev-bzero-%n %r %p /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-temp-key
    

Host john-bzero-agent john-bzero-agent.Default john-bzero-agent.6d3b14ed-9ef6-4f21-89f7-47334d2ce9d8
    IdentityFile /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-temp-key
    UserKnownHostsFile /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-known_hosts
    ProxyCommand /Users/johncmerfeld/work/code/zli/bin/zli-macos ssh-proxy --configName=dev -s dev-bzero-%n %r %p /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-temp-key
    

Host dev-bzero-*
    IdentityFile /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-temp-key
    UserKnownHostsFile /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-known_hosts
    ProxyCommand /Users/johncmerfeld/work/code/zli/bin/zli-macos ssh-proxy --configName=dev -s %n %r %p /Users/johncmerfeld/Library/Preferences/bastionzero-zli-nodejs/bzero-temp-key
```

The only question I have (and we'd want to find this out from product): do we want some kind of comment text in this file to explain what these other things are? I could imagine a user being confused when they see all this extra content around `Host`

src/handlers/ssh-proxy/ssh-proxy.handler.ts
SaiDadireddy: [68fbbcf](https://github.com/bastionzero/zli/pull/505/commits/68fbbcf78dc67424de2119de7924bbf2d071fcd7)

src/handlers/generate/generate-ssh-config.handler.ts
SaiDadireddy: [68fbbcf](https://github.com/bastionzero/zli/pull/505/commits/68fbbcf78dc67424de2119de7924bbf2d071fcd7)

johncmerfeld (approved):
I still think you should confirm this output file with product and make sure they're comfortable with how it looks, but the code looks good to me

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -66,8 +66,16 @@ function formatBzConfigContents(sshTargets: SshTargetsResponse[], identityFile:
     for (const target of sshTargets) {
         // only add username if there is exactly one -- otherwise, user must specify user@host
         const user = target.targetUsers.length === 1 ? `User ${target.targetUsers[0].userName}` : ``;
+        // only need to check envId because either both id and name exist or neither do
+        if(target.environmentId) {
+            contents += `
+Host ${target.targetName} ${target.targetName}.${target.environmentName} ${target.targetName}.${target.environmentId}`;
```
lipmas, line 72: I like this approach of having  having multiple host alias for the same config. One thing that might not be super obvious but i tested is what happens when you have multiple alias' with the same name. So if there were multiple targets with the same name there would be two different configurations with the same `${target.targetName}`. This doesnt cause any errors though and in this case if you do `ssh ${target.targetName}` the first one defined in the config file will be used.

If you didnt want the first one though thats still fine because you can still use the `${target.targetName}.${target.environmentName}` or `${target.targetName}.${target.environmentId}` alias instead to use the more specific target.

I do agree it would be nice to add a comment or something explaining this in the config file itself though.

SaiDadireddy: Yep this is right. There would be two different configuration, with the environment to distinguish. Actually making a change right now to remove the host name option with the environment ID. So, users will only have the option to connect with the environment name which is okay because environments are unique.

SaiDadireddy: Also, the first one defined in the config file is used. However, it should still hit the more than 1 target with the same name error in the backend.

lipmas: > Also, the first one defined in the config file is used. However, it should still hit the more than 1 target with the same name error in the backend.

Ah right. It worked for me but thats because the other target with the same name was offline i think.

SaiDadireddy: [0c325dd](https://github.com/bastionzero/zli/pull/505/commits/0c325dd77aebe5f979d54f240ceba259a4f2d89b)

SaiDadireddy: Removing environment id from ssh config file -> [0c325dd](https://github.com/bastionzero/zli/pull/505/commits/0c325dd77aebe5f979d54f240ceba259a4f2d89b)

PR #510, opened 2022-08-23
Improve SSH config file management

SaiDadireddy:
## Description of the change

The goal of this user story is to provide context in `bzero-bz-config` and the user's `ssh/config` file to make it easy to use for those who do not read the manual.

The original SSH configuration file should contain a descriptive comment with each Include statement and a note at the end stating no other changes have been made. This descriptive comment will reinforce the value of managing SSH through BastionZero.

If the user loses access after these files and contents are added, then running `zli generate sshConfig` will delete the bz config file and remove the corresponding include statement in the user's config file. If there is only one include statement, then the note is also removed.  

## Testing

You will need the [ticket](https://commonwealthcrypto.atlassian.net/browse/CWC-2036)  to reference the output of the test cases. Refer to Ann's comment.

**Test case 1:**
You have a target connect policy that gives you `ssh tunnel` or `file transfer` access.
1. Run `zlil generate sshConfig` 
2. Expect to see the bz file generated with a commented out file description at the top
3. Expect to see both the include statement and a description appended to the top of the config file
4. If you haven't run this before, there will also be a note added after the include statement

**Test case 2:**
You have a target connect policy that gives you neither `ssh tunnel` nor `file transfer` access and you have not generated bzconfig file before. So, there should be no include statement in the user's config file. 
1. Run `zlil generate sshConfig` 
2. Expect to see messages `No action as been taken. You do not have tunnel or file transfer access to any targets. BZ SSH configuration file not generated.`
3. Your config file should not change and there should be no bz config file should be generated.

**Test case 3:**
You have a target connect policy that gives you neither `ssh tunnel` nor `file transfer` access and you have generated a bzconfig before and the Include statement exists in the your config. (A user who lost ssh tunnel or file transfer access after running the command)
1. Run `zlil generate sshConfig` 
2. Expect to see message `You do not have tunnel or file transfer access to any targets. BZ SSH configuration file not generated.`
3. Your config file should no longer have the include statement, its description and the note. The bz file should also be deleted.

**Test case 4: Multiple bz config file edge case**
You are a user that had multiple bz config files before this. This means that you had multiple Include statements before running this command. You can re generate these bz config files to get the comments.
For ex:
```
Include /Users/saidadireddy/.ssh/dev-bzero-bz-config

Include /Users/saidadireddy/.ssh/dev-bzero-bz-config-test
```
1. Run `zlil generate sshConfig`
2. Expected output:

```
# Please read the config file below for additional information
# regarding the use of the BastionZero SSH configuration file.
Include /Users/saidadireddy/.ssh/dev-bzero-bz-config

# Note: no changes, other than this insertion, have been made to your
# existing configuration.
#********************************************************************************

Include /Users/saidadireddy/.ssh/dev-bzero-bz-config-test
```

If you then run `zlil generate sshConfig --mySshPath ~/.ssh/config --bzSshPath ~/.ssh/dev-bzero-bz-config-test`, nothing will change in the user's config file but the corresponding bz file will be generated. 

In order for the change to be reflected in the user's config file, the user needs to remove the extra include statement and run the command again. After doing so, the expected user's config file will look like:

```
# Please read the config file below for additional information
# regarding the use of the BastionZero SSH configuration file.
Include /Users/saidadireddy/.ssh/dev-bzero-bz-config-test

# Please read the config file below for additional information
# regarding the use of the BastionZero SSH configuration file.
Include /Users/saidadireddy/.ssh/dev-bzero-bz-config

# Note: no changes, other than this insertion, have been made to your
# existing configuration.
#********************************************************************************
```

**Test case 5: Deleting when there are multiple includes**
You are a user that had multiple bz config files before this and multiple corresponding Include statements in your config file (left over from the previous test case).
1. Remove all ssh tunnel and file transfer access that you have
2. Run `zlil generate sshConfig`
3. Expect to see message `You do not have tunnel or file transfer access to any targets. BZ SSH configuration file not generated.`
4. Expect for your config file to now look like:
```
# Please read the config file below for additional information
# regarding the use of the BastionZero SSH configuration file.
Include /Users/saidadireddy/.ssh/dev-bzero-bz-config-test

# Note: no changes, other than this insertion, have been made to your
# existing configuration.
#********************************************************************************
```
5. Run `zlil generate sshConfig --mySshPath ~/.ssh/config --bzSshPath ~/.ssh/dev-bzero-bz-config-test` and expect to see both the last Include statement and also the Note to be removed. (The note is removed because there was only one Include statement left. It's removed when there is no comment with the phrase `BastionZero SSH configuration` left in the file. The check for this comment happens after every time an Include statement is removed.)

You can also run `npm test generate-ssh-config`.

**backend branch:** 
**bzero branch:** develop

- [x] Yes

Release Notes: 

## Related JIRA tickets

Relates to JIRA: CWC-2036

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -19,20 +22,27 @@ export async function generateSshConfigHandler(argv: yargs.Arguments<generateSsh
...
-    // Link the ssh config path, with our new bzero-ssh config path
-    linkNewConfigFile(userConfigPath, bzConfigPath, logger);
+        // Link the ssh config path, with our new bzero-ssh config path
+        // Also include an introductory statement
+        linkNewConfigFile(userConfigPath, bzConfigPath, logger);
 
-    logger.info(`SSH configuration generated successfully! See ${bzConfigPath} for list of reachable targets`);
+        logger.info(`SSH configuration generated successfully! See ${bzConfigPath} for list of reachable targets`);
+    } else {
+        logger.info(`SSH configuration not generated. Current user does not have tunnel or file transfer access to any targets..`);
```
asamborski, line 44: Let's remove the double period here

SaiDadireddy: Good catch, addressing the edge case we talked about earlier right now, so I'll include this in that commit!

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -25,14 +28,23 @@ export async function generateSshConfigHandler(argv: yargs.Arguments<generateSsh
...
+        fs.writeFileSync(bzConfigPath, bzConfigContentsFormatted);
 
-    logger.info(`SSH configuration generated successfully! See ${bzConfigPath} for list of reachable targets`);
+        // Link the ssh config path, with our new bzero-ssh config path
+        // Also include an introductory statement
+        linkNewConfigFile(userConfigPath, bzConfigPath, logger);
+        logger.info(`SSH configuration generated successfully! See ${bzConfigPath} for list of reachable targets`);
+    } else {
+        deleteFBzConfigContents(userConfigPath, bzConfigPath, logger);
+        logger.info(`SSH configuration not generated. Current user does not have tunnel or file transfer access to any targets.`);
```
johncmerfeld, line 46: There's a couple things here I think we should tweak:
  - Let's put the length check immediately after `sshTargets` is initialized so we don't do any unnecessary work
  - Further, if  `sshTargets.length == 0`, I don't think we need to delete the include statement from the user's existing config file -- really we should touch that file as little as possible to avoid messing it up
    - There is of course the situation where the user *used* to have tunnel access to some targets, and no longer does. In that case I think it is sufficient to delete the bz config file. The "include pointing at nothing" won't cause any problems, and we save ourselves some risk and code complexity but not bothering to remove it

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -116,13 +186,91 @@ function linkNewConfigFile(userConfigFile: string, bzConfigFile: string, logger:
...
+    }
+}
+
+/**
+ * Attaches an 'Include path/to/bz-config' line to the user's ssh config file, if not there already
+ * Also includes an intro statement to give more context
+ * @param userConfigFile {string} path of the user's config file
+ * @param bzConfigFile {string} path of the BZ config file
+ * @param logger {Logger}
+ */
```
johncmerfeld, line 226: As noted above I think this function can be clipped, but in any case this doc comment describes the function above thsi one

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -60,12 +72,69 @@ function getFilePaths(userSshPath: string, bzSshPath: string, configPrefix: stri
...
+        let user = ``;
+        if(target.targetUsers.length === 1) {
+            user += `User ${target.targetUsers[0].userName}`;
+        } else if(target.targetUsers.length > 1) {
+            user += `# User `;
+            for(const targetUser of target.targetUsers) {
+                user += `${targetUser.userName} `;
+            }
+        } else {
+            user += ``;
```
johncmerfeld, line 135: Wait, why is this needed?

SaiDadireddy: good catch, removed [0a1fb12](https://github.com/bastionzero/zli/pull/510/commits/0a1fb1271892d284bf81064c2a81db7ac61e97b1)

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -19,25 +29,38 @@ export async function generateSshConfigHandler(argv: yargs.Arguments<generateSsh
...
+        logger.info(`SSH configuration generated successfully! For a list of reachable targets, see ${bzConfigPath}`);
+    } else {
+        // Build our ssh config file -- note that by using this function with 'true' we are chosing to add the prefix before our hostname token in the proxycommand
+        const { prefix } = await buildSshConfigStrings(configService, processName, logger, true);
+        // Determine the user's ssh and bzero-ssh config path
+        const { userConfigPath, bzConfigPath } = getFilePaths(argv.mySshPath, argv.bzSshPath, prefix);
 
-    logger.info(`SSH configuration generated successfully! See ${bzConfigPath} for list of reachable targets`);
+        deleteBzConfigContents(userConfigPath, bzConfigPath, logger);
+        logger.info('SSH BZ configuration file not generated.');
```
johncmerfeld, line 57: Less important: I think it's more accurate to call this a "BastionZero SSH configuration file" (BZ is fine too but it's a BZ SSH config, not an SSH BZ config)
More important: Right now this reads like an error. We should probably explain *why* it wasn't generated, e.g.:
```
logger.info('You do not have tunnel or file transfer access to any targets. BZ SSH configuration file not generated.');
```

SaiDadireddy: Yep! Thanks, this was just a placeholder before I talk to Ann. Will keep this in mind.

johncmerfeld (approved):
Tested this out; I believe it does everything it's require to do -- apart from my comment on the "zero case" log message, looks great to me!

src/handlers/generate/generate-ssh-config.handler.ts
johncmerfeld: Ahh got it, great!

lipmas (approved):

PR #516, opened 2022-08-26
Agent failsafe & recovery part 2: manual restart message

johncmerfeld:
## Description of the change

In order to provide a break-glass failsafe for agents that have entered a broken state but won't restart themselves because their control channel is still functioning, we want to give customers a way to manually restart their bzero / kube agents. This change adds the `zli target restart` command to achieve this.

## Testing

If you have an bzero or kube agent running the appropriate code, you can trigger a restart with:
```sh
# replace 'zli' with your local zli
zli target restart <your-target-name>
```

You should see a message acknowledging that the restart was initiated, and your target will come back online in about a minute.

Give this a try with target name, target id, and target-name.Env-name/id

**backend branch:** develop
**bzero branch:** develop

- [x] Yes

Release Notes: Add `zli target restart` command

## Related JIRA tickets

Relates to JIRA: CWC-2018

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain: We are adding an unauthenticated way to trigger actions on the agent. A compromised bastion could restart customers' agents. However, this is not significantly different from the current reality, where a compromised bastion could simply disconnect from customers' agents.


src/cli-driver.ts
```diff
@@ -162,7 +164,8 @@ export class CliDriver
         'policy',
         'register',
         'generate',
-        'api-key'
+        'api-key',
+        'target', // TODO: needed here?
```
lipmas, line 168: yes needs to be an oauthCommand if its making any api requests to bastion. The oauth middleware will ensure the user is logged in and error otherwise.

src/cli-driver.ts
```diff
@@ -179,6 +182,7 @@ export class CliDriver
         'ssh-proxy',
         'generate',
         'policy',
+        'target', // TODO: needed here?
```
lipmas, line 185: yeah might as well have google analytics data for when customers use this

src/system-tests/tests/system-test.ts
```diff
@@ -354,11 +354,11 @@ if (API_ENABLED) {
     }
 }
 
-if (AGENT_RECOVERY_ENABLED && BZERO_ENABLED && process.env.TEST_RUNNER_KUBE_CONFIG) {
+if (AGENT_RECOVERY_ENABLED && BZERO_ENABLED && KUBE_ENABLED && process.env.TEST_RUNNER_KUBE_CONFIG) {
```
lipmas, line 357: technically i think we could change this to
```
AGENT_RECOVERY_ENABLED && (BZERO_ENABLED || KUBE_ENABLED) && process.env.TEST_RUNNER_KUBE_CONFIG)
```
since we could run the tests as long as there are either bzero targets or kube cluster created.

lipmas (approved):

src/system-tests/tests/system-test.ts
johncmerfeld: But in the case where BZERO_ENABLED = true and KUBE_ENABLED = false, wouldn't the kube tests still run, and fail?

lipmas: oh true good point yeah we would need to condition the kube tests on `KUBE_ENABLED` too within the agent recovery suite

lipmas (approved):

PR #517, opened 2022-08-29
Agent Recovery Suite

lipmas:
## Description of the change

+ Adds new system test environment variables that expose the test runner cluster config so we can construct k8s client to interact with the cluster as part of the tests. Dependent on https://github.com/bastionzero/cwc-infra/pull/526

+ Add test that restarts the bastion and verifies that the agent control channel comes back online and we can connect to the target

## Testing

**backend branch:**
**bzero branch:** 

- [x] Yes

Release Notes: Adds new agent recovery system test suite to test agent recovery functions when various backend services fail 

## Related JIRA tickets

Relates to JIRA: CWC-2053

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

### DIFF
--- a/src/http-services/events/events.http-server.ts
+++ b/src/http-services/events/events.http-server.ts
@@ -5,6 +5,7 @@ import { ConnectionEventDataMessage } from '../../../webshell-common-ts/http/v2/
 import { CommandEventDataMessage } from '../../../webshell-common-ts/http/v2/event/types/command-event-data-message';
 import { KubeEventDataMessage } from '../../../webshell-common-ts/http/v2/event/types/kube-event-data-message.types';
 import { UserEventDataMessage } from '../../../webshell-common-ts/http/v2/event/types/user-event-data-message.types';
+import { TargetType } from '../../../webshell-common-ts/http/v2/target/types/target.types';
 
 
 export class EventsHttpService extends HttpService
@@ -50,5 +51,13 @@ export class EventsHttpService extends HttpService
             eventCount: count?.toString()
         };
         return this.Get('user', params);
+    }
+
+    public GetAgentStatusChangeEvents(targetId: string, targetType: TargetType) {
+        const params = {
+            targetId,
+            targetType,
+        };
+        return this.Get('agentstatuschange', params);
     }
 }
--- a/src/http-services/events/events.http-server.ts
+++ b/src/http-services/events/events.http-server.ts
@@ -53,7 +53,7 @@ export class EventsHttpService extends HttpService
         return this.Get('user', params);
     }
 
-    public GetAgentStatusChangeEvents(targetId: string, startTime?: Date, endTime?: Date): Promise<AgentStatusChangeData> {
+    public GetAgentStatusChangeEvents(targetId: string, startTime?: Date, endTime?: Date): Promise<AgentStatusChangeData[]> {
         const params = {
             targetId: targetId,
             startTimestamp: startTime?.toJSON(),
--- a/src/http-services/events/events.http-server.ts
+++ b/src/http-services/events/events.http-server.ts
@@ -5,7 +5,7 @@ import { ConnectionEventDataMessage } from '../../../webshell-common-ts/http/v2/
 import { CommandEventDataMessage } from '../../../webshell-common-ts/http/v2/event/types/command-event-data-message';
 import { KubeEventDataMessage } from '../../../webshell-common-ts/http/v2/event/types/kube-event-data-message.types';
 import { UserEventDataMessage } from '../../../webshell-common-ts/http/v2/event/types/user-event-data-message.types';
-import { TargetType } from '../../../webshell-common-ts/http/v2/target/types/target.types';
+import { AgentStatusChangeData } from '../../../webshell-common-ts/http/v2/event/types/agent-status-change-data.types';
 
 
 export class EventsHttpService extends HttpService
@@ -53,11 +53,13 @@ export class EventsHttpService extends HttpService
         return this.Get('user', params);
     }
 
-    public GetAgentStatusChangeEvents(targetId: string, targetType: TargetType) {
+    public GetAgentStatusChangeEvents(targetId: string, startTime?: Date, endTime?: Date): Promise<AgentStatusChangeData> {
         const params = {
-            targetId,
-            targetType,
+            targetId: targetId,
+            startTime: startTime?.toJSON(),
+            endTime: endTime?.toJSON(),
         };
+
         return this.Get('agentstatuschange', params);
     }
 }
--- a/src/http-services/events/events.http-server.ts
+++ b/src/http-services/events/events.http-server.ts
@@ -56,8 +56,8 @@ export class EventsHttpService extends HttpService
     public GetAgentStatusChangeEvents(targetId: string, startTime?: Date, endTime?: Date): Promise<AgentStatusChangeData> {
         const params = {
             targetId: targetId,
-            startTime: startTime?.toJSON(),
-            endTime: endTime?.toJSON(),
+            startTimestamp: startTime?.toJSON(),
+            endTimestamp: endTime?.toJSON(),
         };
 
         return this.Get('agentstatuschange', params);
--- a/src/system-tests/tests/suites/agent-recovery.ts
+++ b/src/system-tests/tests/suites/agent-recovery.ts
@@ -15,7 +15,7 @@ import { Environment } from '../../../../webshell-common-ts/http/v2/policy/types
 import { VerbType } from '../../../../webshell-common-ts/http/v2/policy/types/verb-type.types';
 
 export const agentRecoverySuite = (testRunnerKubeConfigFile: string, testRunnerUniqueId: string) => {
-    describe('agent recovery suite', () => {
+    describe('Agent Recovery Suite', () => {
         let k8sApi: k8s.CoreV1Api;
         let k8sExec: k8s.Exec;
         let policyService: PolicyHttpService;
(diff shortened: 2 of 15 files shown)


src/system-tests/tests/system-test.ts
```diff
@@ -306,11 +308,6 @@ if(IN_PIPELINE) {
     agentContainerSuite();
 }
 
-// BZero only test suites
-if(BZERO_ENABLED) {
-    dynamicAccessSuite();
-}
-
```
johncmerfeld, line 313: Wow, never noticed this duplication before. Nice one

src/system-tests/tests/suites/agent-recovery.ts
```diff
@@ -0,0 +1,145 @@
...
+        let connectionService: ConnectionHttpService;
+        let connectTestUtils: ConnectTestUtils;
+
+        beforeAll(async () => {
+            policyService = new PolicyHttpService(configService, logger);
+            testUtils = new TestUtils(configService, logger, loggerConfigService);
+            connectionService = new ConnectionHttpService(configService, logger);
+
+            // const kubeConfigFileContent = fs.readFileSync(testRunnerKubeConfigFile).toString();
+            // logger.info(`Test runner kube config file is ${testRunnerKubeConfigFile} with contents: ${kubeConfigFileContent}`);
```
johncmerfeld, line 32: Want to leave these logs in?

src/system-tests/tests/suites/agent-recovery.ts
```diff
@@ -0,0 +1,145 @@
...
+        });
+
+        // Called before each case
+        beforeEach(() => {
+            connectTestUtils = new ConnectTestUtils(connectionService, testUtils);
+        });
+
+        bzeroTestTargetsToRun.forEach(async (testTarget) => {
+            it(`247517: bastion restart ${testTarget.awsRegion} - ${getDOImageName(testTarget.dropletImage)}`, async () => {
+                const testStartTime = new Date();
```
johncmerfeld, line 69: Since this will probably be a recurring pattern, you could make `testStartTime` a global variable and reassign it in `BeforeEach()`

src/system-tests/tests/suites/agent-recovery.ts
```diff
@@ -0,0 +1,145 @@
...
+            });
+        });
+
+        // Called before each case
+        beforeEach(() => {
+            connectTestUtils = new ConnectTestUtils(connectionService, testUtils);
+        });
+
+        bzeroTestTargetsToRun.forEach(async (testTarget) => {
+            it(`247517: bastion restart ${testTarget.awsRegion} - ${getDOImageName(testTarget.dropletImage)}`, async () => {
```
johncmerfeld, line 68: I think we usually add the case ID as a member variable of the `BaseTarget` type, and then populate it inside the `bzeroTestTargetsToRun` object so that each target gets its own case ID. At least that's what I've typically done

src/system-tests/tests/system-test.ts
lipmas: hah yeah rebasing is hard...that one has been in there for a while :facepalm:

src/system-tests/tests/suites/agent-recovery.ts
lipmas: I thought i could get away with this because the agent recovery tests wont run in the pipeline (only run when there is a test runner) so we wont test them against any other targets. But yeah just in case that assumption changes I can move the test case ID definition to `BaseTarget`

ymarcus93: Or we can try the pattern I started in `db.ts`, so `BaseTarget` doesn't get massive:

https://github.com/bastionzero/zli/blob/586703d26dd6fa713aa6a820a2bc058c3ceb1958/src/system-tests/tests/suites/db.ts#L32-L81

https://github.com/bastionzero/zli/blob/586703d26dd6fa713aa6a820a2bc058c3ceb1958/src/system-tests/tests/suites/db.ts#L446

src/system-tests/tests/suites/agent-recovery.ts
ymarcus93: I did something like this for multi-db but didn't reassign it in global beforeeach

https://github.com/bastionzero/zli/blob/0eb4f0238c06dcee3948544e36ef54b99258e7ce/src/system-tests/tests/system-test.ts#L59

We should add that, so this function gets smaller list as well (as it will only look for events after that specific test started instead of the entire system test run):

https://github.com/bastionzero/zli/blob/586703d26dd6fa713aa6a820a2bc058c3ceb1958/src/system-tests/tests/utils/test-utils.ts#L115-L119

src/system-tests/tests/suites/agent-recovery.ts
lipmas: oh great yeah ill just do that then

lipmas: [7e82716](https://github.com/bastionzero/zli/pull/517/commits/7e8271670fa9b11fa1f276a005f85eeb9ea89f3b)

src/system-tests/tests/suites/agent-recovery.ts
lipmas: [7e82716](https://github.com/bastionzero/zli/pull/517/commits/7e8271670fa9b11fa1f276a005f85eeb9ea89f3b)

src/system-tests/tests/suites/agent-recovery.ts
lipmas: Ok i refactored `EnsureConnectionEvent` as well to take in a required `startTime` argument and have all test suites use a local start time instead of the global one [2ba6de3](https://github.com/bastionzero/zli/pull/517/commits/2ba6de395ca1852d9806cd3b2c81915ca4ae9236)

src/http-services/events/events.http-server.ts
```diff
@@ -51,4 +52,14 @@ export class EventsHttpService extends HttpService
         };
         return this.Get('user', params);
     }
+
+    public GetAgentStatusChangeEvents(targetId: string, startTime?: Date, endTime?: Date): Promise<AgentStatusChangeData> {
```
johncmerfeld, line 56: Shouldn't this return `Promise<AgentStatusChangeData[]>` (array of objects)?

lipmas: yup definitely. Weird that this didnt cause any compiler errors

lipmas: [6eb33a8](https://github.com/bastionzero/zli/pull/517/commits/6eb33a806979d46dba2b94b21854a018bda8bf7b)

src/http-services/events/events.http-server.ts
```diff
@@ -51,4 +52,14 @@ export class EventsHttpService extends HttpService
         };
         return this.Get('user', params);
     }
+
+    public GetAgentStatusChangeEvents(targetId: string, startTime?: Date, endTime?: Date): Promise<AgentStatusChangeData[]> {
+        const params = {
+            targetId: targetId,
+            startTimestamp: startTime?.toJSON(),
+            endTimestamp: endTime?.toJSON(),
```
ymarcus93, line 60: So `?` resolves it to `null` if not specified. Does the `GET` endpoint for this new endpoint understand `null` or will it complain? I think I tested something like this when adding a new `GET` endpoint and I think it didn't like null on backend when it binds

src/system-tests/tests/suites/agent-recovery.ts
```diff
@@ -0,0 +1,161 @@
...
+                // Stop the systemd service on the bastion container to simulate bastion going down temporarily
+                logger.info('stopping bastion container');
+
+                // In practice both the SIGKILL and the stop commands still
+                // result in agent going offline because of
+                // control-channel-disconnect- so the websocket is still closing
+                // normally and updating the status in the database. The harsh
+                // command is probably not working as intended because it only
+                // kills the start.sh script which calls dotnet run but not the
+                // actual Webshell.WebApp process itself.
```
ymarcus93, line 101: Oh...that's unfortunate. But anyways this will get changed soon

src/http-services/events/events.http-server.ts
lipmas: Doesnt it resolve to `null` i thought it would be `undefined` which would just not be serialized when sending query params

src/system-tests/tests/suites/agent-recovery.ts
lipmas: yeah i could hack something together with pkill and match `Webshell.WebApp` but since this is changing soon anyway i just left it.

src/http-services/events/events.http-server.ts
ymarcus93: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

Yup you're correct. Ok I think I can simplify the GET endpoint (for list DB) I added to use `?` as well

> this is a way of saying that when foo is defined, foo.bar.baz() will be computed; but when foo is null or undefined, stop what we’re doing and just return undefined.”

ymarcus93: Both of these actually:

https://github.com/bastionzero/zli/blob/586703d26dd6fa713aa6a820a2bc058c3ceb1958/src/http-services/events/events.http-server.ts#L17-L29

https://github.com/bastionzero/zli/blob/586703d26dd6fa713aa6a820a2bc058c3ceb1958/src/http-services/connection/connection.http-services.ts#L78-L88

src/system-tests/tests/suites/agent-recovery.ts
```diff
@@ -0,0 +1,161 @@
...
+        });
+
+        // Called before each case
+        beforeEach(() => {
+            testStartTime = new Date();
+            connectTestUtils = new ConnectTestUtils(connectionService, testUtils);
+        });
+
+        bzeroTestTargetsToRun.forEach(async (testTarget) => {
+            it(`${fromTestTargetToCaseIdMapping(testTarget)}: bastion restart ${testTarget.awsRegion} - ${getDOImageName(testTarget.dropletImage)}`, async () => {
```
ymarcus93, line 85:
```
Agent Recovery Suite
--
712 | ✓ [object Object]: bastion restart us-east-1 - bz-ubuntu (93963 ms)
```

Forgot to use dot to access property I think

ymarcus93 (approved):

johncmerfeld (approved):

PR #521, opened 2022-09-01
Add default target user for shell and ssh connections

SaiDadireddy:
## Description of the change

These changes allow a user to set a default target user in their `prod.json` configuration file. 
This default target user will be used when connecting via `zli connect` and `ssh` in the case where a user does not manually enter a target user in their target string. 

For `ssh`, the user must run the `zli generate sshConfig` command in order to write the default target user into their `bzero-bz-config` file. After doing so, the default user will take effect. To revert this and use the target users specified by the policy, they simply need to run `zli configure --targetUser reset` to set default target user to null and run `zli generate sshConfig` again. This is a workaround for an ssh issue where ssh disregards the manually set default user and uses the local OS username instead. 

**Recap:** If a default target user is set and the no target user is set in the target string by the user, `zli connect` will automatically use the default user. For `ssh` it is a two step process where the user will need to regenerate their ssh config file. In either case, the default user can be overridden by specifying a user in the target string. 

## Testing

**Test 1**: Configuring targetUser - success case
1. Run: `zli configure default-targetuser ec2-user`
2. Expect to see `ec2-user` set for `targetUser` in `connectConfig` in your `dev.json`
3. Expect to see message `Local default target user for shell, ssh, and scp set to ec2-user. This has been saved to your config here: ...`

**Test 2**: Configuring targetUser - no argument 
1. Run: `zli configure default-targetuser`
2. Expect to see an error saying `A target user must be provided for this command.`
3. Expect to see message `Local default target user for shell, ssh, and scp has been removed.`

**Test 3**: Reset default target user
1. Run: `zli configure default-targetuser --reset`.
2. Expect to see `null` set for `targetUser` in `connectConfig` in your `dev.json`

**Test 4**: Set default target user and use `--reset` flag
1. Run: `zli configure default-targetuser ec2-user --reset`.
2. Expect to see an error saying `Arguments targetUser and reset are mutually exclusive`

**Test 5**: Configure command without flag
1. Run: `zli configure`.
2. Expect normal behavior (file paths for config, zli logs, and daemon logs to be printed)

**Test 6**: Using default target user for `zli connect`
1. Create a bzero target with name `bzero-test-target1`
2. Run: `zli configure default-targetuser ec2-user`
4. Run: `zli connect bzero-test-target1`
5. Expect to see `ec2-user` as target user when in the shell and in the connection and command event logs

**Test 7**: Using default target user for `zli connect`
1. Same steps as Test 5, except set default target user to ssm-user

**Test 8**: Using default target user for `zli connect`
1. Same steps as Test 6, except run `zli connect ssm-user@bzero-test-target1`
2. Expect `ssm-user` to be used as target user and not `ec2-user`

**Test 9**: Ssh config files should reflect the default user
1. Create a bzero target with name `bzero-test-target1`
2. Run: `zli configure default-targetuser default-user`
4. Run `zli generate sshConfig` 
5. Expect to see `default-user` as the User for each configuration in your `bzero-bz-config` file

**Other ssh success tests**:
Using the two commands `zli configure default-targetuser ...` and `zli generate sshConfig`, replicate tests 6-9. Set a valid Unix default user and connect via ssh.

**Failure cases:**
1. If there is no default target user set, and multiple target users are allowed by a policy, then connecting to a target without specifying a target user in the target string, the normal error message `Server Error: Multiple allowed target users found for target ...` should be printed.
2. Whether there is a default target user set or not, if there is no valid policy to connect to the given target, the normal error message `Server Error: You do not have a TargetAccess policy setup to access ... with Verb ...` should be printed.
3. If the default target user is not allowed by a policy, then connecting with that target user should result in `Server Error: You do not have permission to connect as targetUser: ...` being printed.

**backend branch:** 
**bzero branch:** develop

- [x] Yes

Release Notes: 

## Related JIRA tickets

Relates to JIRA: CWC-1512

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/system-tests/tests/suites/connect.ts
```diff
@@ -186,5 +187,24 @@ export const connectSuite = () => {
                 testPassed = true;
             }, 60 * 1000);
         });
+
+        bzeroTestTargetsToRun.forEach(async (testTarget: TestTarget) => {
+            it(`${testTarget.connectDefaultUserCaseId}: zli connect with default targetuser - ${testTarget.awsRegion} - ${testTarget.installType} - ${getDOImageName(testTarget.dropletImage)}`, async () => {
+                // set the default user to ssm-user
+                await callZli(['configure', '--targetUser', 'ssm-user']);
```
lipmas, line 194: I think this be using `bzeroTargetCustomUser` instead of `ssm-user` for bzero targets. We might have kept the `ssm-user` around for backwards compatibility in bzero targets but eventually this will go away.

src/handlers/configure/config.handler.ts
```diff
@@ -0,0 +1,32 @@
...
+        configService.setConnectConfig(connectConfig);
+        logger.info(loggerStatement);
+    }
+
+    // if the flag isn't specified, the value is undefined
+    if(argv.targetUser === undefined) {
+        logger.info(`You can edit your config here: ${configService.configPath()}`);
+        logger.info(`You can find your zli log files here: ${loggerConfigService.logPath()}`);
+        logger.info(`You can find your daemon log files here: ${loggerConfigService.daemonLogPath()}`);
+    }
```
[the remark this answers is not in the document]
SaiDadireddy: @asamborski I like this idea, what do you think?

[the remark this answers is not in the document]
asamborski: Yes I think this sounds cleaner. Just to visualize what this would play out to look like: 
* zli configure defaultTargetUser <targetuser>: set the default target user to <target user>
* zli configure defaultTargetUser: error returned that <targetuser> is required
* zli configure defaultTargetUser --reset: default target user is reset to null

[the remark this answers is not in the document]
SaiDadireddy: Yes, but the target user would be a positional so imagine `zli configure defaultTargetUser ec2-user` if no argument is specified then it should give an error. Reset will work as you mentioned.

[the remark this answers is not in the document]
asamborski: This will work out well if we do incorporate zli default-targetgroup in the future because it will be a subcommand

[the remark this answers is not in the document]
SaiDadireddy: @lipmas [dcc36a6](https://github.com/bastionzero/zli/pull/521/commits/dcc36a6439a8cf9c4b070299c434160ffbd2c0e4)

src/cli-driver.ts
```diff
@@ -653,8 +655,22 @@ export class CliDriver
...
             .command(
                 'configure',
-                'Returns config file path',
-                () => {},
+                'Retrieve paths for config file, zli logs and daemon logs. See help menu for setting defaults.',
+                (yargs) => {
+                    return yargs
+                        .command(
+                            'defaultTargetUser [targetUser]',
+                            'Set a local default target user',
```
asamborski, line 663: we should specify this only applies to SSH

src/handlers/configure/config-default-targetuser.command-builder.ts
```diff
@@ -0,0 +1,26 @@
...
+{targetUser: string} &
+{reset: boolean};
+
+export function configDefaultTargetUserCommandBuilder(yargs : yargs.Argv<{}>) : yargs.Argv<configDefaultTargetUserArgs> {
+    return yargs
+        .positional('targetUser',
+            {
+                type: 'string',
+                demandOption: false,
+                description: 'Set a local default target user',
```
asamborski, line 13: same comment here - should specify it's for ssh

src/handlers/configure/config-default-targetuser.command-builder.ts
```diff
@@ -0,0 +1,26 @@
...
+            }
+        )
+        .option('reset',
+            {
+                type: 'boolean',
+                demandOption: false,
+                description: 'Reset the local default target user',
```
asamborski, line 20: same comment here - should specify it's for ssh

src/handlers/configure/config-default-targetuser.command-builder.ts
```diff
@@ -0,0 +1,26 @@
...
+                alias: 'r'
+            }
+        ).conflicts('targetUser', 'reset')
+        .example('$0 configure defaultTargetUser ec2-user', 'Set ec2-user as a default target user for shell, ssh, and scp')
```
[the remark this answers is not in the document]
johncmerfeld: Sorry just saw this after talking it over with Sai. Basically I think either is fine; we're arguably referring to the programs `ssh` and `scp` here; you could argue we're referring to the protocols SSH and SCP but I think this is fine!

For example, the man page for scp writes it like this:
image

src/handlers/generate/generate-ssh-config.test.ts
```diff
@@ -103,4 +107,98 @@ describe('Generate ssh config suite', () => {
...
+            targetUser: 'ec2-user',
+            targetGroups: undefined,
+            targetType: TargetType.Bzero*/
+        expect(getUniversalConnectionSpy).toHaveBeenCalledWith(mockUniversalConnectionRequest);
+
+        // Call zli configure to clear the local default user
+        await mockArgv(['configure', 'defaultTargetUser', '--reset'], async () => {
+            const driver = new CliDriver();
+            await driver.run(process.argv.slice(2), true);
+        });
```
johncmerfeld, line 202: This should probably go in an `AfterEach` or `AfterAll` to ensure it runs even if the test hits an error

src/cli-driver.ts
```diff
@@ -653,8 +655,22 @@ export class CliDriver
             )
             .command(
                 'configure',
-                'Returns config file path',
-                () => {},
+                'Retrieve paths for config file, zli logs and daemon logs. See help menu for setting defaults.',
+                (yargs) => {
+                    return yargs
+                        .command(
+                            'defaultTargetUser [targetUser]',
```
johncmerfeld, line 662: Our other multi-word commands use dashes instead of camelcase. I think we should do that here as well

src/handlers/configure/config-default-targetuser.handler.ts
```diff
@@ -0,0 +1,28 @@
...
+        logger.error('A target user must be provided for this command.');
+        await cleanExit(0, logger);
+    }
+
+    let connectConfig: ConnectConfig;
+    let loggerStatement;
+    if(argv.reset === undefined) {
+        connectConfig = { targetUser: argv.targetUser };
+        loggerStatement = `Local default target user for shell, ssh, and scp set to ${argv.targetUser}. This has been saved to your config here: ${configService.configPath()}`;
+    } else {
```
johncmerfeld, line 21: Feel free to not do this, but if the user both passes `--reset` and sets a user, we might want to log that we're ignoring the user and just doing the reset

src/handlers/generate/generate-ssh-config.handler.ts
```diff
@@ -144,6 +144,11 @@ function formatBzConfigContents(sshTargets: SshTargetsResponse[], identityFile:
             }
         }
 
+        // if a default user is set, then override the target user string
+        if(configService.getConnectConfig().targetUser) {
+            user = `User ${configService.getConnectConfig().targetUser}`;
+        }
```
johncmerfeld, line 150: Assuming the user string was originally a commented out list of possible users, do we still want to include that for reference? Too convoluted?

src/handlers/configure/config-default-targetuser.handler.ts
SaiDadireddy: I think it would make more sense to only allow one or the other, which is the current functionality.

src/cli-driver.ts
SaiDadireddy: Good point, adding that in

src/handlers/generate/generate-ssh-config.test.ts
SaiDadireddy: [4ec6d81](https://github.com/bastionzero/zli/pull/521/commits/4ec6d8133e4f8de92f7a519f5cb7a87791dc24a8)

src/cli-driver.ts
SaiDadireddy: [4ec6d81](https://github.com/bastionzero/zli/pull/521/commits/4ec6d8133e4f8de92f7a519f5cb7a87791dc24a8)

src/handlers/configure/config-default-targetuser.handler.ts
johncmerfeld: Aha, my mistake, I didn't realize we'd set them as mutually exclusive. Great!

src/cli-driver.ts
```diff
@@ -653,8 +655,22 @@ export class CliDriver
             )
             .command(
                 'configure',
-                'Returns config file path',
-                () => {},
+                'Retrieve paths for config file, zli logs and daemon logs. See help menu for setting defaults.',
```
johncmerfeld, line 658: Sorry to be a nag but there's one thing about this that I just noticed. The help menu `zli configure help` show this:
```
zli configure

Retrieve paths for config file, zli logs and daemon logs. See help menu for setting defaults.

Commands:
  zli configure default-targetuser [targetUser]  Set a local default target user for shell, ssh, and scp

Options:
      --debug    Flag to show debug logs  [boolean] [default: false]
  -s, --silent   Silence all zli messages, only returns command output  [boolean] [default: false]
      --help     Show help  [boolean]
      --version  Show version number  [boolean]
```

Just from this I don't think it's clear that `zli configure` on its own is a valid command. Is there a way we can add an example for that?

johncmerfeld (approved):
See my comment regarding the help menu, but otherwise -- Tested and it looks good to me!

lipmas (approved):

src/cli-driver.ts
SaiDadireddy: Adding in the following: 
```
Examples:
  zli configure  Retrieve paths for config file, zli logs and daemon logs.
```

SaiDadireddy:
```
zli configure

Retrieve paths for config file, zli logs and daemon logs. See help menu for setting defaults.

Commands:
  zli configure default-targetuser [targetUser]  Set a local default target user for shell, ssh, and scp

Options:
      --debug    Flag to show debug logs  [boolean] [default: false]
  -s, --silent   Silence all zli messages, only returns command output  [boolean] [default: false]
      --help     Show help  [boolean]
      --version  Show version number  [boolean]

Examples:
  zli configure  Retrieve paths for config file, zli logs and daemon logs.
```

SaiDadireddy: @johncmerfeld [e12a9f8](https://github.com/bastionzero/zli/pull/521/commits/e12a9f89f77cefbf063026a7f6b72e48defd4715)

johncmerfeld: @asamborski what do you think of the above?

SaiDadireddy: Fyi though - we can't control where Examples is written in

asamborski: Yeah I'm assuming since default-targetuser is a subcommand, yargs doesn't recognize just zli configure as a standalone command?

SaiDadireddy: It does the commands section you see is technically listing subcommands

SaiDadireddy: Here is what `zli policy` looks like for reference:
```
zli policy

List, create and update functionality for policies

Commands:
  zli policy list [type]                                    List all policies
  zli policy describe-cluster-policy <clusterName>          List the detailed information about what policies apply to a given cluster
  zli policy users                                          List the BastionZero users in the organization
  zli policy groups                                         List the organization's SSO groups
  zli policy targetusers <policyName>                       List the given policy's target users
  zli policy targetgroups <policyName>                      List the given policy's target groups
  zli policy add-user <policyName> <idpEmail>               Add a user to an existing policy
  zli policy add-group <policyName> <groupName>             Add a group to an existing policy
  zli policy add-targetuser <policyName> <targetUser>       Add a target user to an existing policy
  zli policy add-targetgroup <policyName> <targetGroup>     Add a target group to an existing policy
  zli policy create-cluster                                 Create a cluster policy. See help menu for required and optional flags.
  zli policy create-tconnect                                Create a target connect policy. See help menu for required and optional flags.
  zli policy create-recording                               Create a session recording policy. See help menu for required and optional flags.
  zli policy create-proxy                                   Create a proxy policy. See help menu for required and optional flags.
  zli policy delete-user <policyName> <idpEmail>            Delete a user from an existing policy
  zli policy delete-group <policyName> <groupName>          Delete a group from an existing policy
  zli policy delete-targetuser <policyName> <targetUser>    Delete a target user from an existing policy
  zli policy delete-targetgroup <policyName> <targetGroup>  Delete a target group from an existing policy

Options:
      --debug    Flag to show debug logs  [boolean] [default: false]
  -s, --silent   Silence all zli messages, only returns command output  [boolean] [default: false]
      --help     Show help  [boolean]
      --version  Show version number  [boolean]
```

johncmerfeld: Yeah the tricky difference there of course is that `zli policy` isn't a valid command by itself

SaiDadireddy: Yep, was just trying to show how subcommands are listed as `Commands:` by yargs.

johncmerfeld: Well, it's probably the best we can do using yargs...

PR #535, opened 2022-09-19
Send logs to S3 Bucket

SaiDadireddy:
## Description of the change

This pull request is for the `zli send-logs` feature. This feature allows users to package their zli, daemon, and agent logs (for a specific bzero target) and send them to an S3 bucket managed by the BastionZero dev team. 

The feature will send logs from the current day (the day the command is run), and the previous day (filtered by the datestamp in each log). This will allow us to have anywhere between the past 24-48 hours of logs. 

**Helpful note:** Last two days means that if I run the command today on Oct 4th, then all the logs starting from 12:00am Oct 3rd until this moment will be retrieved, which is why I listed it as the past 24-48 hours. (Query logs for Oct 3 and Oct 4)

The filesystem in the S3 bucket is as follows:
`Month/Customer_Email_Domain/Customer_Email/Request_UUID/[your posted zipped log files]`

This feature is supported by the following commands:
`zli send-logs` which will zip the zli and daemon logs stored on the user's machine and post them to S3. 
`zli send-logs --target [some_agent_target]` which will zip the bzero/kube logs stored on the agent `some_agent_target` and post them to S3. 
`zli send-logs --all [some_agent_target]` which will do a conjunction of both previous commands. 

## Testing

**Test 1**
1. Run `zlil configure` and make sure that you have logs existing in your daemon and zli dev files to post to S3.
2. Run `zli send-logs`.
4. Expect to see the unique UUID for this request printed to the console and also to see a slack notification in `#customer-logs-alarms`. The slack notification will also contain this UUID for reference along with the full object path.
5. Go to the `bastionzero-dev-customer-logs` S3 bucket in AWS and find your files using the filesystem listed above for reference.
6. Download the zip files, unzip them and expect to see all **zli and daemon** logs, timestamped with the last two days. 

 **Test 2**
1. Run `zli send-logs --target [some_agent_target]`.
3. Follow steps 3 and 4 listed for Test 1
4. Download the zip files, unzip them and expect to see all **agent** logs for the target you specified. You should see the current `bzero-agent.log` file and also up to 2 rotated log files for bzero agents. The explanation for how rotated log files are handled can be found in [the bzero pull request](https://github.com/bastionzero/bzero/pull/153). You should be able to see a log file with logs from the past 48 hours for kube agents. 

 **Test 3**
1. Run `zlil configure` and make sure that you have logs existing in your daemon and zli dev files to post to S3.
2. Run `zli send-logs --all [some_agent_target]`.
3. Follow steps 3 and 4 listed for Test 1
7. Download the zip files, unzip them and expect to see the **zli, daemon, and agent** logs, timestamped with the last two days. This command is essentially a combination of test 1 and 2. You can expect to see the zip file for agent logs and zip file for zli and daemon logs grouped under the same folder labeled with the UUID for this request. 

**backend branch:** send-logs-to-S3
**bzero branch:** develop
**charts branch:**

- [x] Yes

Release Notes: Allows users to send logs to BastionZero

## Related JIRA tickets

Relates to JIRA: CWC-2024

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/system-tests/tests/suites/send-logs.ts
```diff
@@ -0,0 +1,115 @@
...
+        // build the bucket name
+        const bucket = `bastionzero-${configService.getConfigName()}-customer-logs`;
+
+        beforeEach(() => {
+            const bzLoggerFolderPath = path.join(process.env.HOME, 'Library', 'Preferences', 'bastionzero-logger-nodejs');
+            if(!fs.existsSync(`${bzLoggerFolderPath}/bastionzero-daemon-dev.log`)) {
+                // manually adding this file since CustomReporter deletes the daemon-dev file
+                const toWrite = 'Test message to write to file';
+                try {
+                    fs.writeFileSync(`${bzLoggerFolderPath}/bastionzero-daemon-dev.log`, `${toWrite}\n`);
```
SaiDadireddy, line 46: This doesn't work for the PR system tests I assume because of the path, will look into fixing this today.

package.json
```diff
@@ -39,12 +39,15 @@
...
     "@jest/test-result": "^27.5.1",
     "@kubernetes/client-node": "^0.16.1",
+    "@types/adm-zip": "^0.5.0",
     "@types/humanize-duration": "^3.27.1",
     "@types/jest": "^27.5.0",
     "@types/pg": "^8.6.5",
     "@types/uuid": "^8.3.4",
     "@typescript-eslint/eslint-plugin": "^4.17.0",
     "@typescript-eslint/parser": "^4.17.0",
+    "adm-zip": "^0.5.9",
```
lipmas, line 49: hm shouldnt this be in `dependencies` instead of `devDependencies` ?

src/http-services/targets/bzero/bzero.http-services.ts
```diff
@@ -29,4 +28,8 @@ export class BzeroTargetHttpService extends HttpService {
     public RestartBzeroTarget(request: RestartBzeroAgentTargetRequest): Promise<void> {
         return this.Post(`restart`, request);
     }
-}
\ No newline at end of file
+
+    public UploadBzeroTargetLogs(request: RetrieveBzeroAgentLogsRequest): Promise<void> {
```
lipmas, line 32: try to make the naming consistent - this should probably be named similarly to the request `RetrieveBzeroAgentLogs`

src/system-tests/tests/suites/send-logs.ts
lipmas: You should be able to use the `LoggerConfigService` [which is exported in system-test.ts](https://github.com/bastionzero/zli/blob/eb12e42280d36d4fa6cb16e79527dbe61ee834e8/src/system-tests/tests/system-test.ts#L61) to get the [log file paths ](https://github.com/bastionzero/zli/blob/b243f1e67c4a3ce4c27bdb8525056d4398fc7086/src/services/logger/logger-config.service.ts#L62-L68) for both zli/daemon.

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        logger.warn(`Triggering ${parsedTarget.name} to send agent logs to BastionZero!`);
+    }
+
+    // if the target flag is undefined then we will send zli and daemon logs
+    // target flag indicates only agent logs will be sent to S3
+    if(argv.target === undefined) {
+        const uploadLogsHttpService = new UploadLogArchiveHttpService(configService, logger);
+        const bzLoggerFolderPath = path.join(process.env.HOME, 'Library', 'Preferences', 'bastionzero-logger-nodejs');
```
lipmas, line 101: can use `loggerConfigService.configPath()` for this base path

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        const zliLogPath = loggerConfigService.logPath();
+        const daemonLogPath = loggerConfigService.daemonLogPath();
+        const { currDateZli, filteredZliContents, filteredDaemonContents } = await getFilteredLogContents(logger, zliLogPath, daemonLogPath);
+
+        const tempZliFilePath = `${bzLoggerFolderPath}/bastionzero-zli-${currDateZli}.log`;
+        const tempDaemonFilePath = `${bzLoggerFolderPath}/bastionzero-daemon-${currDateZli}.log`;
+        const tempZipFilePath = `${bzLoggerFolderPath}/log-archive.zip`;
```
lipmas, line 109: just in case deleting fails i think its also a good idea to include requestId in the temp file names.

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,173 @@
...
+            await bzeroTargetService.UploadBzeroTargetLogs({
+                targetName: parsedTarget.name,
+                targetId: parsedTarget.id,
+                envId: parsedTarget.envId,
+                envName: parsedTarget.envName,
+                uploadLogsRequestId: uploadLogsRequestId
+            });
+        } catch (error) {
+            logger.error(error);
+            await cleanExit(1, logger);
```
lipmas, line 91: why cleanExit here instead of failing through to still sending zli/daemon logs

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        // delete the two temporary log files
+        // delete them before posting in case of server error
+        try {
+            fs.unlinkSync(tempZliFilePath);
+            fs.unlinkSync(tempDaemonFilePath);
+        } catch(err) {
+            logger.error(`Error deleting temporary files: ${err}`);
+            await cleanExit(1, logger);
+        }
```
lipmas, line 143: id move this + zip file delete into a helper function and call it both in the try and catch block after sending the file before cleanExit

package.json
SaiDadireddy: Yep, originally I had it there because I thought I would be using it for system tests as well, but that's not the case anymore. Removed here [3593227](https://github.com/bastionzero/zli/pull/535/commits/3593227d63368b9454b88b9b5c8f202f7e345bef)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: I understood the criteria as sending all three together but I agree that it's a good idea to send the zli/daemon anyways when the user uses the `--all` flag and runs into an error with agent logs like the specified target does not exist. [7863b33](https://github.com/bastionzero/zli/pull/535/commits/7863b331d526730f281bf449a1c7ae1f8dca82e4)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: [7863b33](https://github.com/bastionzero/zli/pull/535/commits/7863b331d526730f281bf449a1c7ae1f8dca82e4)

src/http-services/targets/bzero/bzero.http-services.ts
SaiDadireddy: [7863b33](https://github.com/bastionzero/zli/pull/535/commits/7863b331d526730f281bf449a1c7ae1f8dca82e4)

src/system-tests/tests/suites/send-logs.ts
SaiDadireddy: [74c85c3](https://github.com/bastionzero/zli/pull/535/commits/74c85c36dc2d22a1d5555f01b1aebcd7b2667adf)

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+        try {
+            readStream = fs.createReadStream(tempZipFilePath);
+        } catch(err) {
+            logger.error(`Error creating read stream from the zip file: ${err}`);
+            await cleanExit(1, logger);
+        }
+
+        // post zli and daemon logs directly to UploadLogArchiveController
+        try {
+            const uploadLogArchiveResponse = await uploadLogsHttpService.UploadLogArchive({
```
johncmerfeld, line 137: Why do we need this boolean? Won't the API return a 200 if it succeeded and an error otherwise?

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+                userEmail: configService.me().email,
+                uploadLogsRequestId: uploadLogsRequestId,
+                logArchiveZip: readStream,
+                logType: 'zlidaemon'
+            });
+
+            if(uploadLogArchiveResponse.success) {
+                logger.warn('Zli and daemon logs have been sent to BastionZero!');
```
johncmerfeld, line 145: Let's use `logger.info` here instead

src/system-tests/tests/suites/send-logs.ts
```diff
@@ -0,0 +1,115 @@
...
+
+                const objectPrefix = `${month}/${emailDomain}/${userEmail}/${generatedUuid}/`;
+
+                const opts = { Bucket: bucket, Prefix: objectPrefix };
+                let allKeys: string[] = [];
+                try {
+                    allKeys = await getAllKeys(opts, []);
+                } catch(err) {
+                    console.log(err);
+                }
```
johncmerfeld, line 105: Why catch this? Shouldn't the test fail if `getAllKeys` fails? i.e. this block can just be:
```ts
const allKeys = await getAllKeys(opts, []);
```

src/system-tests/tests/suites/send-logs.ts
```diff
@@ -0,0 +1,115 @@
...
+
+            const objectPrefix = `${month}/${emailDomain}/${userEmail}/${generatedUuid}/`;
+
+            const opts = { Bucket: bucket, Prefix: objectPrefix };
+            let allKeys: string[] = [];
+            try {
+                allKeys = await getAllKeys(opts, []);
+            } catch(err) {
+                console.log(err);
+            }
```
johncmerfeld, line 76: Same as below

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+            fs.unlinkSync(tempDaemonFilePath);
+            fs.unlinkSync(tempZipFilePath);
+        } catch(err) {
+            logger.error(`Error deleting temporary files: ${err}`);
+            await cleanExit(1, logger);
+        }
+    }
+
+    if(agentLogsSent || zlidaemonLogsSent) {
+        logger.warn(`Unique identifier for this request: ${uploadLogsRequestId}`);
```
johncmerfeld, line 164: `logger.info` here too, I think

Do we need to give them more information on what to do with this unique identifier?

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+                logArchiveZip: readStream,
+                logType: 'zlidaemon'
+            });
+
+            if(uploadLogArchiveResponse.success) {
+                logger.warn('Zli and daemon logs have been sent to BastionZero!');
+                zlidaemonLogsSent = true;
+            }
+        } catch(err) {
+            logger.error(err);
```
johncmerfeld, line 149: Here is where I would have thought we'd want to exit on error, because presumably the upload failed

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+    return { currDateZli, prevDateZli, currDateDaemon, prevDateDaemon };
+}
+
+async function getFilteredLogContents(logger: Logger, zliLogFilePath: string, daemonLogFilePath: string) {
+    let zliLogContents: string[];
+    let daemonLogContents: string[];
+    try {
+        zliLogContents = fs.readFileSync(zliLogFilePath, 'utf-8').split('\n');
+        daemonLogContents = fs.readFileSync(daemonLogFilePath, 'utf-8').split('\n');
+    } catch (err) {
```
johncmerfeld, line 44: I think we need to split this try/catch into two separate blocks for better UX. If there's a problem with just one of the files, it's better to be more specific with our user-facing error messages so they can more easily solve the problem themselves

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+import { cleanExit } from '../clean-exit.handler';
+import { UploadLogArchiveHttpService } from '../../http-services/upload-logs/upload-log-archive.http-service';
+import { BzeroTargetHttpService } from '../../http-services/targets/bzero/bzero.http-services';
+import { sendLogsArgs } from './send-logs.command-builder';
+import { parseTargetString } from '../../utils/utils';
+import { LoggerConfigService } from '../../services/logger/logger-config.service';
+
+export function getDates() {
+    const currDate = new Date();
+    const prevDate = new Date();
```
johncmerfeld, line 17: In bzero, we call these `today` and `yesterday`. Just sayin'

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    // zli timestamp is formatted 2006-01-01
+    const currDateZli = currDate.toISOString().slice(0, 10);
+    const prevDateZli = prevDate.toISOString().slice(0, 10);
+
+    // daemon timestamp is formatted Jan  1
+    let currDateDaemon = currDate.toDateString().slice(4, 10);
+    let prevDateDaemon = prevDate.toDateString().slice(4, 10);
+
+    if(currDateDaemon[4] === '0') {
```
johncmerfeld, line 28: Comment would be helpful here. Why do we do this?

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+
+        try {
+            await bzeroTargetService.RetrieveBzeroTargetLogs({
+                targetName: parsedTarget.name,
+                targetId: parsedTarget.id,
+                envId: parsedTarget.envId,
+                envName: parsedTarget.envName,
+                uploadLogsRequestId: uploadLogsRequestId
+            });
+            logger.warn(`Triggering ${parsedTarget.name} to send agent logs to BastionZero!`);
```
johncmerfeld, line 90: I think you can replace `warn` with `info` and replace "Triggering" with "Requesting"

src/http-services/targets/bzero/bzero.http-services.ts
```diff
@@ -29,4 +28,8 @@ export class BzeroTargetHttpService extends HttpService {
     public RestartBzeroTarget(request: RestartBzeroAgentTargetRequest): Promise<void> {
         return this.Post(`restart`, request);
     }
-}
\ No newline at end of file
+
+    public RetrieveBzeroTargetLogs(request: RetrieveBzeroAgentLogsRequest): Promise<void> {
+        return this.Post(`upload-logs`, request);
```
johncmerfeld, line 33: See my comment in webshell-backend but IMO this should be "retrieve-logs" to distinguish it from the actual upload endpoint (and because that's what we call it everywhere else in the code)

src/handlers/send-logs/send-logs.command-builder.ts
```diff
@@ -0,0 +1,30 @@
...
+                description: 'Triggers only target logs to be sent to BastionZero',
+                alias: 't'
+            }
+        )
+        .option('all',
+            {
+                type: 'string',
+                demandOption: false,
+                requiresArg: true,
+                description: 'Triggers zli, daemon and target logs to be sent to BastionZero',
```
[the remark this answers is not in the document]
SaiDadireddy: [738edf6](https://github.com/bastionzero/zli/pull/535/commits/738edf679a8e46a0cf014f3938c048a9f433f1ed)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: [738edf6](https://github.com/bastionzero/zli/pull/535/commits/738edf679a8e46a0cf014f3938c048a9f433f1ed)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: [738edf6](https://github.com/bastionzero/zli/pull/535/commits/738edf679a8e46a0cf014f3938c048a9f433f1ed)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: You're right actually, this isn't needed. I removed it here [738edf6](https://github.com/bastionzero/zli/pull/535/commits/738edf679a8e46a0cf014f3938c048a9f433f1ed). 

This is the corresponding webshell-common change: [bd7a342](https://github.com/bastionzero/webshell-common-ts/pull/90/commits/bd7a34289201551047ccc9f114341ba15fbdb91e)
This is the corresponding backend change: [ae42376](https://github.com/bastionzero/webshell-backend/pull/1318/commits/ae42376524c6e5344754abe9b8bb825f110e1d06)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: I don't think so. We will cover this in the documentation. Returning the identifier is more of just an FYI in case they want to keep track or there is a reason for them to send it to us. [738edf6](https://github.com/bastionzero/zli/pull/535/commits/738edf679a8e46a0cf014f3938c048a9f433f1ed)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: [738edf6](https://github.com/bastionzero/zli/pull/535/commits/738edf679a8e46a0cf014f3938c048a9f433f1ed)

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: I don't see how that's possible. In order for the request id to be shown to the user, atleast one of the two requests will need to succeed and either `agentLogsSent` or `zlidaemonLogsSent` will need to be set to true. They're set to true only after the corresponding request succeeds. If neither request succeeds, then these two boolean values will remain false and the request id is not shown to the user.

src/handlers/send-logs/send-logs.handler.ts
SaiDadireddy: I agree. Change made here: [91ee1ef](https://github.com/bastionzero/zli/pull/535/commits/91ee1eff678e4d734066cbb5a078ad66818af4d6)

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // delete the temporary files
+        try {
+            fs.unlinkSync(tempZliFilePath);
+            fs.unlinkSync(tempDaemonFilePath);
+            fs.unlinkSync(tempZipFilePath);
+        } catch(err) {
+            logger.error(`Error deleting temporary files: ${err}`);
+            await cleanExit(1, logger);
```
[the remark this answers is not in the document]
SaiDadireddy: I agree, at this point the request could succeed so it should just be a warning. Change made here: [91ee1ef](https://github.com/bastionzero/zli/pull/535/commits/91ee1eff678e4d734066cbb5a078ad66818af4d6)

src/handlers/send-logs/send-logs.handler.ts
johncmerfeld: Ah true, forgot about the check on those variables.

johncmerfeld (approved):
This side of the house looks good to me

src/system-tests/tests/system-test-setup.ts
```diff
@@ -136,11 +137,24 @@ export async function setupDOTestCluster(): Promise<RegisteredDigitalOceanKubern
...
+        const upCommand = `git clone -b ${chartsBranch} https://github.com/bastionzero/charts.git`;
+        const pexec = promisify(exec);
+        await pexec(upCommand);
 
-    // install bastionzero helm chart
-    const helmChart = 'bastionzero/bctl-quickstart';
+        // install helm chart from custom branch
+        helmChart = './charts/charts/bctlquickstart';
+    } else {
+        // clone the charts repo at a particular branch
```
SaiDadireddy, line 151: I'll remove this

PR #548, opened 2022-10-03
killDaemon: block until process is no longer running

johncmerfeld:
## Description of the change

We currently have a race condition in code that calls `killDaemon`. Now that we kill with `-2` (SIGINT) instead of `-9` (SIGKILL), the daemon may take a few seconds to shut down gracefully. However, if `killDaemon` is called in order to free up a port for the next connection, it will start that connection before the previous one has exited, causing the second connection to fail.

Also do some refactoring to consolidate code in the `ProcessManagerService` and remove a dependency I think we're better off just writing out.

On `develop`: 
```sh
# this works
zli connect cluster-admin@bzdemo-cluster-y
# then this fails
zli connect cluster-admin@demo-cluster
```

On this branch: 
```sh
# this works
zli connect cluster-admin@bzdemo-cluster-y
# this also works, after a delay of a couple seconds
zli connect cluster-admin@demo-cluster
```

## Testing

**backend branch:** develop
**bzero branch:** develop

- [x] Yes

Release Notes: Fix bug that caused sequential kube connections to fail

## Related JIRA tickets

Relates to JIRA: CWC-2140

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


src/services/process-manager/process-manager.service.ts
```diff
@@ -22,6 +26,23 @@ export class ProcessManagerService {
      * @returns True if the process is still running. False otherwise
      */
     public isProcessRunning(pid: number): boolean {
-        return isRunning(pid);
+        try {
+            return process.kill(pid, 0);
```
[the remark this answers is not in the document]
johncmerfeld: Yeah it seems to be the least-disliked way to check that a process is running

src/services/process-manager/process-manager.service.ts
```diff
@@ -22,6 +26,23 @@ export class ProcessManagerService {
...
+    /**
+     * Block until a given process is no longer running. Suitable for use with general OS processes.
+     * To wait for a daemon child process started in the current execution, use daemon-utils.ts:`waitForDaemonProcessExit`
+     * @param pid Process's PID to wait for
+     */
+    public async waitForProcess(pid: number) {
+        while (this.isProcessRunning(pid)) {
+            // wait half a second and check again
+            await new Promise(r => setTimeout(r, 500));
+        }
```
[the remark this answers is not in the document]
johncmerfeld: I toyed with that idea but couldn't decide what to do if the timeout expired. Throwing an error seems like the right call though

ymarcus93:
(Yuval On mobile) Don't forget about DaemonManagementService (which covers db and soon other daemon types). Disconnect daemons() method there calls kill Daemon method of process manager service. So it needs to call the new wait method as well in order to get this same logic applied for db.

There's an open ticket to remove killDaemon from that utils file once kube and web migrate over to DaemonManagementService.

src/services/process-manager/process-manager.service.ts
```diff
@@ -22,6 +36,28 @@ export class ProcessManagerService {
...
+    /**
+     * Block until a given process is no longer running. Suitable for use with general OS processes.
+     * To wait for a daemon child process started in the current execution, use daemon-utils.ts:`waitForDaemonProcessExit`
+     * @param pid Process's PID to wait for
+     */
+    public async waitForProcess(pid: number, timeout: number = 10000) {
+        setTimeout(() => {
+            const err = new Error(`process is still running after ${timeout} ms`);
+            err.name = 'TIMEOUT';
+        }, timeout);
```
lipmas, line 56: This isnt actually throwing any error is it? Woudlnt you need to set some variable and check this in the while loop and then throw an error for this to work?

The way i generally like to handle a global timeout is to use `Promise.race` to race the timeout against the other promise we want to complete or timeout. See example in test-utils `waitForExpect`:
https://github.com/bastionzero/zli/blob/eb12e42280d36d4fa6cb16e79527dbe61ee834e8/src/system-tests/tests/utils/test-utils.ts#L319-L340

johncmerfeld: Gah, yep. This code underwent a couple changes and the throw got lost. I'm happy to do it the other way though. I wish there was just a Go-like `After` check to use

lipmas: Actually i think that function i linked is general purpose enough it could just be used here.

johncmerfeld: Ultimately decided to roll my own. Moving the one you linked from system-tests to utils affected a lot of other imports and I think the way I wrote it expresses more clearly what's happening here.

src/services/process-manager/process-manager.service.ts
```diff
@@ -45,19 +48,26 @@ export class ProcessManagerService {
...
-        }, timeout);
+            reject(err);
+        });
 
-        while (this.isProcessRunning(pid)) {
-            // wait half a second and check again
-            await new Promise(r => setTimeout(r, 500));
-        }
+        const processDone = new Promise<void>(async (resolve, _) => {
+            while (this.isProcessRunning(pid)) {
```
lipmas, line 64: This while loop would keep running even after the timeout though. You should add a local `done` variable to the `waitForProcess` function scope and sent it to true in the `gloablTimeout`. Then this can be:
```
while (this.isProcessRunning(pid) && !done)
```

johncmerfeld: Ah, I see. Thanks, I'm not very versed in promises.

ff1d077

lipmas: you have to move `done = true` after the `await new Promise(r => setTimeout(r, timeout));` line though -just put it right before `reject()`

johncmerfeld: Pair programming by others means 😆 
8a7b777

lipmas (approved):

PR #549, opened 2022-10-03
Fix shell terminal resize

lipmas:
## Description of the change

Dont start the shell daemon as a detached subprocess so it can still receive signals from the parent process. This is needed in order for the daemon to receive SIGWINCH signals when the parent terminal window is resized. This is now consistent with how ssh starts the daemon process [here](https://github.com/bastionzero/zli/blob/c19737c9c24e592c1ecd857289933d3a5c25ec67/src/handlers/ssh-proxy/ssh-proxy.handler.ts#L154-L162).

> When using the detached option to start a long-running process, the process will not stay running in the background after the parent exits unless it is provided with a stdio configuration that is not connected to the parent. If the parent's stdio is inherited, the child will remain attached to the controlling terminal.

As described in the [nodejs docs](https://nodejs.org/api/child_process.html#child_process_options_detached) starting the shell process as a detached process but specifying stdio `inherit` was still not allowing it to run in the background if the parent process was killed anyway so it was pointless for us to start in detached mode anyway.

## Testing

Create a new shell connection and record the current rows/cols of the terminal with `tput lines` and `tput cols` or with `stty -a`. Then try resizing the terminal and the rows/cols should update.

**backend branch:** 
**bzero branch:** develop

- [x] Yes

Release Notes: Fixes terminal resizing for bzero shell connections.

## Related JIRA tickets

Relates to JIRA: CWC-2110

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


lgmugnier (approved):
works for me!

johncmerfeld (approved):
Works as advertised

PR #551, opened 2022-10-06
Service accounts [CWC-2056]

tflpd:
## Description of the change

This PR adds support for service accounts. Service accounts can be used as regular user and enable headless log in to BastionZero. This way we enable machine-to-machine communication for BastionZero customers. The way this is achieved - without compromising our trustless model - is by allowing the customer to bring a custom trust root that they control. In essence, the customer provides a URL that serves a public Json Web Key. The customer also holds the respective private key and uses that to sign their tokens. Meanwhile, our backend provides a second trust root with an MFA (TOTP) secret.

Related PRs:
[backend](https://github.com/bastionzero/webshell-backend/pull/1369)
[common](https://github.com/bastionzero/webshell-common-ts/pull/94)
[agent](https://github.com/bastionzero/bzero/pull/183)

## Testing

- Contact @tflpd to provide a JWK
- Create a service account using `zli service-account create <provider-file.json>`
- Login with this account using `zli service-account login --providerCreds /path/to/provider-file.json --bzeroCreds /path/to/bzeroCreds.json`
- Try and take any action you would take as a regular user
- Newly registered targets support all service accounts that had been created before their registration. Existing targets need to be set up with `zli service-account configure -a <service-account-username>@<service-account-domain> -t <target-name(s), delimited with spaces>` before you are able to connect to them using a service account

**backend branch:** feat/service-accounts
**bzero branch:** develop
**charts branch:** feat/service-accounts

- [x] Yes

Release Notes: Service accounts

## Related JIRA tickets

Relates to JIRA: CWC-2056 CWC-2220 CWC-2185 CWC-2177 

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain:
This enables a whole new way to auth against BastionZero. It should be thoroughly checked to ensure it cannot be exploited.


src/services/config/config.service.ts
```diff
@@ -263,16 +263,17 @@ export class ConfigService implements ConfigInterface {
...
+    public me(): SubjectSummary
+    {
         const whoami = this.config.get('whoami');
         if (whoami) {
             return whoami;
         } else {
-            throw new Error('User information is missing. You need to log in, please run \'zli login --help\'');
+            throw new Error('Subject information is missing. You need to log in, please run \'zli login --help\'');
         }
     }
```
lipmas, line 274: Because this is changing from `UserSummary` to `SubjectSummary` and the only difference is the ` type: SubjectType` property existing logged in zli users will not have this `type` field set when they upgrade. We could however fix that by creating a conf migration that updates the `whoami` config key to include `type: User`.

src/http-services/service-account/service-account.http-services.ts
```diff
@@ -0,0 +1,70 @@
...
+        super(configService, 'api/v2/service-accounts', logger);
+    }
+
+    public async LoginServiceAccount(req: LoginServiceAccountRequest): Promise<ServiceAccountSummary> {
+        const requestHeaders = {
+            'AccessToken': this.configService.getAccessToken(),
+            'IdToken': this.configService.getIdToken(),
+        };
+
+        // TODO fix this, add the mfa code here
```
lipmas, line 28: this TODO is no longer accurate

src/services/config/config.service.ts
tflpd: Hm, so something like this, right?
765897e2

src/http-services/service-account/service-account.http-services.ts
tflpd: 6f31707a

src/services/config/config.service.ts
lipmas: I think you wanted to do `if(!currentSubject.type)` there instead but otherwise this is good.

Also why using `!!!` instead of just `!`? I have noticed this in  a couple places in this PR and was going to make other comments but imo this is just really bad for readability and we should never do this. The only reason to do this is if you want to coerce the type a boolean before testing if its false/true. But `undefined` and `null` types are already considered [falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) so i dont see any reason to actually do this?

src/system-tests/tests/suites/service-account.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ServiceAccountBzeroCredentials } from '../../../../src/handlers/login/types/service-account-bzero-credentials.types';
+import { bzeroCredsPath, logger, providerCredsPath, RUN_AS_SERVICE_ACCOUNT } from '../system-test';
+import { ServiceAccountProviderCredentials } from '../../../../src/handlers/login/types/service-account-provider-credentials.types';
+import { Logger } from '../../../../src/services/logger/logger.service';
+
+export const serviceAccountSuite = () => {
+    describe('Service Account Suite', () => {
+        // test successfully creating a service account
+        it('481485: Create service account as user', async () => {
+            if (!RUN_AS_SERVICE_ACCOUNT) {
```
lipmas, line 12: Seems like we should just not run the `serviceAccountSuite` if `RUN_AS_SERVICE_ACCOUNT` is true. So we could just it https://github.com/bastionzero/zli/blob/62b0ed6b45c0666f25c9fb8de0607fc4ceb1b759/src/system-tests/tests/system-test.ts#L323 instead.

src/system-tests/tests/suites/service-account.ts
```diff
@@ -0,0 +1,50 @@
...
+                const bzeroCredsFile = JSON.parse(fs.readFileSync(bzeroCredsPath, 'utf-8')) as ServiceAccountBzeroCredentials;
+                await callZli(['service-account', 'rotate-mfa', providerCredsPath]);
+                const editedBzeroCredsFile = JSON.parse(fs.readFileSync(bzeroCredsPath, 'utf-8')) as ServiceAccountBzeroCredentials;
+                expect(editedBzeroCredsFile).toBeTruthy();
+                expect(editedBzeroCredsFile.mfa_secret).toBeTruthy();
+                expect(bzeroCredsFile.mfa_secret).not.toEqual(editedBzeroCredsFile.mfa_secret);
+            } else {
+                logger.info(`Skipping mfa rotation of service account as RUN_AS_SERVICE_ACCOUNT is set to true`);
+            }
+        }, 60 * 1000);
```
lipmas, line 34: For both of these tests it feels like we should try and login to the SA to verify its working. We could call login with a custom `configDir` passed as an option to the zli so we dont interfere with the currently logged in role account user.

src/services/config/config.service.ts
tflpd: 11c70998
So `!!!` is supposed to enhance readability by signaling that we are using as a boolean a non-boolean type.
https://stackoverflow.com/a/57828405/9420697

That being said, if you feel like it is making readability worse that beats the whole purpose of using it so I will stop introducing it and start removing it wherever I see it

src/system-tests/tests/suites/service-account.ts
tflpd: 0fbdccb3

src/system-tests/tests/suites/service-account.ts
tflpd: Do we use a custom config somewhere else already?

src/handlers/login/login.handler.ts
```diff
@@ -45,83 +60,186 @@ function interactiveResetMfa(): Promise<string> {
     });
 }
 
-export async function login(mrtapService: MrtapService, configService: ConfigService, logger: Logger, mfaToken?: string): Promise<LoginResult | undefined> {
+export async function loginSubject(mrtapService: MrtapService, configService: ConfigService, logger: Logger, isServiceAccount: boolean, mfaToken?: string, providerCredsPath?: string, bzeroCredsPath?: string): Promise<LoginResult | undefined> {
```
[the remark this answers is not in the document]
tflpd: Agreed
a2e06a53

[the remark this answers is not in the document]
tflpd: aand this 6e0da7e5

src/handlers/login/login.handler.ts
```diff
@@ -45,83 +60,186 @@ function interactiveResetMfa(): Promise<string> {
...
+            this.logger.error(`Service account ${serviceAccountSummary.email} is not currently enabled.`);
+            await cleanExit(1, this.logger);
+        }
+        const subjectHttpService = new SubjectHttpService(configService, logger);
+        const me = await subjectHttpService.Me();
+        configService.setMe(me);
+
+        // clear temporary SSH files
+        removeIfExists(configService.sshKeyPath());
+        removeIfExists(configService.sshKnownHostsPath());
```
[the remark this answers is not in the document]
tflpd: I am not sure, @johncmerfeld added this, maybe there is a reason it need to be done there?

src/services/config/config.service.ts
lipmas: yeah i dont agree with that SO...for me at least its not very intuitive and hurts readability.

src/handlers/login/login.handler.ts
```diff
@@ -45,83 +60,186 @@ function interactiveResetMfa(): Promise<string> {
...
+    const signedIdToken = jwt.sign(idToken, providerCredsFile.private_key, JWTOptions);
+    const signedAccessToken = jwt.sign(accessToken, providerCredsFile.private_key, JWTOptions);
+
+    return new TokenSet({
+        access_token: signedAccessToken.toString(),
+        token_type: 'Bearer',
+        id_token: signedIdToken.toString(),
+        refresh_token: '',
+        expires_in: idToken.exp,
+        expires_at: 2147483647, // Is this needed?
```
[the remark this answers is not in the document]
tflpd: Just adjusted these two lines to follow what we have for the rest idps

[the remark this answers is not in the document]
tflpd: 220346a0

src/utils/utils.ts
```diff
@@ -1026,4 +1134,25 @@ export function isZliSilent(silent_flag: boolean, json_flag: boolean) {
...
     return str.charAt(0).toUpperCase().concat(str.slice(1));
+}
+
+/**
+ * Creates a json file with bzero credentials in the current path.
+ */
+export function createBzeroCredsFile(mfaSecret: string, orgId: string, idp: IdentityProvider ) {
+    const bzeroCreds: ServiceAccountBzeroCredentials = {
+        mfa_secret: mfaSecret,
+        org: orgId,
```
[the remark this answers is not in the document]
tflpd: a4a1baf1

src/system-tests/tests/suites/service-account.ts
tflpd: Ah, I see okay okay
9ae27b64

src/services/config/config.service.ts
tflpd: a1017561

src/handlers/login/login.handler.ts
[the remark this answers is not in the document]
lipmas: Think this is still wrong we should be using `expires_at` not `expires_in`. At least that is what i have set in my zli config file `tokenSet`. Also `accessToken.exp` is an absolute timestamp so wouldnt make sense for `expires_in` (but  i think we can just leave that one undefined).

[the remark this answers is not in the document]
tflpd: Yes, meant to do the opposite, meant to use `expires_at` there an remove `expires_in` as it is optional and in case of a client whose clock is not synced is not really working great, my bad, thanks 🖤
95e673f3

src/utils/utils.ts
```diff
@@ -1026,4 +1134,25 @@ export function isZliSilent(silent_flag: boolean, json_flag: boolean) {
...
+ */
+export function createBzeroCredsFile(mfaSecret: string, issuer: string, idp: IdentityProvider ) {
+    const bzeroCreds: ServiceAccountBzeroCredentials = {
+        mfa_secret: mfaSecret,
+        issuer: issuer,
+        identity_provider: idp
+    };
+
+    // TODO : Maybe add error handling lambda here
+    fs.writeFileSync('./bzero-credentials.json', JSON.stringify(bzeroCreds));
```
lipmas, line 1150: The create service account command should probably take an output file argument of where to write the bzero credentials file

tflpd: Makes sense, adding it

tflpd: 81cd064d

src/handlers/policy/policy-create/create-policy.command-builder.ts
```diff
@@ -45,12 +46,23 @@ function baseCreatePolicyCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<baseCrea
...
+            {
+                type: 'string',
+                array: true,
+                demandOption: false,
+                requiresArg: true,
+                alias: 'a',
+                description: 'BastionZero subjects (IdP users and service accounts) the policy applies to (SSO emails)'
             }
         )
+        .conflicts('users', 'subjects')
```
lipmas, line 65: How about instead of defining 2 options we just create aliases to keep this backwards compatible with a deprecation warning in yargs. So this instead:

```ts
.option('subjects',
    {
        type: 'string',
        array: true,
        demandOption: true,
        requiresArg: true,
        alias: ['u', 'users'],
        description: 'BastionZero subjects (IdP users and service accounts) the policy applies to (SSO emails)'
    }
)
.deprecateOption('subjects', 'use --subjects instead of --users')
```

src/handlers/policy/policy-create/create-proxy-policy.handler.ts
```diff
@@ -30,25 +28,45 @@ export async function createProxyPolicyHandler(argv: yargs.Arguments<createProxy
         await cleanExit(1, logger);
     }
 
-    const users: Subject[] = await getUsersByEmail(argv.users, userHttpService, logger);
+    let subjectsEmails: string[];
+    if(argv.users) {
+        this.logger.warn('The users flag is deprecated and will be removed soon, please use its equivalent \'subjects\'');
+        subjectsEmails = argv.users;
+    } else
+        subjectsEmails = argv.subjects;
```
lipmas, line 36: Its annoying that we have to keep repeating this code in every handler. See https://github.com/bastionzero/zli/pull/551/files#r1028568850 but if we only had a single option value with aliases this becomes much simpler

src/utils/utils.ts
lipmas: Can we just use `default: ./bzeroCreds.json` and always just use this argument instead

src/handlers/login/login.handler.ts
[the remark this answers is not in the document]
lipmas: Talked to @johncmerfeld and he said it should be fine to move this to `configService.logout();`

src/utils/policy-utils.ts
```diff
@@ -77,146 +91,145 @@ async function parseUniqueResource<T>(resourceNames: string[], allResources: T[]
...
+        clustersIdentifierMap = mapTargetsByNameOrId(targetIdentifiers, clusters);
+    }
+
+    return [
+        bzeroTargetsIdentifierMap,
+        dacTargetsIdentifierMap,
+        dbTargetsIdentifierMap,
+        webTargetsIdentifierMap,
+        clustersIdentifierMap
+    ];
```
lipmas, line 202: Having this return separate dictionaries for each target type is unnecessary and makes this code and the code that uses it more complex and harder to understand. We should be able to construct a single dictionary with key as (name or id) and value TargetSummary. This function can still filter for certain target types but it can just combine them in a single dictionary.

I think this code would do it:

```
const allTargets = mapTargetsByNameOrId(targetIdentifiers, targetTypes.map(t => targetsPerType[t]).reduce((acc, current) => [...acc, ...current], []);
return allTargets;
```

src/utils/policy-utils.ts
```diff
@@ -77,146 +91,145 @@ async function parseUniqueResource<T>(resourceNames: string[], allResources: T[]
...
-        if(!allDynamicNamesMap[t.name]) {
-            allDynamicNamesMap[t.name] = [{ id: t.id, type: TargetType.DynamicAccessConfig }];
-        } else {
-            allDynamicNamesMap[t.name].push({ id: t.id, type: TargetType.DynamicAccessConfig });
-        }
-        allDynamicIdsMap[t.id] = { id: t.id, type: TargetType.DynamicAccessConfig };
-    });
-    const errorTargetType = 'base target or dynamic access config';
-    return await parseMultipleTargetTypes(targetNames, allBzeroNamesMap, allBzeroIdsMap, allDynamicNamesMap, allDynamicIdsMap, errorTargetType, logger);
+    return flattenTargetMaps(targetIdentifiers, allTargetNamesMap, allTargetIdsMap);
```
lipmas, line 217: dont quite follow why we have to have a separate `flattenTargetMaps` function. Whats wrong with just pushing both name/id into the map at the same time:

```ts
const allTargetIdentifiersMap: {[id:string]: Target} = {};
allTargets.forEach(t => {
    if(!allTargetIdentifiersMap[t.name]) {
        allTargetIdentifiersMap[t.name] = [{ id: t.id, type: t.type }];
    } else {
        allTargetIdentifiersMap[t.name].push({ id: t.id, type: t.type });
    }
    allTargetIdentifiersMap[t.id] = { id: t.id, type: t.type };
});
```

src/services/list-targets/list-targets.service.ts
```diff
@@ -167,7 +192,21 @@ export async function listTargets(
...
+            const versionFilteredTargets = targets
+                .filter(
+                    t => configService.getConfigName() != 'prod' ||
+                                                    (configService.getConfigName() == 'prod' &&
+                                                        (t.type == TargetType.DynamicAccessConfig ||
+                                                        gt(parse(t.agentVersion), new SemVer(minAgentVersion)))
+                                                    )
+                );
+            targetsPerType[targets[0].type] = versionFilteredTargets;
+        }
```
lipmas, line 209: Few things:

1. Shouldnt this be `if(targets.length > 0)`?
2. Why are we ignoring DATs?
3. Its probably better if this is separated into its own function to filter a `TargetSummary[]`. Not everyone who calls this method cares about filtering agent version so its its better in general to compose filters like this only when they are needed.

src/utils/utils.ts
tflpd: 93b8d0f8

src/handlers/policy/policy-create/create-policy.command-builder.ts
tflpd: Will run this by Product, this has a side effect of showing a deprecate notice when a user is using the not deprecated option `subjects`

src/utils/policy-utils.ts
```diff
@@ -77,146 +91,145 @@ async function parseUniqueResource<T>(resourceNames: string[], allResources: T[]
...
-            allClusterNamesMap[c.name] = [{ id: c.id }];
-        } else {
-            allClusterNamesMap[c.name].push({ id: c.id });
+/**
+ * Checks that in the provided dictionary every identifier is mapped to a target
+ * @param targetsIdentifierMap A dictionary of target identifiers (targetName or targetId) -> target
+ */
+export async function checkAllIdentifiersExist<T extends Target>(logger: Logger, errorTargetType: string, targetsIdentifierMap: Dictionary<T[]>) {
+    for (const targetIdentifier in targetsIdentifierMap) {
+        if(!targetsIdentifierMap[targetIdentifier].length) {
```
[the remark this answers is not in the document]
tflpd: Every identifier is in the dictionary. If it has not been mapped to a target the identifier is just mapped to an empty array
https://github.com/bastionzero/zli/blob/4f41bf77fa5170692b718027927a4595bd078987/src/utils/policy-utils.ts#L227

src/handlers/policy/policy-create/create-policy.command-builder.ts
tflpd: Yup, since we already printed some misleading messages recently Product prefers to keep this as is

lipmas: We also dont have to add the `deprecateOption` bit

src/utils/policy-utils.ts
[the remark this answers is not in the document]
lipmas: Ok so it would make more sense to use `targetsIdentifierMap[targetIdentifier].length == 0` in that case. But also see my other related comments about the util methods because i think it would make more sense to return a dictionary with only the keys that exist and then this function could just check that the key is not undefined with `if(!targetsIdentifierMap[targetIdentifier])`

src/utils/policy-utils.ts
tflpd: The reason it is not unnecessary is that we need to later be able to apply restrictions based on the types (e.g. there should be only one bzero or one dac that maps to a specific identifier).
I recognize this is a complex method but it is so to be able to support all use cases without the code duplication we used to have.
We could indeed alleviate some complexity here doing what you suggest but we would have to add that some complexity later where we perform checks based on types combinations.
Specifically, we would have to create a checker function that goes through allTargets to ensure there is not more than one target per identifier (for clusters).
Then we would have to create a checker function that has as an argument two types and goes through allTargets to ensure there is not more than one target of type of one of the two passed types per identifier (for dac/bzero and for vt).
Then another 2 respective functions for no targets per identifier.
I am not convinced this will be much less complex, let me know if you had some other alternative in mind

src/utils/utils.ts
lipmas: Ok i also made the function async and resolved the error handling TODO: [ba0a8d6](https://github.com/bastionzero/zli/pull/551/commits/ba0a8d6c8f85a2de33c79a44879c528d275fc8c7)

src/utils/policy-utils.ts
tflpd: This would create a map of all the existing targets of this org. Instead, what we do, is that we create a map of the targets the user specified.
This comes handful later when we provide feedback to the user about their arguments. The user's org may have a naming conflict between targets but if the user has not specified any of the conflicting targets we should not prevent them from proceeding with their command.

That being said I think we could be iterating through the specified identifiers in the first place and for every identifier try to map it in one of allTargets but this sounds like it has worse complexity ( O(mn) vs O(m+n) ). I do not think this is a major difference so if you still believe it is worth it I will change it, I believe it is more readable as is

src/services/list-targets/list-targets.service.ts
tflpd: 1. Yes, it should c85d19b3

tflpd: 2. Because they do not have an agent version to filter on as far as I understand (?) https://github.com/bastionzero/zli/blob/e58398588584c1a917e06ba7afaeefd1d9be3150/src/utils/utils.ts#L1061

tflpd: 3. Yeah I am thinking about that to. I mean, it is an optional argument so not everyone has to use it. If we pulled out the filtering from here we would have to replicate this code or this new function so it felt that this is probably fine having it optional and achieving the least code duplication but I am very open to pull this out on its own function if you feel that's better

lipmas: 2. Ah i see because these are DACs where we dont have an agent version right ok. I would suggest we instead just rely on SemVer `parse` to return null if its not a valid version and instead use that to  check to handle this more generally: 

```
const agentVersion = parse(t.agentVersion);
...
agentVersion != null && gt(parse(agentVersion), new SemVer(minAgentVersion))
```

src/handlers/login/login.handler.ts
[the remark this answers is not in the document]
tflpd: 4f500539

src/handlers/policy/policy-create/create-policy.command-builder.ts
tflpd: But we want customers to stop using/thinking of the `--users` flag

src/services/list-targets/list-targets.service.ts
tflpd: fair fair
488a343c

src/utils/policy-utils.ts
lipmas: Sorry im not following where the extra complexity would come later can you point to a specific place in the code? It seems like for the most part we want to check that there isnt a single target name specified that has 0 or more than 1 target. Combining the targets into a single dictionary makes this check even simpler. For example [here](https://github.com/bastionzero/zli/blob/4f41bf77fa5170692b718027927a4595bd078987/src/handlers/policy/policy-create/create-tconnect-policy.handler.ts#L53-L56) we are already "merging" 2 dictionaries for bzero/DAT targets. But that would be unnecessary if we just returned an already merged dictionary. The code would just look like this:

```ts
const [targetIdentifierMap] = await getTargetsByNameOrId(configService, logger, [TargetType.Bzero, TargetType.DynamicAccessConfig], argv.targets);
checkAllIdentifiersExist(logger, 'bzero target or dynamic access config', targetIdentifierMap);
checkAllIdentifiersAreSingle(logger, 'bzero target or dynamic access config', targetIdentifierMap);
```

tflpd: I guess your point is that we will be controlling the type conflicts by the initial target type filtering which means that in the list that will be checked we will already have only the specified types. Yup, I think that's fair, you are right, adjusting this

src/utils/policy-utils.ts
lipmas: > This would create a map of all the existing targets of this org. Instead, what we do, is that we create a map of the targets the user specified.
This comes handful later when we provide feedback to the user about their arguments. The user's org may have a naming conflict between targets but if the user has not specified any of the conflicting targets we should not prevent them from proceeding with their command.

What im suggesting is constructing a map of all existing targets in the org and then iterating over the user specified targets to make sure they exist and dont have any naming conflicts (more than 1 target exists). This seems equivalent to me unless im missing something...and this approach doesnt require creating this extra dictionary with keys of the targets the user specified.

tflpd: `allTargets` are all the org targets, not all the targets the user specified, thus this code would not be iterating over the user specified targets, it would be iterating over all targets of the user's org (?)

src/handlers/service-account/configure-service-account.handler.ts
```diff
@@ -0,0 +1,111 @@
...
+    {
+        logger.error(`The provided subject ${argv.serviceAccount} is not a service account.`);
+        await cleanExit(1, logger);
+    }
+
+    // Get all targets if specified otherwise get just the specified ones
+    let targetIds: string[] = [];
+    let _unusedMaps: any;
+    if(argv.all) {
+        const targets = await listTargets(configService, logger, [TargetType.Bzero, TargetType.DynamicAccessConfig, TargetType.Cluster, TargetType.Db, TargetType.Web]);
```
lipmas, line 52: The only types of targets that make sense to configure here are `TargetType.Bzero` and `TargetType.Cluster`. DynamicAccessConfig are used to register bzero targets on demand. So there is no control channel to send to - and DATs will anyways always pickup the latest SA configuration because they are dynamically registered. Web/db targets are just virtual targets that proxy to an underlying bzero target so they also dont make sense here.

src/handlers/service-account/configure-service-account.handler.ts
```diff
@@ -0,0 +1,111 @@
...
+        let clustersIdentifierMap: Dictionary<Target[]> = {};
+
+        let mergedIdentifierMap: Dictionary<Target[]> = {};
+
+        [bzeroIdentifierMap, _unusedMaps, dbIdentifierMap, webIdentifierMap, clustersIdentifierMap] = await getTargetsByNameOrId(
+            configService,
+            logger,
+            [TargetType.Bzero, TargetType.DynamicAccessConfig, TargetType.Cluster, TargetType.Db, TargetType.Web],
+            argv.target,
+            CONFIGURE_MIN_AGENT_VERSION);
```
lipmas, line 69: We should only use `TargetType.Bzero` and `TargetType.Cluster`. target types Again this would be simplified if we just had `getTargetsByNameOrId` return a single already merged dictionary.

src/services/list-targets/list-targets.service.ts
lipmas: 3. Yeah my preference would be to pull it out into its own function. The only place using the filter right now is policy-utils `getTargetsByNameOrId` so would just need to add an extra filtering step there. Having it be its own function means we can reuse it though in other contexts which i think may be useful in the future too.

src/cli-driver.ts
```diff
@@ -266,23 +294,28 @@ export class CliDriver
...
+                if(!this.oauthCommands.has(baseCmd) || isServiceAccountLogin)
                     return;
                 await oAuthMiddleware(this.configService, this.logger);
             })
             .middleware(async (argv) => {
+                const isServiceAccountLogin = argv._[0] == 'service-account' && argv._[1] == 'login';
                 const isGenerateBash = argv._[0] == 'generate' && argv._[1] == 'bash';
-                if((this.adminOnlyCommands.has(baseCmd) || isGenerateBash) && !this.configService.me().isAdmin){
+                const isGenerateKubeYaml = argv._[0] == 'generate' && argv._[1] == 'kubeYaml';
+                if(!isServiceAccountLogin &&
```
lipmas, line 307: If service accounts can have admin role - shouldnt we just rely on  the `this.configService.me().isAdmin` check and not skip SAs in this middleware?

src/cli-driver.ts
```diff
@@ -254,7 +281,8 @@ export class CliDriver
                 this.logger.setGAService(this.GAService);
             })
             .middleware(async (argv) => {
-                if(!this.GACommands.has(baseCmd))
+                const isServiceAccountLogin = argv._[0] == 'service-account' && argv._[1] == 'login';
```
lipmas, line 284: I know this is a bit late to bring up...But i will point out if we had this as a login subcommand/option and not a `service-account` subcommand (i.e zli login service-account) then we dont have to add all these custom checks in the middleware.

src/system-tests/tests/suites/rest-api/service-accounts.ts
```diff
@@ -0,0 +1,159 @@
...
+        let serviceAccount: ServiceAccountSummary;
+        let expectedServiceAccount: ServiceAccountSummary;
+
+        beforeAll(async () => {
+            serviceAccountService = new ServiceAccountHttpService(configService, logger);
+            mfaService = new MfaHttpService(configService, logger);
+        });
+
+        test(`481487: Create service account as user`, async () => {
+            if (!RUN_AS_SERVICE_ACCOUNT) {
```
lipmas, line 20: you should use `testIf(!RUN_AS_SERVICE_ACCOUNT, ...)` to have jest actually skip this test (and report the test as skipped). https://github.com/bastionzero/zli/blob/496285c9c7398bb551d09ec613983a17765ffa84/src/system-tests/tests/utils/utils.ts#L19-L27

src/system-tests/tests/suites/rest-api/service-accounts.ts
```diff
@@ -0,0 +1,159 @@
...
+                expect(serviceAccount).toEqual(expectedServiceAccount);
+            } else {
+                logger.info(`Skipping service account creation as RUN_AS_SERVICE_ACCOUNT is set to true`);
+            }
+        }, 15 * 1000);
+
+        test(`481488: Get a service account's data by ID`, async () => {
+            const serviceAccountSummaryFromId = await serviceAccountService.GetServiceAccount(serviceAccount.id);
+            expect(serviceAccountSummaryFromId).toEqual(serviceAccount);
+        }, 15 * 1000);
```
lipmas, line 53: Wont this test and others in this suite fail if `RUN_AS_SERVICE_ACCOUNT` is set because `serviceAccount` here will not defined because we skipped the `Create service account as user` test. I think we might want to only run this suite if `if (!RUN_AS_SERVICE_ACCOUNT)`

src/system-tests/tests/suites/rest-api/service-accounts.ts
```diff
@@ -0,0 +1,159 @@
...
+
+        test(`481487: Create service account as user`, async () => {
+            if (!RUN_AS_SERVICE_ACCOUNT) {
+                const request: CreateServiceAccountRequest = {
+                    email: 'example-sa@sadomain.com',
+                    jwksURL: 'https://www.googleapis.com/service_accounts/v1/jwk/example-sa@sa-domain.iam.gserviceaccount.com',
+                    jwksURLPattern: 'https://www.googleapis.com/service_accounts/v1/jwk/*sa-domain.iam.gserviceaccount.com',
+                    externalId: '103439939706013570431'
+                };
+                const response = await serviceAccountService.CreateServiceAccount(request);
```
lipmas, line 27: This could just be calling `zli service-account create providersFile` like we do in system test setup

src/system-tests/tests/suites/rest-api/mfa.ts
```diff
@@ -1,31 +1,35 @@
...
             expect(mfaSummary).toEqual(mfaSummaryById);
         }, 15 * 1000);
 
-        test('5604: Enable MFA and verify it is enabled', async () => {
-            await mfaService.EnableMfa(userId);
-            const mfaSummary = await mfaService.GetCurrentUserMfaSummary();
-            expect(mfaSummary.enabled).toBe(true);
-            expect(mfaSummary.verified).toBe(false);
+        test('5604: Enable MFA and verify it is enabled as a user', async () => {
+            if (!RUN_AS_SERVICE_ACCOUNT) {
```
lipmas, line 25: you should use `testIf(!RUN_AS_SERVICE_ACCOUNT, ...)` to have jest actually skip this test (and report the test as skipped). https://github.com/bastionzero/zli/blob/496285c9c7398bb551d09ec613983a17765ffa84/src/system-tests/tests/utils/utils.ts#L19-L27

src/system-tests/tests/suites/rest-api/mfa.ts
```diff
@@ -35,28 +39,36 @@ export const mfaSuite = () => {
...
-            await mfaService.ResetSecret(false);
-            actualMfaSummary = await mfaService.GetCurrentUserMfaSummary();
-            expect(actualMfaSummary).toEqual(expectedMfaSummary);
-
-            await mfaService.DisableMfa(userId);
-            // forceSetup set to true should enable MFA
-            await mfaService.ResetSecret(true);
-            actualMfaSummary = await mfaService.GetCurrentUserMfaSummary();
-            expect(actualMfaSummary).toEqual(expectedMfaSummary);
+            if (!RUN_AS_SERVICE_ACCOUNT) {
```
lipmas, line 42: testIf

src/system-tests/tests/suites/rest-api/mfa.ts
```diff
@@ -35,28 +39,36 @@ export const mfaSuite = () => {
...
+            } else {
+                logger.info(`Skipping resetting mfa as RUN_AS_SERVICE_ACCOUNT is set to true`);
+            }
         }, 15 * 1000);
 
         test('5606: Disable MFA and verify it is disabled', async () => {
-            await mfaService.DisableMfa(userId);
-            const mfaSummary = await mfaService.GetCurrentUserMfaSummary();
-            expect(mfaSummary.enabled).toBe(false);
+            if (!RUN_AS_SERVICE_ACCOUNT) {
```
lipmas, line 65: testIf

src/system-tests/tests/suites/rest-api/users.ts
```diff
@@ -12,10 +14,18 @@ export const userRestApiSuite = () => {
             userService = new UserHttpService(configService, logger);
         });
 
-        test(`23988: Get current user's data`, async () => {
-            currentUser = await userService.Me();
-            expect(currentUser).toBeDefined();
-            expect(currentUser).toEqual(configService.me());
+        test(`23988: Get current user's data as a user`, async () => {
+            if (!RUN_AS_SERVICE_ACCOUNT) {
```
lipmas, line 18: testIf

src/utils/policy-utils.ts
tflpd: [661e1b2](https://github.com/bastionzero/zli/pull/551/commits/661e1b20df87ef63058c44505f88909c3de306d2)

src/utils/policy-utils.ts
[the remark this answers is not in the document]
tflpd: 24d89251

src/handlers/service-account/configure-service-account.handler.ts
tflpd: 3dcdbd37

src/handlers/service-account/configure-service-account.handler.ts
tflpd: 3dcdbd37

src/cli-driver.ts
tflpd: Yeah that is fair but I prefer to take the engineering overhead and provide a better UX to Product

src/system-tests/tests/suites/rest-api/service-accounts.ts
tflpd: But this would make this api test a system test instead.

src/utils/policy-utils.ts
tflpd: [1217721](https://github.com/bastionzero/zli/pull/551/commits/12177212e881b37009a983bd3fb45b38b1a14ed5)

lipmas (approved):
LGTM

PR #561, opened 2022-10-14
Fix 401 errors in http services after oauth refresh

lipmas:
## Description of the change

Fixes a bug where `HttpServices` constructed before an `oauth.service.refresh()` which updates the `sessionId`/`sessionToken` were not using the updated values for cookies sent with requests. This is a general fix for this type of error because in the `init` hook of the http request we always read any updated value for `sessionId` or `sessionToken` from the configService.

This caused a failure in our system tests after #546 introduced a new code path that would call `UserHttpService.me()` immediately after a refresh with a http service that was constructed before the refresh. This caused the me api call to return a 401 error and fail the test immediately. 

## Testing

**backend branch:** 
**bzero branch:** 

- [x] Yes

Release Notes: Fixed 401 unauthorized errors in some cases after an oauth refresh

## Related JIRA tickets

Relates to JIRA: CWC-2164

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain:

In some cases we were not using an updated sessionId/sessionToken cookie when sending requests from the zli after a refresh.


ymarcus93 (approved):

src/services/http/http.service.ts
```diff
@@ -22,26 +22,33 @@ export class HttpService {
...
             hooks: {
+                init: [
+                    (_) => {
+                        // Always update the sessionId/sessionToken cookies from
+                        // the config file in case they have changed from a
+                        // oauth.service.refresh()
+                        const sessionId = this.configService.getSessionId();
+                        if (sessionId) {
+                            const sessionIdCookie = new Cookie({key: 'sessionId', value: sessionId, path: '/', secure: true, sameSite: 'Strict'});
+                            this.cookieJar.setCookieSync(sessionIdCookie, this.baseUrl);
```
tflpd, line 38: Would it make sense to try and optimize this since it will be performed per every request? Like set the cookieJar only if the existing value is different than the one provided by conf?
Probably no, only reason I bring this up is because of the frequency this will be executed

tflpd (approved):

src/services/http/http.service.ts
lipmas: yeah i thought about that too. Problem is im not sure what we are optimizing for exactly. Like we could check if the cookies have changed and only set them if they have but that would require calling `cookieJar.GetCookies(...)` so is there some reason to expect the write operation is a lot more expensive? I believe its just updating an in-memory dictionary .

tflpd: Yeah I guess both are in-memory, on the one case it is a read, the other case a write, but either way minimal difference

PR #568, opened 2022-10-26
Update system tests for default MFA

SaiDadireddy:
## Description of the change

This pull request contains the corresponding system test changes for the default organization-wide MFA feature. More information on the feature can be found in the [backend pr](https://github.com/bastionzero/webshell-backend/pull/1413).

This pull request creates a handler for the `zli register` command that will add an mfa secret and use that for verification when a test runner is initialized. Also used in the system tests when a totp is required during re-registration. The infra changes can be found in the [infra pr](https://github.com/bastionzero/cwc-infra/pull/579).

## Testing

You can run `bzero-qa test sys-tests` command locally and expect to see the mfa secret printed after the test runner registers against the idp. Expect to see tests running. If something doesn't work you'll see a `401 Unauthorized` error.

**backend branch:**
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: 

## Related JIRA tickets

Relates to JIRA: CWC-2095

Relevant HotFix PR Number:

- [X] Yes
- [ ] No

If yes, please explain:
This will improve our security posture by requiring the `zli register` command to setup an mfa secret if one does not exist and verify mfa with totp if the session is not mfa verified.


src/system-tests/tests/system-test.ts
```diff
@@ -227,25 +226,22 @@ beforeAll(async () => {
...
 
     // Reset sessionId and sessionToken to get unique session for this test
     configService.setSessionId('');
     configService.setSessionToken('');
 
-    // Reset ssh config key paths because these are different for the IdPLogin
-    // tests (run as ec2-user) which uploads the config that system test uses
-    configService.clearSshConfigPaths();
+    // Call register to create new session token and id, and verify session with totp
+    const sysTestMfaSecret = process.env.MFA_SECRET;
```
lipmas, line 238: should check that this is not undefined/empty and throw an exception

src/utils/utils.ts
```diff
@@ -1184,4 +1184,14 @@ export function jsonDateReviver(key: any, value: any): any {
...
 export function customJsonParser(text: string) {
     return JSON.parse(text, jsonDateReviver);
-}
\ No newline at end of file
+}
+
+/**
+ * Custom function to extract the secret from the url returned by reset
+ */
+export function extractSecretFromUrl(mfaSecretUrl: string) {
```
lipmas, line 1192: `extractMfaSecretFromUrl`

src/system-tests/tests/suites/rest-api/policies/policies-organization-controls.ts
```diff
@@ -11,14 +13,20 @@ export const organizationControlsPolicySuite = () => {
...
         let policyService: PolicyHttpService;
 
+        // Adding an org policy should also add all subjects to the policy
         beforeAll(() => {
             policyService = new PolicyHttpService(configService, logger);
+            const policySubjects: Subject[] = [{id: systemTestUser.id, type: SubjectType.User}];
+            if(systemTestServiceAccount) {
+                // add this to the front of the array
+                policySubjects.unshift({id: systemTestServiceAccount.id, type: SubjectType.ServiceAccount});
+            }
```
lipmas, line 23: what are we testing here by adding the SA to the org control policy?

src/system-tests/tests/suites/rest-api/mfa.ts
```diff
@@ -1,68 +1,74 @@
...
         beforeAll(() => {
             mfaService = new MfaHttpService(configService, logger);
             subjectId = configService.me().id;
         });
 
-        testIf(!RUN_AS_SERVICE_ACCOUNT, '5603: Get MFA status and verify it is disabled as a user', async () => {
+        testIf(!RUN_AS_SERVICE_ACCOUNT, '5603: Get MFA status and expect it to be enabled as a user', async () => {
             const mfaSummary = await mfaService.GetCurrentUserMfaSummary();
-            expect(mfaSummary.enabled).toBe(false);
+            expect(mfaSummary.enabled).toBe(true);
```
lipmas, line 21: Hmm so this will be true for new organization/users now when running against a test runner. But in pipeline tests running against cloud-dev and cloud-staging the role account user wont have mfa enabled.

So i guess we either
1. condition this test based on `IN_PIPELINE`
2. or enable MFA for role account on in all system test orgs

src/system-tests/tests/suites/rest-api/mfa.ts
```diff
@@ -1,68 +1,74 @@
...
         }, 15 * 1000);
 
-        testIf(!RUN_AS_SERVICE_ACCOUNT, '5606: Disable MFA and verify it is disabled', async () => {
+        testIf(!RUN_AS_SERVICE_ACCOUNT, '544005: Verify MFA with TOTP after resetting secret', async () => {
+            expect(mfaSecretAfterReset).toBeDefined();
+            const totpPasscode = totp(mfaSecretAfterReset);
+            await expect(mfaService.VerifyMfaTotp(totpPasscode)).resolves.not.toThrow();
+        }, 15 * 1000);
+
+        testIf(!RUN_AS_SERVICE_ACCOUNT, '5606: Disable MFA and expect it to be disabled', async () => {
```
lipmas, line 62: And even if we did enable mfa for system test orgs this test is still going to leave it disabled...still thinking this through

src/system-tests/tests/suites/rest-api/policies/policies-organization-controls.ts
SaiDadireddy: When a new org policy is created, both all users and service accounts are added to the policy. So, checking that here.

src/system-tests/tests/suites/rest-api/mfa.ts
SaiDadireddy: I can make the change for 1 if we go that route. What changes would we make for 2?

src/system-tests/tests/suites/rest-api/policies/policies-organization-controls.ts
lipmas: oh i see this is the expected policy got it thanks

src/system-tests/tests/system-test.ts
SaiDadireddy: [33ea654](https://github.com/bastionzero/zli/pull/568/commits/33ea654777ef822b5005a5af2197ffc6032f030b)

src/utils/utils.ts
SaiDadireddy: [33ea654](https://github.com/bastionzero/zli/pull/568/commits/33ea654777ef822b5005a5af2197ffc6032f030b)

src/system-tests/tests/suites/rest-api/policies/policies-organization-controls.ts
```diff
@@ -11,14 +13,20 @@ export const organizationControlsPolicySuite = () => {
...
             expectedPolicySummary = {
                 id: expect.any(String),
                 type: PolicyType.OrganizationControls,
                 groups: [],
                 name: originalPolicyName,
-                subjects: [],
+                subjects: policySubjects,
```
lipmas, line 29: I think we should probably use  `expect.arrayContaining(policySubjects)` here because when running in pipeline the organizations are going to have more than just the role account/sa subjects.
 https://jestjs.io/docs/expect#expectarraycontainingarray

src/system-tests/tests/suites/rest-api/mfa.ts
lipmas: So how much value is this test actually providing? Im wondering if we could just remove this test and just keep the `Enable MFA and expect it to be enabled as a user` and `Disable MFA and expect it to be disabled` tests which explicitly enable/disable mfa within the test. That way we dont have to worry about these edge cases.

Another idea would be to move the test to `policies-organization-controls.ts` suite. Cause there if we created a new org control policy with mfa enabled then afterwards we could check if mfa is enabled for the system test subject.

SaiDadireddy: @lipmas This is already an existing test. However, previously it was testing to make sure mfa enabled was was false by default. So, we should now be testing that it's enabled by default when an org is created. If we add the test to the `policies-organization-controls.ts` suite, that would be testing a case where an existing org (without mfa) turns on global mfa and a new org policy is created. But I guess either way we're just checking that a new org policy is being created and we're testing if the user is added to the policy by retrieving their mfa summary. So I like this idea, we can remove this test from mfa suite and add it to the org policy suite.

src/system-tests/tests/suites/rest-api/mfa.ts
SaiDadireddy: @SaiDadireddy It's now expected that mfa is enabled by default right? So what if we move this disable test to occur before the enable test? That way we could test both, and leave mfa enabled after this test suite is finished.

src/system-tests/tests/suites/rest-api/policies/policies-organization-controls.ts
SaiDadireddy: Okay good idea, I'll check this out!

src/system-tests/tests/suites/rest-api/mfa.ts
tflpd: > So how much value is this test actually providing?

It depends on what is the goal of our tests right?
If it is that when something breaks we know that something broke then yeah, we can remove this.
If it is that when something breaks, we know what broke, then we should keep it.

I do not see any downsides in solution 1.

Codebuild/system-test-codebuild.yaml
```diff
@@ -36,7 +36,7 @@ phases:
       - git submodule init
       - git submodule update --recursive
       # Install Bzero-QA
-      - git clone git@github.com:bastionzero/cwc-infra.git --branch $CWC_INFRA_BRANCH /cwc-infra/
+      - git clone git@github.com:bastionzero/cwc-infra.git --branch default-MFA /cwc-infra/
```
tflpd, line 39: Leaving a note to revert this eventually, resolve last moment before merging

src/system-tests/tests/suites/rest-api/policies/policies-organization-controls.ts
SaiDadireddy: Not applicable anymore as subjects field of a policy is no longer used in place of `ExcludedSubjects` which is explained in the [backend pr](https://github.com/bastionzero/webshell-backend/pull/1413).

tflpd (approved):

PR #574, opened 2022-11-02
Multi-kube

ymarcus93:
Related PRs: https://github.com/bastionzero/webshell-common-ts/pull/102, https://github.com/bastionzero/webshell-backend/pull/1428

TODO:
- [x] Add case IDs to TestRails
- [x] Change format for BastionZero-managed entries

### Highlights
- Add support for starting multiple Kubernetes connections, without disconnecting any existing connections, on the same machine.
  - User can make connections to distinct targets, or connect to the same target as long as the supplied `targetUser` is distinct from other _running_ daemons on the same machine.
- Add support for listing Kube connections with `zli lc`.
  - By default, `shell`, `db`, and `kube` connections are returned.
- Add support for closing _all_ Kube connections with `zli close`.
- Convert `list-daemons kube` and `disconnect kube` to use shared `DaemonManagementService` class introduced in Multi-DB feature.
- Change how the `zli` updates the user's kube config. Instead of creating a single context+cluster+user entry, a context+cluster entry are now created _per_ connection. There is still only one user entry added.
  - Format:  `bzero-<targetName>_<targetUser>`. ~~**TODO**~~: Change this to `bzero-<targetUser>@<targetName>`.
  - Username for user entry also has prefix `bzero-` instead of just the user's email.
- **Major change**: `zli generate kubeConfig` is no longer required before connecting to a Kubernetes target. The user's kubeconfig is modified during `zli connect`.
  - The command still exists for backwards compatibility, and in case the user has accidentally corrupted their BastionZero-managed entries, they can run `zli generate kubeConfig --update` to restore it based on the zli config's list of kube daemons.
  - Note: We still respect the `KUBECONFIG` environment variable when running `zli generate kubeConfig` or `zli connect`. If the environment variable is not set, we default to `~/.kube/config` as before. It's important to preserve this behavior in case the user does not want BastionZero to touch their config at all.
- `kubectl` is no longer a dependency when updating the user's kube config. `kubectl config view --flatten` / merging logic has been written in typescript. We use the [`kuberenetes-client/javascript`](https://github.com/kubernetes-client/javascript) library to parse the user's kube config and to add type safety when constructing the generated config.
  - The only remaining dependency on `kubectl` in the `zli` is in the _catch-all_ kube handler (`zli kube`). 
- `zli kube` now returns an error if the user's current context is not a BastionZero one to prevent misuse of the command. We also check that the daemon is running before attempting to forward the command to `kubectl`
- A free port to use for the kube daemon is now decided at connect time instead of being a constant loaded from the config. `--customPort` is still allowed in case the user wants to connect with a specific port.
- If configured TLS cert for kube daemon file is missing, a new one is generated. If the user wants to force re-generate all kube daemon security settings, `zli generate kubeConfig --force` can be executed.
- Add support for default `namespace` when using `zli connect` with new `--namespace` flag. This namespace is used as a default when using `kubectl` and other clients that respect the `namespace` field in one's `kubeconfig`.

```      
--force       Force generates new kube daemon security settings. WARNING: Disconnects any running kube daemons  [boolean] [default: false]
```

#### Merging

Review aforementioned merge logic here:

https://github.com/bastionzero/zli/blob/fc755ffe640534d04e3e072170d3f80374a018bb/src/services/kube-management/kube-management.service.ts#L372-L385

#### Filtering

Since we now add multiple context and cluster entries per connection, there is now filtering logic that is executed whenever the user runs `zli ld` or `zli disconnect` to remove entries that are no longer usable (i.e. entries for daemons that are no longer running).

Review filtering logic:

https://github.com/bastionzero/zli/blob/fc755ffe640534d04e3e072170d3f80374a018bb/src/services/kube-management/kube-management.service.ts#L463-L488

The logic should also be immune to context name changes because we primarily determine if an entry is a BastionZero-managed entry by checking the context's referenced username.

#### `kube-ps1`

Try installing https://github.com/jonmosco/kube-ps1 to see your current kube context, including your `targetCluster`+`targetUser`, in your terminal prompt

## Testing

I've added unit tests for merge and filtering logic, and other aspects of this feature. I'm also using property-based testing for some aspects of the unit tests (using [`fast-check`](https://github.com/dubzzz/fast-check) library). System tests have been updated to make multiple Kubernetes connections to the same target (with different `targetUser` combination). `lc`, `ld`, `disconnect`, and `close` are now tested e2e with system tests, with respect to `kube` connections.

Please help run a _manual_ test by starting a kube connection before upgrading your `zli` to this branch. There is migration logic that should maintain your existing config in the new schema. Try running `zli lc`, `zli ld`, and `zli kube -- get pods` after upgrading. Also, try setting your default target group (`zli default-targetGroup --set`) before upgrading and see that it is maintained post-upgrade.

**backend branch:** feat/multi-kube
**bzero branch:** 
**charts branch:** 

- [X] Yes

Release Notes: Add support for starting multiple Kubernetes connections, without disconnecting any existing connections, on the same machine.

## Related JIRA tickets

Relates to JIRA: CWC-2156

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


src/system-tests/tests/utils/zli-utils.ts
```diff
@@ -1,34 +1,11 @@
...
-                }
-            });
-        } catch (e) {
-            rej(e);
-        }
-    });
-
-    await callbackComplete;
+    const { baseCmd, parsedArgv } = makeCaseInsensitive(cliDriver.availableCommands, zliArgs);
+    await cliDriver.getCliDriver(true, baseCmd).parseAsync(parsedArgv, {}, callback);
```
lipmas, line 8: you love to see it

src/handlers/list-daemons/list-daemons.handler.ts
```diff
@@ -141,45 +141,34 @@ async function kubeStatusHandler(
...
-        } catch (err) {
-            logger.error(`Error contacting Kube Daemon. Please try logging in again and restarting the daemon. Error: ${err}`);
-        }
-    }
-}
+    await handleStatus(
+        kubeDaemonManagementService,
+        tableHeader,
+        async (connectionId, result) => {
+            const connDetails = connectionId ? `${result.config.targetCluster} (connId: ${connectionId})` : result.config.targetCluster;
```
lipmas, line 155: `${result.config.targetUser}@${result.config.targetCluster}`

src/services/config/config.service.ts
```diff
@@ -94,6 +96,70 @@ export class ConfigService implements ConfigInterface {
                     if (appName)
                         config.set('serviceUrl', this.getServiceUrl(appName));
                 },
+
+                '>=6.13.0': (config: Conf<BastionZeroConfigSchema>) => {
```
lipmas, line 100: move this to the bottom of migrations so they are in chron order

src/handlers/connect/kube-connect.handler.ts
```diff
@@ -25,22 +29,51 @@ export async function startKubeDaemonHandler(
...
-        logger.error('Please make sure you have created your kubeconfig before running connect. You can do this via "zli generate kubeConfig"');
-        return 1;
+        let warnMsg = 'There is already a kube daemon running for this target and user';
+        if (matchingContextName) {
+            warnMsg += ` with context: ${matchingContextName}`;
+        }
+        logger.warn(warnMsg);
+
+        await cleanExit(1, logger);
     }
```
lipmas, line 64: So i think this check needs to be moved to before we call the universal connection endpoint that creates the connection https://github.com/bastionzero/zli/blob/a7ef169918b92eddc8c4ceddd1253f83f40e52f7/src/handlers/connect/connect.handler.ts#L40. Im noticing even when i hit this error because i already have a connection to the same target/with the same target user the connection is still opened in the backend and on the agent (and it eventually times out).

Alternatively maybe we could move this check to the backend and just thow an exception there?

ymarcus93: Yes. I noticed that as well. In fact, I think this is a general issue with all types of errors that can be thrown in the respective connection type connect handlers after the `CreateUniversalConnection` API has been called. For example, I believe the connection remains open, until the timeout happens and daemon never connects, when using a port that is already in use ("It looks like an application is already using port: xyz").

Could we fix this problem generally for all possible errors in the try-catch block for all the sub-handlers by calling `CloseConnection` in the `catch` block here? https://github.com/bastionzero/zli/blob/a7ef169918b92eddc8c4ceddd1253f83f40e52f7/src/handlers/connect/connect.handler.ts#L70-L74

src/services/close-connections/close-connections.service.ts
```diff
@@ -39,4 +40,21 @@ export async function closeDbConnections(
...
+export async function closeKubeConnections(
+    configService: ConfigService,
+    logger: Logger
+): Promise<void> {
+    const connectionHttpService = new ConnectionHttpService(configService, logger);
+
+    // Send close requests concurrently and wait for all promises to
+    // resolve/reject
+    const openKubeConnections = await connectionHttpService.ListKubeConnections(ConnectionState.Open);
+    await Promise.all(openKubeConnections.map(conn => connectionHttpService.CloseConnection(conn.id)));
```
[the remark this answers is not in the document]
ymarcus93: > Maybe we should try..catch the actual connectionHttpService.CloseConnection call here so that we dont reject immediately if one of the connections fails to close and we will actually try to close all connections.

I think I can use `Promise.allSettled` and then log as warnings the rejected promises: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

> Also since this is using the generic close endpoint this could probably be a shared function for both kube/db since same comment applies there.

Agreed. I'll pull this into a function and pass the list of open connections

src/handlers/connect/kube-connect.handler.ts
lipmas: > Could we fix this problem generally for all possible errors in the try-catch block for all the sub-handlers by calling CloseConnection in the catch block here?

Yeah we should definitely do that as well. 

I do think it still might be worth trying to catch this case explicitly before creating the connection in the backend though. Creating the connection and then closing it immediately just adds unnecessary overhead and potential for bugs. I guess same applies for port already in use (which i think can only happen now [if using a custom port](https://github.com/bastionzero/zli/blob/c6a09cdd5eed2167e68bd24ca8a5035ecb8ae217/src/handlers/connect/kube-connect.handler.ts#L69-L77)?). We could just move that code into the connect-handler before the `CreateUniversalConnection` call.

ymarcus93: It also happens with configured local port on backend for db and web targets and then it not being available on user's machine

ymarcus93: Ok, so I modified the try-catch here e9873ea. However, looking at the code again, I'm not sure how to move both the custom port check and kube daemon already exists check _before_ calling `CreateUniversalConnection`. We don't know what the type of connection it is before this API is called, so it doesn't seem possible to bring these specific checks up a level without changing how the connect API works. What do you think? Do you have any suggestions?

lipmas: Ok yeah i see the predicament. I guess it doesnt make sense to do any kube specific sanity checks before we know this is a kube target.  The only other way i could see would be to move the checks to the backend but that is not possible for the custom port and for the kube daemon already existing we would prevent a user from creating a connection to the same cluster on 2 different machines which also doesnt seem ideal. Ok i think the best we can do is just rely on the try...catch you added to handle these cases and close the connection immediately.

src/services/kube-management/kube-management.service.ts
```diff
@@ -0,0 +1,607 @@
...
+                // destEntriesMap takes precedence if entry.name matches
+                // something srcEntries has
+                mergedEntries.push(foundTheirEntry);
+                // Remove from map, so by end of iteration destEntriesMap
+                // includes new entries only
+                destEntriesMap.delete(entry.name);
+            } else {
+                // Keep src's entry if not found in dest
+                mergedEntries.push(entry);
+            }
```
lipmas, line 407: Couldnt you just simplify this to only add src entries if not found in destination entries and not mutate destEntriesMap?

```
if (! foundTheirEntry) {
    mergedEntries.push(entry);
}
```

src/handlers/generate/generate-kube-config.handler.ts
```diff
@@ -1,178 +1,105 @@
...
+            kubeDaemonConfig.targetCluster,
+            kubeDaemonConfig.targetUser,
+            kubeDaemonConfig.localPort,
+            kubeSecurityConfig.token,
+            kubeDaemonConfig.defaultNamespace
+        );
+
+        // Merge generated config with rolling config
+        rollingKubeConfig = mergeKubeConfig(rollingKubeConfig, generatedKubeConfig);
     }
```
lipmas, line 81: so this will set the current context to whatever the last kubeDaemonConfig is in the for loop right?

src/services/kube-management/kube-management.service.ts
ymarcus93: Oh yeah that's simpler. 6ce3842062adf2e41b8b7a175a87c143369974f9

src/handlers/generate/generate-kube-config.handler.ts
ymarcus93: Yes, and it will be the last `connect` they did against a kube target since map order is preserved

lipmas: oh neat ok yeah i was wondering about that.

lipmas (approved):

PR #584, opened 2022-11-15
Send-logs should still work when daemon log file doesn't exist

SaiDadireddy:
## Description of the change

Currently, if the user has not worked with anything daemon related, they won't have any daemon logs in their local. So, when they run the `zli send-logs` command - the command fails, existing zli logs are not sent and an error is returned. 

The fix: Check if file exists. If file doesn't exist, then we will simply continue, which will create an empty zip entry with no logs for daemon logs. Reading file will no longer hit `ENOENT`, so errors from reading will still result in the command failing and an error being logged. 

## Testing

Check and make sure you don't have a local zli daemon dev file (the prod one is fine since this is testing in dev). Run `zli send-logs` command and expect to see the zip uploaded to the S3 bucket. When you download the zip and open it, it should still have 2 files, one for zli logs (not empty) and one for daemon logs (empty).

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: 

## Related JIRA tickets

Relates to JIRA: CWC-2254

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:


tflpd (approved):

src/handlers/send-logs/send-logs.handler.ts
```diff
@@ -49,8 +51,10 @@ async function getFilteredLogContents(logger: Logger, zliLogFilePath: string, da
     try {
         daemonLogContents = fs.readFileSync(daemonLogFilePath, 'utf-8').split('\n');
     } catch (err) {
-        logger.error('Error reading local daemon log files');
-        await cleanExit(1, logger);
+        if(err.code !== 'ENOENT') {
+            logger.error('Error reading local daemon log files');
+            await cleanExit(1, logger);
+        }
```
lipmas, line 57: I think using `fs.existsSync` might be a little cleaner way to handle this instead of catching the error.

SaiDadireddy: Agreed [4710747](https://github.com/bastionzero/zli/pull/584/commits/4710747dd4f13ffbdfdf9980b7b0d80d9c1fa9b3)

lipmas (approved):

PR #585, opened 2022-11-16
Fix: Yet better daemon shutdown logic

johncmerfeld:
## Description of the change

Basically there were two problems with how we were shutting down daemons:

  1. The timeout was too short; a daemon that fails to close its connections might wait as long as 10 seconds to gracefully exit. Obviously the ZLI needs to wait longer than 10 seconds for that to happen. I increased this timeout slightly
  2. if the daemon times out, the ZLI shouldn't just throw up its hands and fail. It should attempt to SIGKILL the daemon, which is uncatchable. So I added that

I also switched some log messages *back* to info level from debug level because otherwise the user is waiting a long time with no clue why

### Update

This was made somewhat redundant after #574 was merged. However, it still provides an important utility for safely shutting down daemons

## Testing

Easiest way to test this is to simulate a stuck daemon. I suggest adding a line [here](https://github.com/bastionzero/bzero/blob/develop/bctl/daemon/daemon.go#L67) before rebuilding the ZLI: 

  1. To test the first case, add this line: `time.Sleep(10 * time.Second)` -- the ZLI will wait the 10 seconds and still successfully allow the daemon to shut down
  2. To test the second case, add this line: `time.Sleep(10 * time.Minute)` -- the daemon will not exit on its own and the ZLI will have to kill it

For both cases, it's helpful to be logged into prod and connect to two separate cluster targets, but it's sufficient to just `zli connect` and `zli disconnect` to see the effects.

**backend branch:** develop
**bzero branch:** develop
**charts branch:** 

- [x] Yes

Release Notes: Make daemon shutdown resilient to hung connections

## Related JIRA tickets

Relates to JIRA: CWC-2261

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


src/utils/daemon-utils.ts
```diff
@@ -298,22 +298,36 @@ async function deleteIfExists(pathToFile: string) {
...
+                    if (!killed) {
+                        logger.warn(warningMsg('process may no longer exist'));
+                    }
+                    logger.info(successMsg);
+                } catch (err: any) {
+                    logger.warn(warningMsg(err));
+                }
             } else {
-                logger.warn(`Attempt to kill the daemon running on pid ${localPid} failed: ${err}\nConsider running \`kill -9 ${localPid}\` to force kill it`);
+                logger.warn(warningMsg(err));
```
ymarcus93, line 330: Can we move this `try catch` logic where we first try to `SIGINT` then `SIGKILL` if we've timed out into a new function in `processManager`, and call it from the `DaemonManagementService` here?: https://github.com/bastionzero/zli/blob/ba48e81673b167b0c73155c8570cc1713a2a1129/src/services/daemon-management/daemon-management.service.ts#L185-L200

Alternatively, move this logic into the dameon manaagement service directly. Otherwise, this logic will not be preserved for multi-db, and soon to be released, multi-kube connections.

Also, keep in mind the issue being fixed probably slightly goes away (still exists but at a lesser rate) since we don't disconnect existing daemon anymore when they try to connect again. We find a random port, or respect `--customPort` or what's stored on backend (for db connections).

After multi-kube is released, the only entity calling this `killDaemon` util function is web connections as they've not been migrated yet to multi-web and `DaemonManagementService`

johncmerfeld: Sure, happy to move it there, although this also affects `zli disconnects` more broadly (less critical but still present). I figured this would be nullified slightly by multi-kube but still, it's an easily-fixable bug that exists today

johncmerfeld: [5bb6ef0](https://github.com/bastionzero/zli/pull/585/commits/5bb6ef01502f1a764130097364c2e0b83c7be2c6)

ymarcus93 (approved):

PR #598, opened 2022-12-07
Feat/onelogin integration

SaiDadireddy:
## Description of the change

This pull request adds changes to allow a user to be able to sign in via OneLogin. It also covers the corresponding changes need to for running system tests with OneLogin as an idP.

## Testing

Login normally to my dev stack `saic.bastionzero.com` or your own via the zli (`zli login`), using the OneLogin sign in option. You will need to be added as a user to our `bastionzero` org first in onelogin so reach out to Sai or Thanos to be added.

**backend branch:** feat/onelogin-integration
**bzero branch:** develop
**charts branch:** 

- [x] Yes

Release Notes: OneLogin support for ZLI

## Related JIRA tickets

Relates to JIRA: CWC-2199

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain:
Adds a new way to auth against BastionZero via OneLogin. Backend pr covers adding an authentication layer to validate the tokens we receive from OneLogin.


src/services/config/config.service.ts
```diff
@@ -509,6 +520,8 @@ export class ConfigService implements ConfigInterface {
             return 'offline_access openid email profile User.Read';
         case IdentityProvider.Okta:
             return 'offline_access openid email profile';
+        case IdentityProvider.OneLogin:
+            return 'openid profile';
```
lipmas, line 524: so `offline_access` isnt needed here? For other IdPs this scope is usually needed in order to get a refresh token

src/services/oauth/oauth.service.ts
```diff
@@ -221,6 +220,7 @@ export class OAuthService implements IDisposable {
             prompt = 'consent';
             break;
         case IdentityProvider.Microsoft:
+        case IdentityProvider.OneLogin:
             prompt = 'login';
```
lipmas, line 224: whats the difference between `login` and `prompt` for onelogin ?

src/services/config/config.service.ts
SaiDadireddy: We can't use it because it causes an error when using it with the Auth code flow which is what we're using to get the id_token https://developers.onelogin.com/openid-connect/scopes

lipmas: Ok so when using Authorization Code flow for one login refresh token is included in the `TokenSet` then right?

SaiDadireddy: From the link above:
```
refresh_token | Only returned if a Refresh Token Timeout period has specified in your OpenId Connect app settings via the OneLogin portal.
```

src/services/oauth/oauth.service.ts
SaiDadireddy: OneLogin doesn't have a `prompt`. Only `login` and `none`. `login` means the user will be prompted with login and none means they won't be prompted. If they don't have a current session, then a `login_required` error is returned. See more here https://developers.onelogin.com/openid-connect/api/authorization-code

Codebuild/system-test-codebuild.yaml
```diff
@@ -36,7 +36,7 @@ phases:
       - git submodule init
       - git submodule update --recursive
       # Install Bzero-QA
-      - git clone git@github.com:bastionzero/cwc-infra.git --branch $CWC_INFRA_BRANCH /cwc-infra/
+      - git clone git@github.com:bastionzero/cwc-infra.git --branch feat/onelogin-integration /cwc-infra/
```
lipmas, line 39: reminder to switch this before merging

lipmas (approved):

src/services/config/config.service.ts
SaiDadireddy: https://developers.onelogin.com/openid-connect/api/authorization-code-grant this is the right link actually

tflpd (approved):

src/services/config/config.service.ts
SaiDadireddy: Update: checked the app and the default timeout period for refresh tokens (when one isn't specified) is 30 days

PR #603, opened 2022-12-13
[SystemTests] Dont yum update when installing bzero

lipmas:
## Description of the change

yum update will upgrade all system packages which may have side-effects. For example upgrading postgres 12.12 -> 12.13 caused postgres to quit unexpectedly and fail system tests in the pipeline.

## Testing

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Fix system test issue for al2 where upgrading postgres was causing db suite to fail

## Related JIRA tickets

Relates to JIRA: CWC-2308

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:


lgmugnier (approved):
do we want to to remove the apt updates as well?

lipmas:
> do we want to to remove the apt updates as well?

Nope i dont think so because`apt update` is actually required after adding the apt-repository before we can install bzero. In general though `apt update` has different semantics than `yum update` only yum actually upgrades packages whereas apt will just fetch the latest package version metadata and `apt upgrade` must be run separately.

PR #632, opened 2023-02-21
Add authorized Github Actions JIT [CWC-2400]

tflpd:
## Description of the change

This PR enables customers to create a list of authorized Github Actions which can create expiring policies for users and service accounts that perform routine tasks during a deployment/release. Essentially when a commit is pushed on a customer's Github `master` branch, a custom action which we provide to the customer gets triggered. This action, if added on the list of authorized actions for this customer, will be allowed to create some expiring policies which then a service account in the customer's pipeline can utilize to temporarily gain access to some critical resources.

This PR adds support for the management of these actions in BastionZero.

## Testing

**backend branch:** feat/github-jit
**bzero branch:** develop
**charts branch:** 

- [x] Yes

Release Notes: Add authorized Github Actions JIT

## Related JIRA tickets

Relates to JIRA: CWC-2400

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain:
This PR introduces a new entity that can create expiring policies, Github Actions. Only authorized Github Actions are allowed to do so and they are also only allowed to access their own policy endpoint and nothing else.


ymarcus93 (approved):

src/handlers/authorized-github-action/delete-authorized-github-action.handler.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ConfigService } from '../../services/config/config.service';
+import { cleanExit } from '../clean-exit.handler';
+import yargs from 'yargs';
+import { AuthorizedGithubActionHttpService } from '../../http-services/authorized-github-action/authorized-github-action.http-services';
+import { deleteAuthorizedGithubActionArgs } from './delete-authorized-github-action.command-builder';
+
+export async function deleteAuthorizedGithubActionHandler(configService: ConfigService, logger: Logger, argv : yargs.Arguments<deleteAuthorizedGithubActionArgs>) {
+    const authorizedGithubActionHttpService = new AuthorizedGithubActionHttpService(configService, logger);
+
+    const authorizedGithubActions = await authorizedGithubActionHttpService.ListAuthorizedGithubActions();
```
SaiDadireddy, line 11: This might be intentional so feel free to ignore if that's the case, but maybe it would be better to use `GetAuthorizedGithubAction()` here since you already have the specific `githubActionId` so that you don't have to filter after? If you choose to go down this route, the way `GetAuthorizedGithubAction` is set up right now it seems to throw a `BadHttpRequestException` so would either need a try/catch here or return null from the backend if not found.

src/handlers/authorized-github-action/list-authorized-github-action.handler.ts
```diff
@@ -0,0 +1,39 @@
...
+    configService: ConfigService,
+    logger: Logger,
+){
+    const authorizedGithubActionHttpService = new AuthorizedGithubActionHttpService(configService, logger);
+    const authorizedGithubActions = await authorizedGithubActionHttpService.ListAuthorizedGithubActions();
+    if(!! argv.json) {
+        // json output
+        console.log(JSON.stringify(authorizedGithubActions));
+    } else {
+        if (authorizedGithubActions.length === 0){
```
SaiDadireddy, line 22: I think this check should be done before the main if statement so that if user chooses to print json, it'll tell them that there are no authorized Github Actions instead of printing nothing.

src/handlers/authorized-github-action/list-authorized-github-actions.command-builder.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function listAuthorizedGithubActionsCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<listAuthorizedGithubActionsArgs> {
+    return yargs
+        .option(
+            'json',
+            {
+                type: 'boolean',
+                default: false,
+                demandOption: false,
+                alias: 'j',
```
SaiDadireddy, line 15: NIT: Can we add a description here as well to keep it consistent with other list commands we have in zli?`Formats the output in JSON format`

src/handlers/authorized-github-action/list-authorized-github-action.handler.ts
tflpd: When a user expects `json` we only provide `json` responses bc this is probably used in a script and a custom message would break that script, see other references where we do the same. Is there somewhere we do not apply this logic?

SaiDadireddy: Yep you're right. Feel free to resolve!

SaiDadireddy (approved):

src/handlers/authorized-github-action/delete-authorized-github-action.handler.ts
tflpd: So, couple of things
- To use `GetAuthorizedGithubAction` we need to have the BZ `Guid` of this action which we do not, we only have the Github ID
- Throwing a `BadHttpRequestException` and returning a `404` is perfectly valid, it will be taken care [here](https://github.com/bastionzero/zli/blob/f70a8b894bec530b622ca4e32f460383eff4b203/src/services/http/http.service.ts#L120-L121)

So this way both places, if required individually, are ready to handle gracefully an invalid Github ID

src/handlers/authorized-github-action/list-authorized-github-actions.command-builder.ts
tflpd: 0ab53775

lipmas (approved):

PR #643, opened 2023-03-10
Fix race condition when provisioning multiple daemons

johncmerfeld:
## Description of the change

As noted in the ticket, this is a fix of the underlying race condition and a fix to how we inform the user that the daemon failed to start.

Quibbles:
  - I don't like adding new packages to our dependencies, but [async-lock](https://www.npmjs.com/package/async-lock) seems solid enough and has a small footprint.
  - Maybe you don't like my use of `fs.existsSync`. Feel free to change
  - See the `TODO` I added. We should get product's input on this language

Please note that I am about to leave town and this should probably be merged before I get back. Feel free to get on in there and make changes!

### Additional comments after change of ownership (@johncmerfeld  -> @lgmugnier )
Before this PR, the pattern was to overwrite the existing daemon executable whenever we were about to spawn a new daemon child process. John made it so that we were only ever creating a new daemon executable if there wasn't any and my (@lgmugnier) changes made it so that we are creating a versioned daemon (aka daemon executable includes zli version) so that we are only ever creating new daemons if there are none of the version we are on and then try to delete any old ones.

## Testing

You can spin up multiple agents, but the easiest way is to just pound one agent with a bunch of concurrent requests.

Run `zli generate sshConfig` and then execute this script

```python
#!/usr/bin/env python3
from multiprocessing import Process
from subprocess import call 

if __name__ == '__main__':
    for _ in range(10):
        command = "echo hello"
        target_name = f'root@john-bzero-agent' # <-- your agent
        p = Process(target=call, args=(['ssh', target_name, command],))
        p.start()
```

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Fix race condition when provisioning multiple daemons

## Related JIRA tickets

Relates to JIRA: CWC-2489, CWC-1937

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


ymarcus93:
Version improvement looks good to me so that we don't break upgrade process. However, I think we can remove the usage of the `async-lock` dependency. My testing from yesterday shows it's not working across parallel executions of `zli` processes (remove the `daemon` file in `~/Library` before running the python script), so I'm not sure what benefit we're getting from it.

This PR definitely helps alleviate the race condition from multiple `zli connect` calls once the daemon file exists because all future calls will just see the file exists and return. But I don't think `async-lock` is what is helping that.

src/utils/daemon-utils.ts
```diff
@@ -527,15 +533,17 @@ export function waitForDaemonProcessExit(logger: Logger, loggerConfigService: Lo
...
                 case DAEMON_EXIT_CODES.DAEMON_PANIC: {
-                    logger.error(`daemon process terminated unexpectedly -- for more details, try running the zli with the --debug flag set`);
+                    logger.error(`Daemon process terminated unexpectedly -- for more details, try running the zli with the --debug flag set`);
                     break;
                 }
+                case DAEMON_EXIT_CODES.FAILED_TO_START_126:
+                case DAEMON_EXIT_CODES.FAILED_TO_START_127:
                 case null: {
-                    logger.error(`daemon process terminated while starting up`);
+                    logger.error(`Failed to establish connection to target, Please try again or contract your administrator`);
```
ymarcus93, line 542: Did we mean contact?

src/utils/daemon-utils.ts
```diff
@@ -238,70 +240,90 @@ export async function startDaemonInDebugMode(finalDaemonPath: string, cwd: strin
...
-    // Check if the file exists, delete if so
-    if (fs.existsSync(pathToFile)) {
-        // Delete the file
-        fs.unlinkSync(pathToFile);
-    }
+            // Grant execute permission
+            const chmod = utils.promisify(fs.chmod);
+            await chmod(finalDaemonPath, 0o755);
+
+            return lockfile.unlockSync('copyExecutableToLocalDir', {
```
ymarcus93, line 316: Is there a reason the `unlock` has `sync` but `lock` does not?

src/utils/daemon-utils.ts
ymarcus93: Typo is still here: `contract`

ymarcus93:
Tested again successfully. With `daemon` existing and without.

src/utils/daemon-utils.ts
lgmugnier: I wanted to use the async version because it gives a clear dilineation on what we care about locking vs not. We are awaiting it regardless, so it happens syncronously

ymarcus93 (approved):

PR #649, opened 2023-03-20
Support 1:1 IdP usernames as target user in system tests

SaiDadireddy:
## Description of the change

This PR makes necessary system test changes to allow test runners to connect to linux targets via shell and ssh using the `roleaccount` idp username as the target user. There are no explicit zli changes for this feature but the command mentioned in the testing section below should work as described.

## Testing
- `zli lc` - should list your open shell/ssh connection and show your idp username as the target user 
- `zli lt -d` - should list the targets available to you and should show your idp username as the available target user
- `zli policy list` - should list all available policies and should show `{username}` as the target user since this is a general list view of the policies and not specific to one user.
- `zli policy add-targetuser` - should work as excepted, should be able to add `{username}` as a target user 
- `zli policy delete-targetuser` - should work as excepted, should be able to delete `{username}` as a target user 
- `zli generate sshConfig` - should add your idp username as an available target user to the bzero target entry in the bz config file, whether `{username}` is one of many target users on that policy or the only one

**backend branch:** idp-usernames
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Support 1:1 IdP usernames as target user

## Related JIRA tickets

Relates to JIRA: CWC-117

Relevant HotFix PR Number:

- [ ] Yes
- [X] No

If yes, please explain:

- [X] Yes
- [ ] No

Docs branch name is idp-usernames


src/system-tests/tests/system-test-setup.ts
```diff
@@ -24,6 +24,7 @@ import { MfaHttpService } from '../../http-services/mfa/mfa.http-services';
 
 // User to create for bzero targets to use for connect/ssh tests
 export const bzeroTargetCustomUser = 'bzuser';
+export const idpUsernameTargetCustomUser = 'roleaccount';
```
lipmas, line 27: Just making a note that this only works because all of our IdPs have a user with email in the form `roleaccount@...`. I double checked this is the case for all of our IdPs by checking whoami in the config files of each. Im curious if this works for service accounts though- so i added the `ci-run-as-service-account` label to the PR

src/system-tests/tests/utils/connect-utils.ts
```diff
@@ -84,6 +84,40 @@ export class ConnectTestUtils {
         return await this.runShellConnectTestHelper(connectTarget, stringToEcho, exit);
     }
 
+    public async runShellConnectTestIdpUsername(connectTarget: ConnectTarget, stringToEcho: string, exit: boolean) {
```
lipmas, line 87: I must be missing something- why did we have to create a new helper method instead of just reusing `runShellConnectTestHelper`. What is different in this case?

SaiDadireddy: Not much actually, I was trying to avoid adding another argument to `runShellConnectTest` but this is better [c64f50a](https://github.com/bastionzero/zli/pull/649/commits/c64f50a6449841a41a354f0464ec2a7d80cdfa9f). Imports cleanup -> [1262bc7](https://github.com/bastionzero/zli/pull/649/commits/1262bc721fec08219b1a25fce66f0795d7c04409)

lipmas (approved):

PR #656, opened 2023-04-05
Updates necessitated by MFA grace period feature

chad-cwc:
## Description of the change

The goal here is to do the minimum set of changes required for the ZLI to work with the back end changes for the MFA grace period. Another ZLI PR will be done soon that does a better job of integrating the MFA grace period feature.

The login handler changes will allow users to experience the new grace period, but they won't necessarily be aware of it. After the grace period expires, they will be forced to set up MFA as usual. During the grace period, logging in will function as if MFA is not enabled.

## Testing

**backend branch:** feat/mfa-grace-period
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Update for MFA grace period

## Related JIRA tickets

Relates to JIRA: CWC-2520

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


src/handlers/login/login.handler.ts
```diff
@@ -104,6 +104,7 @@ export async function loginUserHandler(configService: ConfigService, logger: Log
         }
         break;
     case MfaActionRequired.RESET:
+    case MfaActionRequired.RESET_DEFER_ALLOWED:
```
chad-cwc, line 107: I believe this is actually optional, and everything will work fine without it (at least based on the current back end implementation). So, I'm asking Product if we want this behavior or the behavior that will result from removing this.

chad-cwc: Changed it based on conversation with Product.

https://github.com/bastionzero/zli/pull/656/commits/e0a14a736cecd7dab9ccb87d69e38dcc672c0af2

src/handlers/register/register.handler.ts
```diff
@@ -29,7 +29,7 @@ export async function  registerHandler(mfaSecret: string, configService: ConfigS
         }
         const totpPasscode = totp(mfaSecret);
         await mfaService.VerifyMfaTotp(totpPasscode);
-    } else if (resp.mfaActionRequired == MfaActionRequired.RESET) {
+    } else if (resp.mfaActionRequired == MfaActionRequired.RESET || resp.mfaActionRequired == MfaActionRequired.RESET_DEFER_ALLOWED) {
```
SaiDadireddy, line 32: System tests are not going to test the grace period right? It should still take the mfa token created here and inject it when the [test runner is being initialized](https://github.com/bastionzero/cwc-infra/blob/919ab0f6b7169795496677a7072fb36a3cf03bbf/Bzero-QA/bzero_qa/scripts/test/systemTests/runnerInit.py#L262-L269)?

lipmas (approved):

src/handlers/register/register.handler.ts
chad-cwc: Cool. I will look into whether we should add some system tests in my next ZLI PR, which is intended to fully integrate the grace period into the ZLI UX.

SaiDadireddy (approved):

tflpd (approved):

PR #657, opened 2023-04-06
Windows support for ssh, connect, and web

lgmugnier:
## Description of the change

Add windows support for the zli!

Uses changes from 
https://github.com/bastionzero/bzero/pull/234

Currently, the PR only covers the first wave of changes to the zli to get it to work on windows. This includes a lot of small, somewhat independent modifications to mostly get connect, ssh, and web running although there are some changes that have ramifications for other plugins. Neither kube nor db will work on this branch because they require config changes which will come in the next wave.

Causes for the changes include:
- Windows allows spaces in its file paths, to counteract this for things like generating the ssh config, we must add quotes around everything
- unlink does not do exactly the same thing as it does in unix and rm is a much safer and more cross-platform command
- Windows environment variables are often names differently (HOMEPATH instead of HOME etc.)
- we now spawn using the file path of the daemon executable instead of that of the process. This is because providing the full file path in windows (because of spaces and other stuff) didn't work and this is the correct use of spawn anyways.
- we cannot rely on the `openssl` existing as a command so we use a library to generate certificates now

## Testing

I recommend building the binary for windows from your mac with `npm run release:windows` and then moving that over to your windows machine to run. With that binary you can run most zli commands. The PR DOES NOT include support db or kube.

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Adds windows support for our zli!

## Related JIRA tickets

Relates to JIRA: CWC-2444

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


src/handlers/generate/ssh/generate-ssh-config.command-builder.ts
```diff
@@ -16,5 +17,5 @@ export function generateSshConfigCmdBuilder(yargs: yargs.Argv<{}>) : yargs.Argv<
             description: 'Specifies an alternate location for the BastionZero config file'
         })
         .example('$0 generate sshConfig', 'Create and link an ssh config file based on your organization\'s policies')
-        .example('$0 generate sshConfig --mySshPath path/to/config --bzSshPath path/to/bz-config', `Optionally specify filepaths (defaults to ${process.env.HOME}/.ssh/config and ${process.env.HOME}/.ssh/bz-config respectively)`);
+        .example(`$0 generate sshConfig --mySshPath ${path.normalize('path/to/config')} --bzSshPath ${path.normalize('path/to/bz-config')}`, `Optionally specify filepaths`);
```
lgmugnier, line 20: I removed the defaults from this line because I though it was misleading, we should have the defaults set in the objects above instead of keeping them as nil but that breaks system tests for some reason. I will figure out how to fix that in a different PR.

package.json
```diff
@@ -150,7 +156,8 @@
     "scripts": [],
     "assets": [
       "node_modules/figlet/fonts/Standard.flf",
-      "bzero/bctl/daemon/daemon"
+      "bzero/bctl/daemon/daemon",
+      "bzero/bctl/daemon/daemon.exe"
```
johncmerfeld, line 160: Woop woop!

src/utils/utils.ts
```diff
@@ -1133,11 +1133,9 @@ export function getZliRunCommand(): string {
 
 export function removeIfExists(file: string): void {
     try {
-        fs.unlinkSync(file);
+        fs.rmSync(file, {force:true});
     } catch (err) {
-        if (err.code !== 'ENOENT') {
-            throw err;
-        }
+        throw (`Error deleting file: ${file}. Error: ${err}`);
```
johncmerfeld, line 1138: I disagree with this logic change. Like it or not, callers of this function don't expect it to throw an error if they try to delete a nonexistent file

johncmerfeld, line 1138: IMO no reason to catch this only to throw it again. 

I mean honestly this function doesn't need to exist if we're using `fs.rmSync`. We should just get rid of it

src/handlers/ssh-proxy/ssh-proxy.handler.ts
```diff
@@ -90,7 +91,7 @@ export async function sshProxyHandler(
         if (createUniversalConnectionResponse.sshScpOnly) {
             exitCode = await bzeroTransparentSshProxyHandler(configService, logger, sshTunnelParameters, createUniversalConnectionResponse, loggerConfigService);
         } else {
-            exitCode = await bzeroOpaueSshProxyHandler(configService, logger, sshTunnelParameters, createUniversalConnectionResponse, loggerConfigService);
+            exitCode = await bzeroOpaqueSshProxyHandler(configService, logger, sshTunnelParameters, createUniversalConnectionResponse, loggerConfigService);
```
johncmerfeld, line 94: whoops

src/utils/daemon-utils.ts
```diff
@@ -35,25 +34,29 @@ const WAIT_UTIL_USED_ON_HOST_RETRY_TIME = 100;
  * spawns daemon as a subprocess with inherited stdio and returns a promise that
  * resolves when the daemon process exits with an exit code
  * @param logger the logger service to use to report errors if the daemon exits
- * @param path path to the daemon process
+ * @param daemonPath path to the daemon process
  * @param args args to pass to the daemon
  * @param cwd current working directory to use for the spawned subprocess
```
johncmerfeld, line 39: Can remove

src/utils/daemon-utils.ts
```diff
@@ -35,25 +34,29 @@ const WAIT_UTIL_USED_ON_HOST_RETRY_TIME = 100;
...
  * @param customEnv any custom environment variables to set for the spawned
  * process in addition to parent process environment
  * @returns A promise that resolves with the daemon process exit code
  */
-export function spawnDaemon(logger: Logger, loggerConfigService: LoggerConfigService, path: string, args: string[], customEnv: object, cwd: string): Promise<number> {
+export function spawnDaemon(logger: Logger, loggerConfigService: LoggerConfigService, daemonPath: string, args: string[], customEnv: object): Promise<number> {
```
johncmerfeld, line 44: This seems like a significant logic change. Why should we not be able to spawn daemons in whatever `cwd` we please?

src/utils/daemon-utils.ts
```diff
@@ -79,7 +82,7 @@ export async function spawnDaemonInBackground(logger: Logger, loggerConfigServic
         cwd: cwd,
         env: { ...customEnv, ...process.env },
         detached: true,
-        shell: true,
+        shell: false,
```
johncmerfeld, line 85: Reminding myself to test this code on macos. Is it possible that `shell` needs to be dependent on `platform`?

src/utils/daemon-utils.ts
```diff
@@ -252,12 +292,16 @@ export async function copyExecutableToLocalDir(logger: Logger, configPath: strin
...
     }
 
-    const daemonName = 'daemon-' + version;
+    const daemonName = 'daemon' + version;
     const configDir = path.dirname(configPath);
 
     let daemonExecPath: string;
     let finalDaemonPath: string;
-    if (process.platform === 'linux' || process.platform === 'darwin') {
+
```
johncmerfeld, line 300: I said elsewhere in a review that I disagree with the `if(process.platform === 'linux' || process.platform === 'darwin'` and I still feel that way here. We should have a windows check and otherwise execute the unix code

src/utils/daemon-utils.ts
```diff
@@ -171,26 +174,64 @@ export async function generateNewCert(pathToConfig: string, name: string, config
         const pathToCsr = path.join(pathToConfig, `${name}Csr${prefix}.pem`);
         const pathToCert = path.join(pathToConfig, `${name}Cert${prefix}.pem`);
 
-        // Generate a new key
-        try {
-            execSync(`openssl genrsa -out ${pathToKey}`, options);
-        } catch (e: any) {
-            reject(e);
-        }
+        const subject = [{
```
johncmerfeld, line 177: This whole certificate block should be split out into its own function

src/utils/utils.ts
lgmugnier: b0bc21e

src/utils/daemon-utils.ts
lgmugnier: b0bc21e

src/utils/daemon-utils.ts
```diff
@@ -157,7 +160,7 @@ export interface DaemonTLSCert {
  * @returns Path to the key, path to the cert, path to the certificate signing request.
  */
 export async function generateNewCert(pathToConfig: string, name: string, configName: string): Promise<DaemonTLSCert> {
-    const options: ExecSyncOptions = { stdio: 'ignore' };
+    // const options: ExecSyncOptions = { stdio: 'ignore' };
```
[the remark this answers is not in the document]
lgmugnier: removed! b0bc21e

src/utils/daemon-utils.ts
lgmugnier: unix system tests all pass and I couldn't find any reasoning for it. I haven't been able to break it yet but I did recently find a `hideWindow` option for spawn that I will try out so we can keep the option just in case. The reason I made it false on windows is because windows opens up a new window for each daemon

src/utils/daemon-utils.ts
lgmugnier: The reasoning behind this came from [this](https://stackoverflow.com/questions/21356372/node-child-process-spawn-not-working-with-spaces-in-path-on-windows) discussion. I was hitting the issue described and one of the commenters made it seem like this was the correct way to do it and I couldn't think of any reason why not. I haven't been able to find any more official documentation saying that he is right but this is something we have to do for windows and I was hoping I wouldn't need to have different behavior for different os's here

src/utils/daemon-utils.ts
```diff
@@ -252,12 +290,16 @@ export async function copyExecutableToLocalDir(logger: Logger, configPath: strin
         prefix = path.join(__dirname, '../../');
     }
 
-    const daemonName = 'daemon-' + version;
+    const daemonName = 'daemon' + version;
```
tflpd, line 293: Where else is this used? Could it create some backwards compatibility issues?
I believe not but I am far from an expert on this side of things

tflpd (approved):

src/handlers/generate/kube/generate-kube-yaml.handler.ts
```diff
@@ -49,7 +48,7 @@ export async function generateKubeYamlHandler(
 
     // Show it to the user or write to file
     if (outputFileArg) {
-        await util.promisify(fs.writeFile)(outputFileArg, kubeYaml.yaml);
+        fs.writeFileSync(outputFileArg, kubeYaml.yaml, { mode: 600 });
```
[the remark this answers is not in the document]
lgmugnier: maybe I misunderstand but why is my replacement a blocking implementation and the previous one isn't? It sounds like the original is better, could you explain why?

src/utils/daemon-utils.ts
lgmugnier: nope this is a good catch, I was messing with something when I hit a bug and I didn't clean it up correctly.

lgmugnier: although just to clarify, it doesn't hurt us in any way and won't cause any issues but it's an unnecessary change

src/handlers/generate/kube/generate-kube-yaml.handler.ts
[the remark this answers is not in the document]
lgmugnier: but we're awaiting that async function which means we're blocking on it until it concludes anyways. I don't understand the difference between awaiting an async function and a synchronous one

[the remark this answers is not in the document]
johncmerfeld: This cleared it up for me: https://stackoverflow.com/a/46004756/8414180

src/utils/daemon-utils.ts
lgmugnier: I can find no reason we would need `shell: true` to be the case here and via a lot of testing and running system tests, I'm pretty confident this does not change existing behavior. It is possible that it was needed in the past but is no longer.

src/handlers/generate/kube/generate-kube-yaml.handler.ts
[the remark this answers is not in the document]
lgmugnier: I guess then, it sounds like based on the file write function and where it is in the greater function and how it's running in the zli, I don't see any real difference between async or sync here. I'm happy to turn it back if it is a big deal but I see it as essentially equivalent. I didn't notice any performance issues at the human scale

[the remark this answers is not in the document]
johncmerfeld: I agree that this doesn't affect the performance of the ZLI as such. However, we ought to adhere to the principle of least change where we can, and changing it from async to sync falls outside of what the PR is about.

src/utils/daemon-utils.ts
johncmerfeld: Sounds good. It doesn't actually change which directory we're using for unix, and if we ever need to run daemons in specific directories we can revisit this matter.

src/utils/daemon-utils.ts
johncmerfeld: Confirmed, all good!

johncmerfeld (approved):

PR #659, opened 2023-04-10
Close shell connections on logout

johncmerfeld:
## Description of the change

Currently, shell sessions are the only connections that stay open after a `zli logout`. In the absence of a shell daemon manager 

## Testing

Open a shell session in one window, and log out in another

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Close shell connections on logout

## Related JIRA tickets

Relates to JIRA: CWC-2523

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:

- [ ] Yes
- [ ] No


johncmerfeld:
> Why `null` for web/db/kube?

They have their own mechanism for closing connections during logout, which I didn't want to disrupt. Still, I didn't want to make the `logoutDetector` param optional; whoever is working with daemons should be aware of what they're doing.

src/utils/daemon-utils.ts
```diff
@@ -72,9 +82,18 @@ export function spawnDaemon(logger: Logger, loggerConfigService: LoggerConfigSer
  * @param args daemon command line args
  * @param customEnv any custom environment variables to set for the spawned
  * process in addition to parent process environment
+ * @param logoutDetected an observable that fires when the user logs out
  * @returns The spawned child process
  */
-export async function spawnDaemonInBackground(logger: Logger, loggerConfigService: LoggerConfigService, cwd: string, daemonPath: string, args: string[], customEnv: object): Promise<cp.ChildProcess> {
+export async function spawnDaemonInBackground(
```
lgmugnier, line 88: Because none of the callers of this function will send `logoutDetected`, do we need it as a passed arg and instead just always pass it as `null` in the call to `reportDaemonExitErrors` below?

johncmerfeld: I think it belongs as a parameter here. What you're proposing would declutter some of the higher-level code that uses this function, but I don't think it's correct that it shouldn't be possible to pass a logout listener to this function.

However I did edit some of the doc strings to make it clearer that this is optional

tflpd (approved):

PR #662, opened 2023-04-11
The Config Issue

lgmugnier:
## Description of the change

Please see the referred ticket for a full explanation, but the short of it is that we need to put some of our credential values (`tokenset` and `mrtap`) into the windows registry.

## Testing

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [ ] Yes

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-2516

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:

- [ ] Yes
- [ ] No


src/services/logger/logger-config.service.ts
```diff
@@ -21,6 +21,9 @@ export class LoggerConfigService {
         // overlap and use the same configuration file.
         if(configDir) {
             configDir = path.join(configDir, projectName);
+        } else if (process.platform === 'win32') {
+            // conf defaults to roaming app data but our stuff belongs in local
+            configDir = path.join(process.env.LOCALAPPDATA, projectName)
```
johncmerfeld, line 26: Okay so even on windows we're still using conf for the logger configuration?

src/services/config/config.service.ts
```diff
@@ -94,7 +95,14 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
...
         const serviceUrl = this.buildServiceUrl(configName);
 
-        this.config = new UnixConfig(projectName, configName, configDir, isSystemTest, serviceUrl);
+        if (process.platform === 'win32') {
+            this.config = new WindowsConfig(logger, projectName, configName, configDir, isSystemTest, serviceUrl);
+        } else if (process.platform === 'linux' || process.platform === 'darwin') {
+            this.config = new UnixConfig(projectName, configName, configDir, isSystemTest, serviceUrl);
+        } else {
+            logger.error(`Unable to find config: unsupported operating system ${process.platform}`);
+        }
```
johncmerfeld, line 104: IMO there's no reason to exclude aix, freebsd, openbsd, and sunos from ZLI support; they're all unix-based, so this can just be 
```suggestion
        } else {
            this.config = new UnixConfig(projectName, configName, configDir, isSystemTest, serviceUrl);
        }
```

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+            hive: Registry.HKCU, // registry hive HKEY_CURRENT_USER
+            key:  WINDOWS_REGISTRY_KEY
+        });
+
+        // Create our Windows Registry key, no-op if already exists
+        this.regKey.create((err: Error) => {
+            if (err) {
+                logger.error(`failed to create new key in Window's Registry: ${err}`);
+            }
+        })
```
johncmerfeld, line 75: I think future engineers would appreciate some context here as to what we're doing and why. Even if it's just a couple well-chosen links

src/services/logger/logger-config.service.ts
johncmerfeld: Oh, I see, we're even still using it for most config values. Understood

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+
+        this.config = new Conf<ConfigSchema>({
```
johncmerfeld, line 77: Assuming we don't have to move literally everything to the registry, is there a way we can use less duplicated code between this `Conf` and the unix one? For example, we could treat this `ConfigSchema` as the base type and then extend it to include the extra Unix fields. Using that and some similar techniques, it seems like most of the next ~150 lines could be factored out. Maybe there are other reasons we don't want to do that, though?

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+                sshKnownHostsPath: undefined,
+                webConfig: getDefaultWebConfig(),
+                connectConfig: getDefaultConnectConfig(),
+                globalKubeConfig: getDefaultGlobalKubeConfig(),
+                dbDaemons: {},
+                kubeDaemons: {}
+            },
+            accessPropertiesByDotNotation: true,
+            clearInvalidConfig: true,    // if config is invalid, delete
+            migrations: {
```
johncmerfeld, line 107: In fact I don't even think we need these migrations, since no one will be using a windows ZLI prior to this version

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    async getTokenSet(): Promise<TokenSet> {
+        let tokenSet: TokenSet;
+
+        try {
+            const result = await this.getWindowsRegistryValue('tokenSet');
+            tokenSet = JSON.parse(result);
+        } catch (err) {
+            // We hit this error if we're getting the token before it is set, for example
+            // after it has been cleared. We can safely ignore it.
```
johncmerfeld, line 292: Not sure I understand this

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+        // this.config.onDidChange('tokenSet', callback);
+    }
+
+    getWhoami(): SubjectSummary {
+        return this.config.get('whoami');
+    }
+
+    getGaToken(): string {
+        return this.config.get('GAToken');
+    }
```
johncmerfeld, line 249: Again, it's possible this is more trouble than it's worth, but I'm wondering whether we shouldn't just do this the old-fashioned inheritance way and have a parent class that implements all of these `config` getters/setters, and then a Windows child class that overrides only the methods it needs to. What do you think?

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+            this.regKey.get(key, (err: Error, result: Registry.RegistryItem) => {
+                if (err) {
+                    reject(err);
+                }
+
+                if (result != null) {
+                    resolve(result.value);
+                } else {
+                    // If our result is null then it means the value exists but it's been cleared
+                    resolve('');
```
johncmerfeld, line 278: Why is `''` the correct return here? Or rather, why does this function return a `Promise<string>` instead of a `Promise<string | null>`?

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+        this.config.delete('sessionToken');
+    }
+
+    clearClientSecret(): void {
+        this.config.delete('clientSecret');
+    }
+
+    clearTokenSet(): void {
+        this.regKey.remove('tokenSet', (err: Error) => {
+            // Ignore errors if the key can't be found which is normal
```
johncmerfeld, line 467: lol, why is this normal?

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    setKubeDaemons(kubeDaemons: DaemonConfigs<KubeConfig>): void {
+        this.config.set('kubeDaemons', kubeDaemons);
+    }
+
+    setMrtap(data: MrtapConfigSchema): void {
+        this.regKey.set('mrtap', Registry.REG_SZ, JSON.stringify(data), (err: Error) => {
+            if (err) {
+                this.logger.error(`failed to save mrtap config: ${err}`);
+            }
```
johncmerfeld, line 444: This seems like a fatal error, no? I'm curious why we're essentially ignoring it

src/services/logger/logger-config.service.ts
lgmugnier: yes, exactly. I really wanted to just do the bare minimum for now. It's possible there is a future PR where everything moves to the windows registry but until I have proof that's what  we need to do, it'll stay with conf.

src/services/config/config.service.ts
lgmugnier: I don't disagree with this but I don't know if we do want to make this change because it shifts some os's into an "implicit support" category which we don't have any explicit testing for in our matrix. I think we should get a thumbs up from the team before making the change and then make sure it's consistent across all of these such if statements.

src/services/config/windows-config.service.ts
lgmugnier: I added some more context in 1f18238 but I'm not sure how much more to clarify without pasting wikipedia links about how the windows registry works unless you think a little blurb about how the windows registry works is helpful here? something like:
"The Windows Registry is key-value storage. It has "hives" which are the root keys and generally separate out further nesting into priveleged categories. Our key lives in the application ("Software") section in a special new subkey "BastionZero""

src/services/config/windows-config.service.ts
lgmugnier: Please check 1f18238 to see if I handled the error correctly. Essentially, now I throw errors if we hit one and pass that up until I get to a function that can call `cleanExit`

src/services/config/windows-config.service.ts
lgmugnier: By virtue of how remove works, we often hit the error that the key we're looking for doesn't exist. I think it was a bit confusing before so I've added a new function to check if we're hitting that kind of error and not report if that's the case. I also consolidate this exact point into a comment in that function. Check it out to see if it's any clearer 1f18238

src/services/config/windows-config.service.ts
lgmugnier: I wanted to consolidate your comments in this vein into one so I chose this comment. 

This is a great idea, I think I refrained from doing that because 1 I was poisoned by go but 2 because I did think that moving everything to the windows registry was a very real prospect. I'm happy to make this change now even if we do still need to move everything to the registry. And I'm happy to be doing so in a language that has good, traditional inheritance.

src/services/config/windows-config.service.ts
johncmerfeld: I actually think that would be super helpful, yeah!

src/services/config/windows-config.service.ts
johncmerfeld: I see. Yep, appreciate the additional info in the code and the function is a good idea!

src/handlers/middleware.handler.ts
```diff
@@ -79,6 +79,11 @@ export async function bzCertValidationInfoMiddleware(mrtapService: MrtapService,
         // Update the Org BZCert Validation parameters
         const orgHttpService = new OrganizationHttpService(configService, logger);
         const orgBZCertValidationInfo = await orgHttpService.GetUserOrganizationBZCertValidationInfo();
-        mrtapService.setOrgBZCertValidationInfo(orgBZCertValidationInfo);
+
+        try {
+            mrtapService.setOrgBZCertValidationInfo(orgBZCertValidationInfo);
+        } catch(e) {
+            throw e;
```
johncmerfeld, line 86: No need to try/catch/throw. If setOrgBZCertValidationInfo throws an error, it'll get thrown automatically here

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,496 @@
...
+        });
+    }
+
+    private isKeyNotFoundError(err: Error): boolean {
+        // Sometimes we try to clear or get a token that has already been cleared, which means 
+        // we'll hit an error letting us know the key doesn't exist. This is a result of the 
+        // nature of the winreg remove() function.
+        // ref: https://fresc81.github.io/node-winreg/Registry.html#remove__anchor
+        return err.message.indexOf('The system was unable to find the specified registry key or value') != -1
```
johncmerfeld, line 494:
```suggestion
        return err.message.includes('The system was unable to find the specified registry key or value');
```

src/services/config/windows-config.service.ts
johncmerfeld: Well-handled, I think! But added one more comment

src/services/config/config.service.ts
tflpd: For what is worth, I agree that we should only be using what we are testing. There should be a clear, Product-specified list of OSs we support and have our code support them and test for them in some way. It may be cumbersome to test for them in every PR so we could just test all of them in the CI, but still, that's the proper way of doing things

src/services/config/windows-config.service.ts
lgmugnier: b3429fd

src/handlers/middleware.handler.ts
lgmugnier: wasn't sure about that so thank you for clarifying!

src/services/config/config.service.ts
lgmugnier: alright, conclusion from the meeting sides with @johncmerfeld. We can easily put all of our unix support into an "else" statement.

src/services/config/windows-config.service.ts
lgmugnier: 2077e00

src/handlers/middleware.handler.ts
lgmugnier: 2077e00

src/services/config/config.service.ts
lgmugnier: 2077e00

tflpd (changes requested):

src/handlers/login/login.handler.ts
```diff
@@ -66,19 +66,25 @@ export async function loginUserHandler(configService: ConfigService, logger: Log
...
+            // Pass it in as we login
+            await oAuthService.login(async (t) => {
+                configService.setTokenSet(t);
+                mrtapService.setInitialIdToken(await configService.getIdToken());
+            }, nonce);
+        }
+    } catch (e) {
+        logger.error(`failed to login: ${e}`);
+        return undefined;
     }
```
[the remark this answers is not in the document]
lgmugnier: because `setTokenSet` can also throw an error

src/handlers/login/login.handler.ts
```diff
@@ -66,19 +66,25 @@ export async function loginUserHandler(configService: ConfigService, logger: Log
...
+            // Create our Nonce
+            const nonce = mrtapService.createNonce();
+            // Pass it in as we login
+            await oAuthService.login(async (t) => {
+                configService.setTokenSet(t);
+                mrtapService.setInitialIdToken(await configService.getIdToken());
+            }, nonce);
+        }
+    } catch (e) {
+        logger.error(`failed to login: ${e}`);
```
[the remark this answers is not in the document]
lgmugnier: checking with @asamborski about the error messages

src/cli-driver.ts
```diff
@@ -328,7 +328,13 @@ export class CliDriver
...
                 if(!this.oauthCommands.has(baseCmd) || baseCmd === 'register' || isServiceAccountLogin)
                     return;
-                await bzCertValidationInfoMiddleware(this.mrtapService, this.configService, this.logger);
+
+                try {
+                    await bzCertValidationInfoMiddleware(this.mrtapService, this.configService, this.logger);
+                } catch(e) {
+                    this.logger.error(`failed to save new mrtap configuration data: ${e}`);
+                    await cleanExit(1, this.logger);
+                }
```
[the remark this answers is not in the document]
lgmugnier: 67a122c

src/handlers/login/login.handler.ts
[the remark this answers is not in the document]
lgmugnier: checked and made changes in 67a122c

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,496 @@
...
+                    }
+                }
+            },
+            watch: watch
+        });
+        this.path = this.config.path;
+    }
+
+    onTokenSetChange(callback: (newValue?: TokenSetParameters, oldValue?: TokenSetParameters) => void): void {
+        // this.config.onDidChange('tokenSet', callback);
```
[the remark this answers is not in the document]
lgmugnier: 398548a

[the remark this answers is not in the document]
tflpd: should `logoutOnTokenSetCleared` be added in the interface?

src/handlers/login/login.handler.ts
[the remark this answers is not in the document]
lgmugnier: 76616a4 I've fixed it so that they start with a capital letter but I don't understand why error would need to end in a period. It doesn't seem like we have that convention overall and I disagree that we would want it particularly since errors are things that get bubbled up and it's hard to keep track of punctuation.

src/handlers/register/register.handler.ts
```diff
@@ -17,7 +17,13 @@ export async function  registerHandler(mfaSecret: string, configService: ConfigS
     // Force refresh ID token and access token because it is likely expired in system tests
     // We are force refreshing the tokens so register does not need to be an oauth command
     const newTokenSet = await oauthService.refresh();
-    configService.setTokenSet(newTokenSet);
+
+    try {
+        configService.setTokenSet(newTokenSet);
+    } catch (e) {
+        logger.error(`failed to save tokens after oath refresh: ${e}`);
```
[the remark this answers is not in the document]
lgmugnier: 76616a4

src/handlers/login/login.handler.ts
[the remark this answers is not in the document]
lgmugnier: They do get concatenated if you have a `reject(blah)` and then a `catch (e) { logger.error('hit error: ${e}') }`. 

In this situation, there are very few periods in the login flow messages and I think having punctuation only at the end of error messages is awkward.

There is a clear lack of cohesion in the existing code base, we should just get a consensus decision and fix everything at once in a separate PR.

src/services/config/windows-config.service.ts
lgmugnier: e7f4ed7

src/services/config/windows-config.service.ts
```diff
@@ -0,0 +1,512 @@
...
+                    resolve(undefined);
+                }
+            });
+        });
+    }
+
+    private async setWindowsRegistryValue(key: string, value: any): Promise<void> {
+        return new Promise((resolve, reject) => {
+            this.regKey.set(key, Registry.REG_SZ, JSON.stringify(value), (err: Error) => {
+                if (!err) {
```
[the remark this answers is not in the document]
lgmugnier: 4b7b154

tflpd (approved):

johncmerfeld (approved):
Love this!!

PR #664, opened 2023-04-20
Keycloak

SaiDadireddy:
## Description of the change

## Testing

**backend branch:** feat/keycloak
**bzero branch:** feat/keycloak
**charts branch:** 

- [ ] Yes

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-2517

Relevant HotFix PR Number:

- [X] Yes
- [ ] No

If yes, please explain:
Adds a new way to authenticate against BastionZero via Keycloak. This pr covers adds an authentication layer to validate the tokens we receive from the Keycloak oidc client.

- [ ] Yes
- [ ] No


Codebuild/system-test-codebuild.yaml
```diff
@@ -39,7 +39,7 @@ phases:
       - git submodule init
       - git submodule update --recursive
       # Install Bzero-QA
-      - git clone git@github.com:bastionzero/cwc-infra.git --branch $CWC_INFRA_BRANCH /cwc-infra/
+      - git clone git@github.com:bastionzero/cwc-infra.git --branch feat/keycloak /cwc-infra/
```
SaiDadireddy, line 42: I will revert this once the infra branch is merged into develop

src/services/config/config.service.ts
```diff
@@ -212,6 +213,8 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
             return 'offline_access openid email profile';
         case IdentityProvider.OneLogin:
             return 'openid profile';
+        case IdentityProvider.Keycloak:
+            return 'openid';
```
SaiDadireddy, line 217: Just to note here - the other scopes including `offline_access` and `profile` are returned by default. We can configure which scopes are default and which are optional (meaning we need to specify in the request) in respective client in Keycloak.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bastionzero/zli",
-  "version": "6.19.1",
+  "version": "6.19.2",
```
lipmas, line 3: bump minor version

src/services/config/config.service.ts
lipmas: > We can configure which scopes are default and which are optional (meaning we need to specify in the request) in the respective client in Keycloak.

But does this mean that the customer needs to set these scopes as default in order to work with the zli? What is the downside to including the scopes anyway.

SaiDadireddy: There's no downside to adding them. [8e7b7e6](https://github.com/bastionzero/zli/pull/664/commits/8e7b7e621bd2e80e6cf03d066a9cf9f05df8927f)

tflpd (approved):

lipmas (approved):

PR #673, opened 2023-04-28
MFA Grace Period Integration and UX Improvements

chad-cwc:
## Description of the change

This PR integrates the MFA grace period feature into the ZLI. The goals were to convey that there is a grace period, to allow the user to defer setting up MFA during their grace period, and to provide a UX that is similar to what was already done in the web app.

Other UX changes:
- I changed/added text in various places to improve the UX or to match what is in the web app
- I changed the browser behavior for MFA scenarios, so that the browser doesn't hang and wait for the ZLI MFA interaction to complete. It seemed potentially confusing to the user that the browser window was blank. It didn't seem intuitive. Telling the user the SSO login was successful and directing them to the ZLI seems like a better UX.
- I changed the QR code to be small. But this only looks good on Windows and Linux, so the small QR code is only shown on those platforms, and the large one is used for OS X.

## Testing

1. Enable MFA for a user and log in using the ZLI. When asked about setting up MFA, type 'n'. After saying no, verify that authentication completed successfully by running another command. Log out and log back in with that user to verify that the same MFA setup question is asked.
2. Enable MFA for the same user and log in using the ZLI. When asked about setting up MFA, type 'y'. After saying yes, verify that authentication completed successfully by running another command. Log out and log back in with that user to verify that the MFA setup did indeed complete by seeing the MFA code prompt instead of the MFA setup flow.
3. Disable MFA for the same user and make sure no MFA actions are presented. 

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Allow for MFA grace period in ZLI

## Related JIRA tickets

Relates to JIRA: CWC-2521

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


SaiDadireddy (commented):
Tested out the PR.
1. Works as expected. I set logged in through the zli and typed `n` which allowed me to bypass mfa and I see the message `You have 14 days until MFA setup is mandatory.`
2. For some reason, my google authenticator app isn't able to recognize the QR code that's generated. This is odd because it's able to recognize the QR code when I log in through the webapp (I use the Warp terminal but I tried this on the mac's builtin terminal as well and same thing). Also, I'm not sure if this is expected, but when I crtl + C and exit from the QR code prompt, it logs me in successfully, with the message `MFA setup bypassed. Login successful.`
3. Works as expected

SaiDadireddy:
> Wow. That's interesting and unexpected. I changed the size to be smaller. Have you done the QR code scanning successfully in the terminal in the past? I'm on Linux and had no problem with either size of QR code.

Yep, I remember it working in the past but that was not recently.

chad-cwc:
> > Wow. That's interesting and unexpected. I changed the size to be smaller. Have you done the QR code scanning successfully in the terminal in the past? I'm on Linux and had no problem with either size of QR code.
> 
> Yep, I remember it working in the past but I haven't tried it recently.

I asked Ann to try it on her Mac, too, and she had the same result. She also verified that a released version of the ZLI does work. That made it pretty clear that it's likely the new `small` option I added. Then, I found [even better proof](https://github.com/soldair/node-qrcode/issues/322) on the GitHub of the library we're using. I will work with Product to decide if we try to find a different library or revert back to the larger QR code.

src/services/oauth/oauth.service.ts
```diff
@@ -141,11 +141,12 @@ export class OAuthService implements IDisposable {
 
                 this.logger.info('Login successful');
                 res.write(loginHtml);
+                res.end();
```
tflpd, line 144: Just curious, what change did this bring? I went through the flow both with the old and the new code and coulndt tell

tflpd (approved):

chad-cwc:
> > > Wow. That's interesting and unexpected. I changed the size to be smaller. Have you done the QR code scanning successfully in the terminal in the past? I'm on Linux and had no problem with either size of QR code.
> > 
> > 
> > Yep, I remember it working in the past but I haven't tried it recently.
> 
> I asked Ann to try it on her Mac, too, and she had the same result. She also verified that a released version of the ZLI does work. That made it pretty clear that it's likely the new `small` option I added. Then, I found [even better proof](https://github.com/soldair/node-qrcode/issues/322) on the GitHub of the library we're using. I will work with Product to decide if we try to find a different library or revert back to the larger QR code.

To close the loop on this, I suggested showing the smaller QR code only where it looks good and using the default, larger one on OS X. I also added the secret key as a fallback mechanism in case the QR code is unexpectedly unusable again somewhere else. Product agreed to these changes.

SaiDadireddy (approved):

PR #680, opened 2023-05-03
Refresh lock

lgmugnier:
## Description of the change

Put a lock around our refresh lock so that when multiple daemons are refreshing, they do not conflict and it doesn't accidentally wipe your creds.

## Testing

1. Login
2. Wait 1-2 hours
3. Run the following script:
```
#!/usr/bin/env python3
from multiprocessing import Process
from subprocess import call 
if __name__ == '__main__':
    for _ in range(10):
        command = "echo hello"
        target_name = f'root@lucie-bzero-agent' # Replace with correct name 
        p = Process(target=call, args=(['ssh', target_name, command],))
        p.start()
```

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Make refresh safe

## Related JIRA tickets

Relates to JIRA: CWC-2563

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


tflpd (changes requested):
Should we add a lock around register handler as well?

src/services/oauth/oauth.service.ts
```diff
@@ -17,6 +18,9 @@ import { UserHttpService } from '../../../src/http-services/user/user.http-servi
 import { SubjectType } from '../../../webshell-common-ts/http/v2/common.types/subject.types';
 import { SubjectHttpService } from '../../../src/http-services/subject/subject.http-services';
 
+const lockfile = require('proper-lockfile');
+const retry = require('retry');
```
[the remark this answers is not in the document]
lgmugnier: bad93f0

lgmugnier:
>Should we add a lock around register handler as well? 
When is `zli register` called? If it's only called during system tests then I think we're good.

src/services/oauth/oauth.service.ts
```diff
@@ -333,45 +337,79 @@ export class OAuthService implements IDisposable {
...
-
         // Refresh if the token exists and has expired
-        if (tokenSet) {
-            if (!await this.isAuthenticated()) {
-                if(this.configService.me().type === SubjectType.ServiceAccount){
-                    this.logger.error('Service account session has expired, please log in again.');
-                    await cleanExit(1, this.logger);
+        if(this.configService.me().type === SubjectType.ServiceAccount){
+            throw new Error('Service account session has expired, please log in again.');
+        }
```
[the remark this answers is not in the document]
tflpd: Yup! SAs do not have refresh logic, when their idT expires (once every week atm), they need to log in again

src/services/oauth/oauth.service.ts
```diff
@@ -333,45 +335,77 @@ export class OAuthService implements IDisposable {
...
-                    this.logger.debug(`Refresh Token Error: ${e.message}`);
+                    release();
+
+                    this.logger.debug(`Failed to refresh OAuth tokens. ${e.message}`);
                     if (e instanceof errors.RPError || e instanceof errors.OPError) {
                         throw new RefreshTokenError();
                     } else {
                         throw e;
                     }
                 }
```
tflpd, line 385: If something like this works it would be preferred as Promises are the old way of doing async/await
```
        const release = await lockfile.lock(lockName, {
            realpath: false,
            stale: 5000, // 5 seconds
            retries: {
                retries: 20,
                minTimeout: 300, // The number of milliseconds before starting the first retry
                maxTimeout: 500, // The maximum number of milliseconds between two retries
                maxRetryTime: 5 * 5000 // The maximum time (in milliseconds) that the retried operation is allowed to run
            }
        })

        // Maybe another process refreshed while we were waiting, so we check again
        // in case we don't need to go through the refresh process
        if (await this.isAuthenticated()) {
            tokenSet = await this.configService.getTokenSet();
            await release();
            return tokenSet.id_token
        }

        this.logger.debug('Refreshing OAuth tokens');
        try {
            tokenSet = await this.refresh();
        } catch (e) {
            release();

            this.logger.debug(`Failed to refresh OAuth tokens. ${e.message}`);
            if (e instanceof errors.RPError || e instanceof errors.OPError) {
                throw new RefreshTokenError();
            } else {
                throw e;
            }
        }

        this.configService.setTokenSet(tokenSet);
        this.logger.debug('OAuth tokens refreshed');

        const userHttpService = new UserHttpService(this.configService, this.logger);
        await userHttpService.Register();

        // Update me section of the config in case this is a new login or any
        // user information has changed since last login
        const subjectHttpService = new SubjectHttpService(this.configService, this.logger);
        const me = await subjectHttpService.Me();
        this.configService.setMe(me);

        await release();
        return tokenSet.id_token;
```

tflpd: Note that I have altogether removed the catch block you were using - up to you whether you want more granular control and add one per lock/release or add one that covers all

lgmugnier: 001c74a

johncmerfeld (approved):
Logged in around 7am. Didn't do anything else. Ran the script around 11am and it worked

tflpd (approved):
Nice, some comments need updating, LGTM other than that!

src/services/oauth/oauth.service.ts
tflpd: Oh, glad it works, I wasn't 100% it will!

lipmas (approved):
Tested using the python script on linux and its working.

lipmas:
@lgmugnier Added the `ci-run-google` label to the PR now which it doesnt run by default if you add other ci labels. But this is a required status check. But you will need to push one more commit in order to get it to run that CI.

PR #695, opened 2023-05-25
Add script for removing daemon executables from zli config directory

lipmas:
## Description of the change

Adds a new npm script that removes daemon executables from the zli config directory. I made this a standalone nodejs script so that:

1. It is compatible across all OSs (including windows in principal - although i have not tested)
2. It can reuse the same code the zli uses for both constructing the zli config directory (respecting env variable overrides such as  `ZLI_CONFIG_NAME` and `ZLI_CONFIG_DIR`) and the final daemon executable filename (i.e `daemon-{version}` with version injected from `package.json`)

One note is I needed to nest this script within the `src` directory so that imports would work without any further modifications to `tsconfig.json` but if we wanted to move it up a layer we probably could do that as well if we included it [here](https://github.com/bastionzero/zli/blob/8322de29687105be43f077094613b45a57713b36/tsconfig.json#L63).

## Testing

First make sure there is an existing daemon executable for the current zli version copied to the zli config directory. Running any `zli connect` command should do. Then run `npm run release`. Under-the-hood this will call `npm run clean` which in turn calls the new `npm run clean-daemon` script that removes the daemon executable from the config directory (there will also be std output that shows this executable being removed). The next `zli connect` command that runs should then re-copy the newly compiled daemon as is expected during development workflow.  

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Adds npm script to improve development workflow when making changes to daemon

## Related JIRA tickets

Relates to JIRA: CWC-2640

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


johncmerfeld:
I'm afraid this hangs on Windows:
image

So I tried running the script directly with `ts-node`. That command wasn't recognzied, so I tried `npx ts-node`, which… also hangs
image

Happy to help you debug this if you'd like. Not sure what the root cause might be although I haven't looked closely at the code. Sorry!

johncmerfeld (approved):
Walked through this with Sebby. Turns out we just needed to add a `process.exit(0)` at the end of the cleaning script. Great stuff!

PR #710, opened 2023-06-14
LevelDB

lgmugnier:
## Description of the change

This moves us to LevelDB. Details about why LevelDB was chosen are in the ticket. 

Notable changes:

1. In order to set a custom serviceUrl (such as `https://lucie.bastionzero.com/`) you need to set an environment variable `ZLI_SERVICE_URL`.

## Testing

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Moves our config out of a file and into leveldb. Unix users don't have to do anything since it will all be backwards compatible, but windows users WILL have to login after upgrading.

## Related JIRA tickets

Relates to JIRA: CWC-2562

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [ ] No


Codebuild/system-test-codebuild.yaml
```diff
@@ -36,7 +36,7 @@ phases:
       - git submodule init
       - git submodule update --recursive
       # Install Bzero-QA
-      - git clone git@github.com:bastionzero/cwc-infra.git --branch $CWC_INFRA_BRANCH /cwc-infra/
+      - git clone git@github.com:bastionzero/cwc-infra.git --branch 'leveldb' /cwc-infra/
```
lgmugnier, line 39: temporary until we merge in cwc-infra

src/services/config/config.service.ts
```diff
@@ -101,38 +62,72 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
...
-            process.exit(1);
+        // Check if our old config exists and only migrate it if we're on macos and linux
+        if (process.platform !== 'win32' && LegacyConfigExists(configDir, projectName, configName)) {
+            this.migrateLegacyConfig(configName, configDir, projectName);
         }
 
         this.tokenHttpService = new TokenHttpService(this, logger);
+
+        // TODO: Check with the powers that be what we're supposed to do with this
+        if (isSystemTest) {}
```
lgmugnier, line 83: check for explanation of why we needed this for system tests. vanilla ones certainly pass without me doing anything

src/services/config/config.service.ts
```diff
@@ -1,96 +1,57 @@
...
+
+// All keys defined for our key, value store
+const authUrlKey: string = 'authUrl';
+const clientIdKey: string = 'clientId';
+const clientSecretKey: string = 'clientSecret';
+const authScopesKey: string = 'authScopes';
+const serviceUrlKey: string = 'serviceUrl';
+const tokenSetKey: string = 'tokenSet';
+const callbackListenerPortKey: string = 'callbackListenerPort';
+const gaTokenKey: string = 'GAToken'; // LUCIE: we should make this camelCase to match everything else
```
lgmugnier, line 27: I think this should be 'GaToken' based on how we capitalize things elsewhere?

src/services/config/config.service.ts
```diff
@@ -229,205 +216,337 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
...
-        return knownHostsPath;
-    }
-
     getConfigName(): string {
         return this.configName;
     }
 
     getConfigPath(): string {
-        return this.configPath;
+        return path.dirname(this.store.where);
```
lipmas, line 246: There are a couple places [here](https://github.com/bastionzero/zli/blob/8322de29687105be43f077094613b45a57713b36/src/handlers/configure/config.handler.ts#L7) and [here](https://github.com/bastionzero/zli/blob/8322de29687105be43f077094613b45a57713b36/src/handlers/configure/config-default-targetuser.handler.ts#L20) where we use this function to print a log message which doesnt make sense anymore now that we dont have a config file

src/services/config/config.service.ts
lipmas: This was only used to set a watch mode on the config file so that we could implement push notifications when logout was detected. Otherwise it wasnt needed and can be removed.

However it looks like these are no longer used in the config service now

```
private logoutDetectedSubject: Subject<boolean> = new Subject<boolean>();
public logoutDetected: Observable<boolean> = this.logoutDetectedSubject.asObservable();
```

Is there a plan to address this - because @johncmerfeld relies on this in the ssh proxy handler [here](https://github.com/bastionzero/zli/blob/d81ca538bbb873bbf835325c44c779c1cfe9793c/src/handlers/ssh-proxy/ssh-proxy.handler.ts#L141-L145). One simple fix could just be to convert those references to a simple poller mechanism instead.

lgmugnier: yes I will address this in an upcoming PR, will post when it's ready.

src/services/config/config.service.ts
lgmugnier: First instance is fixed in 2ad7d54, second guy is actually fixed in separate PR: https://github.com/bastionzero/zli/pull/716

johncmerfeld (approved):
Tested this out on MacOS and Windows and it's looking great.

Just have a couple thoughts of other stuff you might do:
  - (had comments about `zli configure` but we addressed during standup. You'll add `get/set` commands, and maybe update the text of the base `zli configure` command so that it still prints the directory but makes it clear you can't edit directly?)
  - Should we remove `zli refresh` from the command roster since it will no longer be used?

src/services/config/leveldb/leveldb.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (db.status === 'closed') {
+            try {
+                await db.open();
+            } catch (err) {
+                if (err instanceof ModuleError) {
+                    lastErrorCode = err.code;
+                    if (err.cause && err.cause.code === 'LEVEL_LOCKED') {
+                        // Another process or instance has opened the database
+                    }
+                }
```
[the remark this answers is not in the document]
lgmugnier: They do not, they are merely incredibly helpful for future devs. This is more documentation if there's something wrong, I wanted to leave breadcrumbs

src/services/config/config.service.ts
```diff
@@ -105,88 +60,110 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
...
+        this.setWhoami(config.getWhoami());
+        this.store.setSync(sshKeyPathKey, config.getSshKeyPath());
+        this.store.setSync(sshKnownHostsPathKey, config.getSshKnownHostsPath());
+        this.setMrtap(config.getMrtap());
+        this.setWebConfig(config.getWebConfig());
+        this.setConnectConfig(config.getConnectConfig());
+        this.setGlobalKubeConfig(config.getGlobalKubeConfig());
+        this.setDbDaemons(config.getDbDaemons());
+        this.setKubeDaemons(config.getKubeDaemons());
+        this.setRDPDaemons(config.getRDPDaemons());
```
lipmas, line 101: Looks like these two getters from the legacy config are the only ones missing here: `getTcpAppPortsConfig` and `getSQLServerDaemons`. Was this intentional or should we add those too?

src/services/config/leveldb/leveldb.ts
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
```
lipmas, line 117: What edge case is this covering? It looks like from the [docs](https://github.com/Level/classic-level#dbopenoptions-callback) explicitly calling `open()` shouldnt be needed as its done in the constructor. But here we call it if `db.status` is `closed` - wouldnt this only happen if we explicitly called `close()` somewhere?

lipmas (approved):
Tested this on linux and the migration worked flawlessly. I also verified using [leveldb-cli](https://github.com/liderman/leveldb-cli) that all my legacy config values appeared in the leveldb database (apart from the 2 i mentioned below).

src/services/config/config.service.ts
lipmas: One more minor user facing message to fix is the configure command help message https://github.com/bastionzero/zli/blob/8a19c584705d7f30842aa8af39abd5870cb8f0a2/src/cli-driver.ts#L421-L424

src/services/config/config.service.ts
lgmugnier: nope, that's a great catch. Those were added after I had written this logic and I forgot to add them after merging in develop

PR #725, opened 2023-06-23
Kill concurrent zlis

lgmugnier:
## Description of the change

Main purpose of this PR is to use IPC channels to kill concurrent ZLIs on logout. This is something we had previously done by listening or polling the config file which could easily lead to unnecessary blocks.

The other two things this does are 
1. Fix a pet peeve where we were assuming there was a trailing / in the service url
2. Fix some random wording originally taken care of the now rejected PR to add a new command to set callback port

## Testing
Build and then login to the zli. Open up several windows with `zli connect ..` and then `zli logout` in a different window.

Make sure your serviceUrl is set correctly by either setting
`export ZLI_SERVICE_URL=https://name.bastionzero.com`
`zli configure set serviceUrl https://name.bastionzero.com`

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [x] Yes

Release Notes: Uses IPC channels to kill concurrent ZLIs on logout

## Related JIRA tickets

Relates to JIRA: CWC-2634

Relevant HotFix PR Number:

- [ ] Yes
- [x] No

If yes, please explain:

- [ ] Yes
- [x] No


src/services/config/config.service.ts
```diff
@@ -3,6 +3,7 @@ import fs from 'fs';
 
 import { TokenSet } from 'openid-client';
 import { Observable, Subject } from 'rxjs';
+// import ipc from 'node-ipc';
```
johncmerfeld, line 6: Can nix

src/services/config/config.service.ts
```diff
@@ -157,6 +265,10 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
     }
 
     logout(): void {
+        if (!this.isSystemTest) {
+            this.notifyLogout();
+        }
```
johncmerfeld, line 270: Probably worth a comment addressing why we need this

johncmerfeld (changes requested):
Tested this on Windows and it's not logging me out, unfortunately

src/services/config/config.service.ts
lgmugnier: honestly, I'm not entirely sure. I just was keeping with what was happening before where (I believe) we wouldn't listen for logouts if we were in system tests but maybe I misinterpreted? @lipmas

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bastionzero/zli",
-  "version": "6.24.2",
+  "version": "6.24.3",
```
johncmerfeld, line 3: Due to the new commands this should definitely be 6.25.0

johncmerfeld (approved):
Tested on Windows and it's working great -- this is a cool implementation that seems easy to manage going forward. Woop!

PR #726, opened 2023-06-23
Add stand alone daemon refresh [CWC-2700]

tflpd:
## Description of the change

The daemon currently refreshes its tokens invoking a zli command. However, now that the desktop app uses the daemon without having a zli sidecar, the daemon needs to be able to detect and refresh tokens on its own.

## Testing

The easiest way to force the daemon to refresh the tokens is commenting out [this](https://github.com/bastionzero/bzero/blob/f5e0d8db5a3647b0a5f42eafbaf8c7a7a7dd7e63/bctl/daemon/mrtap/bzcert/bzcert.go#L60-L62) line. After doing that, try making a connection to a target. The daemon will refresh the tokens and store them. The connection should go through no matter what IdP you are using and both the daemon and future zli commands should continue working. Feel free to use `zli configure get tokenSet` to ensure new values are valid and available.

**backend branch:** 
**bzero branch:**
**charts branch:** 

- [x] Yes

Release Notes: Add stand alone daemon refresh

## Related JIRA tickets

Relates to JIRA: CWC-2700

Relevant HotFix PR Number:

- [x] Yes
- [ ] No

If yes, please explain:

We need to make sure that the refresh process is secure, we do not log or leak the tokens acquired through the process and the zli is still able to continue using them.

- [ ] Yes
- [x] No


johncmerfeld (approved):
Tested and looks good to me. I think we should remove the `refresh` command altogether though

src/services/config/config.service.ts
```diff
@@ -319,6 +320,10 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
         return idp;
     }
 
+    getExternalIdpId(): string {
+        return this.store.getSync(externalIdpIdKey);
+    }
+
```
lipmas, line 326: Discussed offline but this is already part of the [mrtap config](https://github.com/bastionzero/webshell-common-ts/blob/b0711d98a9fa76eb99369cfbe8f9b794e11602a4/mrtap.service/mrtap.service.types.ts#L7) so not needed.

lipmas (approved):

johncmerfeld (approved):

PR #730, opened 2023-07-06
Fix zli oauth refresh race condition 

lipmas:
## Testing

**backend branch:** 
**bzero branch:** 
**charts branch:** 

- [ ] Yes

Release Notes:

## Related JIRA tickets

Relates to JIRA: CWC-XXX

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:

- [ ] Yes
- [ ] No


src/services/daemon-management/daemon-management.service.ts
```diff
@@ -70,18 +71,18 @@ export class DaemonManagementService<T extends DaemonConfig> {
...
+    public async addDaemon(connectionId: string, config: T) {
         if (!connectionId) {
             throw new Error('Cannot add a daemon to local store if connection ID is undefined');
         }
 
-        const daemons = this.daemonStore.getDaemons();
+        const daemons = await this.daemonStore.getDaemons();
         daemons[connectionId] = config;
-        this.daemonStore.setDaemons(daemons);
+        await this.daemonStore.setDaemons(daemons);
```
ymarcus93, line 81: If we call `lockDb` before `getDaemons` and then `unlockDb` after `setDaemons`, then I think we'll squash the race condition / bug that produces an inconsistent zli config state (specifically `zli ld`) when attempting to run `zli connect` in parallel for DB, Kube, etc.

lipmas: [61c042b](https://github.com/bastionzero/zli/pull/730/commits/61c042bccfbffeef3decdd0ce16abe9f537048da)

src/services/config/leveldb/leveldb.ts
```diff
@@ -58,55 +48,84 @@ export class Store {
...
+        } finally {
+            this.db = undefined;
         }
     }
 
-    private set: any;
-    public setSync(key: string, value: string): void {
-        if (!value) {
-            return this.clearSync(key);
+    public async getAsync(key: string): Promise<string> {
```
johncmerfeld, line 72: I know these were previously called `__Sync` but I'm not crazy about having `Async` in the name if we don't also have a synchronous option. Go with your heart on this but I'd rather we just use `get`/`set`/`clear`

src/services/daemon-management/daemon-management.service.ts
```diff
@@ -70,29 +76,39 @@ export class DaemonManagementService<T extends DaemonConfig> {
...
-    public addDaemon(connectionId: string, config: T) {
+    public async addDaemon(connectionId: string, config: T) {
         if (!connectionId) {
             throw new Error('Cannot add a daemon to local store if connection ID is undefined');
         }
 
-        const daemons = this.daemonStore.getDaemons();
-        daemons[connectionId] = config;
-        this.daemonStore.setDaemons(daemons);
+        await this.daemonStore.lockDb();
```
johncmerfeld, line 84: Given that this can throw an error, would it make sense to wrap this in some retry logic? Seems like there could be a very good reason for the DB to be locked at any given moment and we could just wait for it to open up again

src/services/config/config.service.ts
```diff
@@ -186,60 +188,68 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
...
 
-    addToZliIpcChannels(id: number): void {
-        const channels = this.getZliIpcChannels();
+    async lockDb() {
+        await this.store.lockDb();
+    }
+
+    async unlockDb() {
+        await this.store.unlockDb();
+    }
```
johncmerfeld, line 201: Same error-handling question here -- I'm missing how this works in a multi-ZLI environment without throwing errors

johncmerfeld: Oh I see, the direct `lockDb()` and `unlockDb()` methods are only used in cases where we have to get/set multiple things (token refresh, daemons)

src/services/config/config.service.ts
```diff
@@ -372,167 +384,167 @@ export class ConfigService implements IKubeDaemonSecurityConfigService, IKubeCon
...
     }
 
-    getAuthUrl(): string {
-        return this.store.getSync(authUrlKey);
+    async getAuthUrlAsync(): Promise<string> {
+        return await this.store.getAsync(authUrlKey);
     }
 
-    async getTokenSet(): Promise<TokenSet> {
+    async getTokenSetAsync(): Promise<TokenSet> {
```
johncmerfeld, line 422: I find this function name kind of confusing if we don't have a `getTokenSetSync` (and since none of the other methods in this class specify `Async`)

johncmerfeld (approved):
I haven't exercised this super hard but haven't found any behavioral bugs in the basic usage of the zli.

Most of my comments are matters of preference / extra things you could do. Mostly wondering about the error handling around lock/unlock; I think I'm just misunderstanding how it works.

Anywho, this is great! If you want to talk further about any of my comments please just re-request my review or ping me directly since I never get github notifications

src/services/daemon-management/daemon-management.service.ts
ymarcus93: Tested with 50 parallel `zli connect`. It worked

ymarcus93 (approved):
Tested successfully

src/services/config/leveldb/leveldb.ts
lipmas: agreed i can rename these to get/set/clear now that everything is async

src/services/config/config.service.ts
lipmas: oops yeah that one snuck by - i tried to rename everything to drop the `Async` suffix now that we dont have a mixture of sync/async

src/services/config/zli-auth-config.service.ts
```diff
@@ -19,19 +19,22 @@ export class ZliAuthConfigService implements AuthConfigService {
 
     // This is only being used for the signalR hubs that are not versioned yet
     getServiceUrl() {
-        return this.configService.getServiceUrl() + 'api/v1/';
+        throw new Error('SSM is deprecated and will be removed');
+        return '';
```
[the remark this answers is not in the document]
lipmas: Alright my thinking was to try and not couple the ssm removal with this PR as much as possible in case things get messy but let me at least give it a try.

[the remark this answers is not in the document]
lipmas: Ok it really wasnt too bad. Removed all the ssm handler code for shell/ssh that interacted with the config in [0ad32c3](https://github.com/bastionzero/zli/pull/730/commits/0ad32c3ded73efcdd6d255952ef3ce2836f177f2). This means i could just delete this file and update webshell-common-ts submodule back to master since we dont need to update the interface (it can just be removed instead in a follow-up PR).

Still a lot more ssm related cleanup that needs to happen for zli/webshell-common-ts but that can be done in separate PRs. Here i only removed the code that directly interacted with the config service.

src/services/ssm-tunnel/ssm-tunnel.service.ts
```diff
@@ -86,7 +86,7 @@ export class SsmTunnelService
     }
 
     private async setupEphemeralSshKey(identityFile: string): Promise<void> {
-        const bzeroSshKeyPath = this.configService.getSshKeyPath();
+        const bzeroSshKeyPath = await this.configService.getSshKeyPath();
```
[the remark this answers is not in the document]
lipmas: [0ad32c3](https://github.com/bastionzero/zli/pull/730/commits/0ad32c3ded73efcdd6d255952ef3ce2836f177f2)

src/services/config/config.service.ts
lipmas: Right `lockDb` and `unlockDb` are only used in a couple places where we explicitly want to keep a lock on the db for the entire duration. If `lockDb` isnt first called then get/set/clear will still do what it was doing before where it opens up a db handle/lock only for the duration of that specific get/set/clear operation.

I will say there is definitely more optimization that  could be done throughout the zli where we need to read/write multiple values in the same place that will still currently end up opening/closing the db for each operation. Making those optimizations definitely felt out-of-scope for this PR though.

src/services/daemon-management/daemon-management.service.ts
johncmerfeld: I mean before you even get to `open`, there's this:
```ts
    public async lockDb() {
        if(this.db) {
            throw new Error('leveldb is already locked');
        }
        this.db = await open(this.where);
    }
```

If two processes are trying to lock at the same time, won't this cause the second one to error out?

src/services/http/http.service.ts
```diff
@@ -68,18 +44,44 @@ export class HttpService {
         });
     }
 
-    private async setHeaders(extraHeaders? : Dictionary<string>) {
+    protected async getBaseUrl(): Promise<string> {
+        if(! this.baseUrl) {
+            // Build our base url compensating for minor malformations
+            const url = new URL(await this.configService.getServiceUrl());
```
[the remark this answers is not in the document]
lipmas: Hmm so i cant reproduce this on my end actually it seems like its working and the http service will append the trailing slash even if its not there in the config. What steps do you take exactly?

```
➜  zli git:(daemon-refresh-with-async) ✗ zli-dev configure set serviceUrl https://sebby.bastionzero.com 
Value saved to config!
serviceUrl: https://sebby.bastionzero.com
➜  zli git:(daemon-refresh-with-async) ✗ zli-dev configure get serviceUrl                              
https://sebby.bastionzero.com
➜  zli git:(daemon-refresh-with-async) ✗ zli-dev lt -d --debug
  _                                 
 | |__    ____   ___   _ __    ___  
 | '_ \  |_  /  / _ \ | '__|  / _ \ 
 | |_) |  / /  |  __/ | |    | (_) |
 |_.__/  /___|  \___| |_|     \___/ 
                                    
[Info][2023-07-11 13:38:26.858] You can find all logs here: /home/sebby/.config/bastionzero-logger-nodejs/bastionzero-zli-dev.log
[Debug][2023-07-11 13:38:26.859] Using existing database: /home/sebby/.config/bastionzero-zli-nodejs/dev-store
[Debug][2023-07-11 13:38:26.945] Loaded MrTAP keys
[Debug][2023-07-11 13:38:27.135] Successfully tracked event
[Info][2023-07-11 13:38:27.360] Logged in as: sebby@bastionzero.com, bzero-id:7f59e879-5f58-4d3b-bd23-44381980080d, session-id:cd67c90b-15eb-4094-af68-f28a5d5e633d
[Trace][2023-07-11 13:38:27.731] Making request to: https://sebby.bastionzero.com/api/v2/targets/database/
[Trace][2023-07-11 13:38:27.813] Making request to: https://sebby.bastionzero.com/api/v2/targets/bzero/
[Trace][2023-07-11 13:38:27.837] Making request to: https://sebby.bastionzero.com/api/v2/targets/ssm/?showDynamicAccessTargets=true
[Trace][2023-07-11 13:38:27.940] Making request to: https://sebby.bastionzero.com/api/v2/targets/dynamic-access/
[Trace][2023-07-11 13:38:28.020] Making request to: https://sebby.bastionzero.com/api/v2/targets/web/
[Trace][2023-07-11 13:38:28.060] Making request to: https://sebby.bastionzero.com/api/v2/targets/kube/
[Trace][2023-07-11 13:38:28.073] Request completed to: https://sebby.bastionzero.com/api/v2/targets/database/
[Trace][2023-07-11 13:38:28.193] Request completed to: https://sebby.bastionzero.com/api/v2/targets/ssm/?showDynamicAccessTargets=true
[Trace][2023-07-11 13:38:28.206] Request completed to: https://sebby.bastionzero.com/api/v2/targets/bzero/
[Trace][2023-07-11 13:38:28.252] Request completed to: https://sebby.bastionzero.com/api/v2/targets/dynamic-access/
[Trace][2023-07-11 13:38:28.271] Request completed to: https://sebby.bastionzero.com/api/v2/targets/web/
[Trace][2023-07-11 13:38:28.378] Request completed to: https://sebby.bastionzero.com/api/v2/targets/kube/
[Trace][2023-07-11 13:38:28.464] Making request to: https://sebby.bastionzero.com/api/v2/environments/
[Trace][2023-07-11 13:38:28.758] Request completed to: https://sebby.bastionzero.com/api/v2/environments/
...
```

src/services/config/config.service.ts
johncmerfeld: Gotcha, and yeah agreed, no need to optimize beyond the scope of transactions that specifically need this

src/services/daemon-management/daemon-management.service.ts
lipmas: Oh i see. So if two separate zli processes are technically completely different nodejs processes and `this.db` will be undefined in both of them and each will try and acquire the db lock via leveldb. The `lockdb` and `unlockDb` methods are meant to allow synchronization across zli/daemon processes but not within a single zli process.

If we were to call `lockDb()` twice within the same zli process and not await them in series (i.e use a `Promise.All`) then the second one would error from the above code (nodejs is inherently single threaded so one of the promises would assign `this.db` first). But we really shouldnt be writing any zli code that attempts to do this which is why i just decided to throw an error in this case.

johncmerfeld: Ohh yep you're right. I just misunderstood. For some reason I was thinking about multiple ZLIs sharing a process, but of course that's not what would happen. Makes sense and I'm glad we have the error here to catch any potentially incorrect usage of this.

src/services/http/http.service.ts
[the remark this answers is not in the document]
lipmas: [3b32aec](https://github.com/bastionzero/zli/pull/730/commits/3b32aec220d492060e7a71206e2cdb4bafe4a16f)

PR #752, opened 2023-07-24
Add `zli list-agents` command

johncmerfeld:
## Description of the change

Add a new, admin-only command to list agents, as distinct from targets. Try messing around with all the available filters and see if you can break it!

## Testing

**backend branch:** 
**bzero branch:** develop
**charts branch:** 

- [x] Yes

Release Notes: Add `zli list-agents` command

## Related JIRA tickets

Relates to JIRA: CWC-2723

Relevant HotFix PR Number:

- [ ] Yes
- [ ] No

If yes, please explain:

- [x] Yes
- [ ] No

If yes, please provide the docs/zli-reference-manual branch reflecting the changes: N/A


src/handlers/list-agents/list-agents.command-builder.ts
```diff
@@ -0,0 +1,90 @@
...
+            }
+        )
+        .option(
+            'json',
+            {
+                type: 'boolean',
+                default: false,
+                demandOption: false,
+                alias: 'j',
+                description: 'Formats the ouput in JSON format'
```
SaiDadireddy, line 83: nit: `ouput` -> `output`

src/cli-driver.ts
```diff
@@ -556,6 +582,16 @@ export class CliDriver
...
+            .command(
+                ['list-agents', 'la'],
+                'List all bzero agents (filters available)',
+                (yargs) => {
+                    return listAgentsCmdBuilder(yargs, this.agentTypeChoices, this.targetStatusChoices);
+                },
+                async (argv) => {
+                    await listAgentsHandler(this.configService, this.logger, argv);
+                }
+            )
```
lipmas, line 594: Have we thought about just making this a subcommand of agent so its `zli agent list` instead? I know this isnt how `list-targets` currently works but it does seem to be more in the style of other commands we have added (like policy) and imo provides a nicer organizational structure to zli commands.

The downside is the discrepancy between this and list-targets (but maybe we also want to change that one too) and it being slightly longer to type `agent list` then `la` if using the shortcut. But there are things we could do about the latter for instance using `ls` alias i.e `agent ls` (like [docker does](https://docs.docker.com/engine/reference/commandline/image_ls/))  or have `zli agent` by itself just default to listing agents.

If we want to keep things consistent with `list-targets` for now that makes sense but figured id at least throw out the idea.

johncmerfeld: Yeah, I see the merits of this for sure! Will chat with Ann and get back to you

src/handlers/list-agents/list-agents.command-builder.ts
```diff
@@ -0,0 +1,90 @@
...
+                choices: agentTypeChoices,
+                array: true,
+                demandOption: false,
+                alias: 't',
+                requiresArg: true,
+                description: 'Filter results with one or more agent types'
+            }
+        )
+        .option(
+            'status',
```
SaiDadireddy, line 47: Can we alias this so that we can use `-s` for the flag?

SaiDadireddy (commented):
Zli changes look good! I would like to ask if we can add a system test though. I think we can reuse the `list-targets` suite [here](https://github.com/bastionzero/zli/blob/553e6e15c15cb102fe4ad8aefcdd648ac141545d/src/system-tests/tests/suites/list-targets.ts#L8-L61) and add a similar test case for `list-agents` as we already have for `list-targets`.

johncmerfeld:
> Zli changes look good! I would like to ask if we can add a system test though. I think we can reuse the `list-targets` suite [here](https://github.com/bastionzero/zli/blob/553e6e15c15cb102fe4ad8aefcdd648ac141545d/src/system-tests/tests/suites/list-targets.ts#L8-L61) and add a similar test case for `list-agents` as we already have for `list-targets`.

I looked into this but I actually don't think it's worth it. If you look at the `list-targets` system test, it's actually not interacting with the backend, making it effectively a unit test that we house in a different location. Due to the logistical challenges of registering multiple agents in the context of system tests, I don't personally think I have time to do this in the current iteration.

src/handlers/list-agents/list-agents.command-builder.ts
johncmerfeld: Alas, we have a global `--silent` flag aliased to `-s`.

src/cli-driver.ts
johncmerfeld: Product asked for 
> `zli agent list` that has an alias of `zli la`

It requires a bit of finagling but I'm happy with the result!

